PR #205, opened 2023-12-21
vscode fixes

aaronvg:
- decrease playground bundle size
- fix vscode tests not formatting correctly for named arg functs
- reorganize baml init, add links to docs
- some style fixes in playground (play buttons, output panel size).
- always print the test url to console


typescript/vscode-ext/packages/vscode/src/panels/WebPanelView.ts
```diff
@@ -202,6 +202,8 @@ export class WebPanelView {
               )
             let testInputContent: any
 
+            console.log('saveTestRequest.params.type' + saveTestRequest.params.type)
+            console.log('saveTestRequest.params.value ' + JSON.stringify(saveTestRequest.params.value))
```
hellovai, line 206: @aaronvg console logs!

PR #229, opened 2023-12-30
Fix test case issue with including " and \n and other special characters.

hellovai:
Currently this is still quite hacky, we should do something more robust and actually validate the json at compile time in BAML rather than what we are doing today which is evaluating it at runtime.


aaronvg:
Are there any regressions with enum input or output, or single arg strings or booleans?

Issue #255, opened 2024-01-02
[Docs] Redo the overview docs (no ai engineer, product, etc pages)


aaronvg:
Update w/ latest cleaner intro

PR #302, opened 2024-01-05
docs revisions

aaronvg:
- Update overview
- some more tweaks

Address #255


hellovai:
Some of the links are broken FYI that point to level-1

hellovai (changes requested):
see comment above

PR #444, opened 2024-02-28
feat: implement `baml version --check --output json`

sxlijin:
Also implement:

- `baml version --check`
- `baml version --output json` which mimics `baml version --check --output json`, but omits all latest version information
- `baml version` which mimics `baml --version`

Here's the behavior of the new command:

image


engine/baml-cli/src/update_cli.rs
```diff
@@ -0,0 +1,236 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Versions {
+    pub cli: semver::Version,
+    pub py_client: Option<semver::Version>,
+}
+
+impl Versions {
+    pub fn from_url() -> Result<Self, CliError> {
+        let url = "https://raw.githubusercontent.com/GlooHQ/homebrew-baml/main/version.json";
```
aaronvg: can you use BoundaryML as the org name? We just renamed recently

engine/baml-cli/src/update_cli.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+fn update_windows() -> Result<(), CliError> {
+    run_command_with_error("scoop", ["update"], "scoop update")?;
+    run_command_with_error("scoop", ["update", "baml"], "scoop update gloo")
+}
+
+fn update_shell_install() -> Result<(), CliError> {
+    static INSTALL_SCRIPT: &str =
+        "https://raw.githubusercontent.com/GlooHQ/homebrew-baml/main/install-baml.sh";
```
aaronvg: same thing

engine/baml-cli/src/version_command.rs
```diff
@@ -0,0 +1,263 @@
...
+    package_version_command: &str,
+) -> Result<String, CliError> {
+    let cmd = shellwords::split(package_version_command)
+        .map_err(|e| CliError::StringError(e.to_string()))?;
+
+    let mut cmd = build_shell_command(cmd);
+
+    // NOTE(sam): no idea why this has to start in the cwd; this is copied from update_client.rs
+    // according to vbv@ this had to be done for _some_ reason, so just preserving it as closely as i can
+    let cwd = std::env::current_dir()?.canonicalize()?;
```
aaronvg, line 59: yes theres some weird shit that goes on on Windows vs MacOS with this. I appreciate the comment!

engine/baml-cli/src/version_command.rs
```diff
@@ -0,0 +1,263 @@
...
+        cli: CliVersion {
+            current_version: clap::crate_version!().to_string(),
+            latest_version: None,
+        },
+        clients: Vec::new(),
+        vscode: VscodeVersion {
+            latest_version: None,
+        },
+    };
+    let url = "https://raw.githubusercontent.com/GlooHQ/homebrew-baml/main/version.json";
```
aaronvg: same thing -- maybe make a constant?

engine/baml-cli/src/update_cli.rs
sxlijin: this entire file should've been deleted, mb

sxlijin:
here's what the previous iteration of this PR did:



when updates are available (data in this screenshot is wrong, "latest is newer than current" check wasn't implemented yet):
image

when everything is up-to-date:
image

and the json output:
image

PR #468, opened 2024-03-11
test cases should be passed through as expressions

sxlijin:


aaronvg (approved):
LGTM

engine/baml-lib/baml-core/src/generate/generate_ts_client/expression.rs
```diff
@@ -31,9 +31,12 @@ impl ToTypeScript for Expression {
                 Identifier::Ref(r) => format!("\"{}\"", r.join(".")),
                 Identifier::Primitive(p) => p.to_ts(),
             },
-            Expression::String(val) => format!("\"{}\"", val.replace('"', "\\\"")),
+            Expression::String(val) => {
+                format!("\"{}\"", val.replace('"', "\\\"").replace('\n', "\\n"))
```
hellovai: shoudln't we do this for every possible escaped character? or i guess just \n are special because quoted strings can capture them? For clarity sake, i think we should do the same for "\t" as well

engine/baml-lib/schema-ast/src/ast/expression.rs
```diff
@@ -146,9 +150,42 @@ pub enum Expression {
     Map(Vec<(Expression, Expression)>, Span),
 }
 
+impl Expression {
```
hellovai, line 153: nit: could instead implement serde deserialize for this instead.

clients/ts/tests/test_helpers.ts
```diff
@@ -5,11 +5,12 @@ enum Category {
...
 }
 
-
 registerEnumDeserializer({
     type: "string",
     title: "Category",
     enum: ["ONE", "TWO"]
 }, {});
 
+// need to add aliases here
```
hellovai: remove

clients/ts/src/deserializer/enum_deserializer.ts
```diff
@@ -50,43 +65,123 @@ class EnumDeserializer<T extends Record<string, string>> extends BaseDeserialize
...
         }
 
-        const regex = /^[\w#.-]*\w(?=[:\s.]|$)/;
-        const potentialMatch = parsed.match(regex);
-        if (potentialMatch === null || potentialMatch.length === 0) {
-            diagnostics.pushUnknownError(`Invalid format: ${parsed}`);
-            diagnostics.popScope(false);
-            return Result.failed();
+        const search = (contents: string, aliases: () => IterableIterator<[string, keyof T]>): keyof T | undefined => {
+            for (const [alias, value] of aliases()) {
```
hellovai: why is aliases a function? can't we compute the list once?

hellovai (approved):

clients/ts/src/deserializer/enum_deserializer.ts
sxlijin: fixed

engine/baml-lib/baml-core/src/generate/generate_ts_client/expression.rs
sxlijin: leaving a todo to use escape_default here

engine/baml-lib/schema-ast/src/ast/expression.rs
sxlijin: it needs the spans

PR #494, opened 2024-03-21
feat: implement round robin client strategies

sxlijin:
supports both client syntaxes in `strategy`, and also supports a `start` option which specifies either a client name or `'random'` (defaulting to `random`)


hellovai (approved):
lgtm

PR #532, opened 2024-03-27
Pipenv support

smravec:
Added Pipenv support to the init script. New to rust so better double check it if it does not break anything


aaronvg:
Thanks for the PR, we'll merge this in soon. Have you been able to get BAML up and running?

hellovai (changes requested):
This looks pretty great, but can you also add a unit test to this for pyenv?

https://github.com/BoundaryML/baml/tree/canary/engine/.docker/python-tests

the run instructions are noted: https://github.com/BoundaryML/baml/blob/canary/engine/e2e_tests.py

smravec:
Added the unit test just now

smravec:
I was waiting for the ollama support, will try to set it up now

hellovai (approved):
LGTM!

hellovai:
Hi @smravec we actually removed any need on python environments as of this point!

it now works in any python env via:
```sh
> pip install baml-py
> baml-cli init
> baml-cli generate
```

`baml-cli` is now bundled in with `baml-py` and there's no longer a dependency on our CLI which makes the installation process so much easier for users (we hope).

your PR for adding pyenv and the struggle we had with merging it in was actually a key reason why we went down this path. Glad to share more thoughts if you'd like!

PR #544, opened 2024-04-04
Add jsonish support

hellovai:


engine/baml-lib/jsonish/src/deserializer/test_iterative_parser.rs
```diff
@@ -0,0 +1,279 @@
...
+    "key": "value",
+    "array": ["1", "2", "3"],
+    "object": {
+      "key": "value"
+    }
+  }
+);
+
+test_jsonish!(
+  test_json_with_unquoted_values_with_spaces,
```
hellovai, line 219: @villagab4 this passes btw

if the string fix works for you, i'll pause on this for now to mitigate errors

PR #545, opened 2024-04-04
rename

hellovai:
- **do legacy pass on grammar**
- **Working function type with errors**
- **Some code gen works**


engine/baml-lib/baml-core/src/validate/validation_pipeline/validations/functions.rs
```diff
@@ -45,4 +45,33 @@ pub(super) fn validate(ctx: &mut Context<'_>) {
...
+                ctx.push_error(DatamodelError::not_found_error(
+                    "Client",
+                    &client.0,
+                    client.1.clone(),
+                    clients.clone(),
+                ))
+            }
+        }
+
+        // TODO: @sxlijin do any validation on the prompt.
```
hellovai: @sxlijin here!

PR #546, opened 2024-04-05
feat: expose render_prompt to python

sxlijin:
releases not set up yet


engine/baml-core-ffi/index.js
```diff
@@ -219,6 +219,43 @@ switch (platform) {
           loadError = e
         }
         break
+      case 'riscv64':
```
aaronvg, line 222: this was autogenerated right? We do use the build script to generate this index file IIRC

PR #614, opened 2024-05-22
Jump2def for vscode-extension

anish-palakurthi:
Supports enums, classes, functions, clients, retry policies, and template strings.
Adjusted span logic for some symbols, but mainly added functionality to bamlProjectManager, called from TS.

### DIFF
--- a/typescript/vscode-ext/packages/language-server/src/server.ts
+++ b/typescript/vscode-ext/packages/language-server/src/server.ts
@@ -365,15 +365,6 @@ export function startServer(options?: LSOptions): void {
 
       } 
   
-
-      
-      else if (doc.languageId === 'python') {
-        const db = bamlCache.lastBamlDir?.cache
-        
-        if (db) {
-          return MessageHandler.handleDefinitionRequest(db, doc, params)
-        }
-      }
     }
   })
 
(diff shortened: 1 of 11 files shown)


typescript/vscode-ext/packages/language-server/src/server.ts
```diff
@@ -348,20 +348,33 @@ export function startServer(options?: LSOptions): void {
...
+      if (proj) {
+
+        //returns the definition of reference within the project
+        return proj.handleDefinitionRequest(doc, params.position)
+
+      } 
+  
+
+      
+      else if (doc.languageId === 'python') {
```
aaronvg: i dont think we need to do anything for python language -- or what is this supposed to do?

anish-palakurthi: It was already there, so I didn't modify it.

anish-palakurthi: Part of the commented out code^^

aaronvg: we can delete this python-specific code for now. The Since we're going from commenting it out to enabling it, and we don't know what it is supposed to do.

anish-palakurthi: Ok will do.

aaronvg (approved):

PR #619, opened 2024-05-23
 Fixed playground function swapping based on click 

anish-palakurthi:
Previously wasn't restricted to functions in the same file. Was matching with any function whose span the cursor fell into, not just the ones in the same file whose span it fell under.


typescript/playground-common/src/baml_wasm_web/EventListener.tsx
```diff
@@ -1,4 +1,5 @@
 import 'react18-json-view/src/style.css'
+// import * as vscode from 'vscode'
```
aaronvg, line 2: delete

aaronvg (approved):

PR #624, opened 2024-05-29
Fixed end of function swapping bug

anish-palakurthi:
Previously, cursor index was being incorrectly calculated, causing clicks near the end of functions to cause switching to other functions.
Now: fixed cursor index calculation, issue is resolved.

### DIFF
--- a/engine/baml-runtime/src/lib.rs
+++ b/engine/baml-runtime/src/lib.rs
@@ -28,7 +28,6 @@ use anyhow::Result;
 
 use baml_types::BamlMap;
 use baml_types::BamlValue;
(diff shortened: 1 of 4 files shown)


engine/baml-runtime/src/lib.rs
```diff
@@ -28,7 +28,7 @@ use anyhow::Result;
 
 use baml_types::BamlMap;
 use baml_types::BamlValue;
-use cli::LanguageClientType;
+// use cli::LanguageClientType;
```
aaronvg: remove if it's not used.

aaronvg (approved):

PR #643, opened 2024-06-03
Proxy all providers

anish-palakurthi:


typescript/vscode-ext/packages/vscode/src/extension.ts
```diff
@@ -248,11 +248,26 @@ export function activate(context: vscode.ExtensionContext) {
 
   const server = express()
   server.use(cors())
+  // server.use(
+  //   '/',
+  //   createProxyMiddleware({
+  //     target: 'https://api.anthropic.com',
+  //     changeOrigin: true,
+  //   }),
+  //
```
aaronvg: delete

typescript/vscode-ext/packages/vscode/src/extension.ts
```diff
@@ -248,11 +248,26 @@ export function activate(context: vscode.ExtensionContext) {
...
   server.use(
     '/',
     createProxyMiddleware({
-      target: 'https://api.anthropic.com',
       changeOrigin: true,
+      router: (req) => {
+        // Extract the original target URL from the custom header
+        const originalUrl = req.headers['original-url']
```
aaronvg: rename teh header to baml-original-url, and can we remove this header before sending it out to the actual url? Some backend services are strict and could fail if you send extra headers.

anish-palakurthi:
FYI, we now proxy all providers and the port used by the application server is not hardcoded to 8195, but acquired dynamically by the OS to prevent port conflict.

aaronvg (approved):

PR #683, opened 2024-06-16
Add support for dynamic clients

hellovai:
BAML functions now support passing in which model / configuration they use at runtime.
* Enables dynamically picking which options they want to run.
* Tracing captures dynamic properties
* Retry policies must be defined in BAML


irdbl:
Do you have any idea of when this will be merged? We are eagerly awaiting for this release for some use cases regarding custom headers and params for openai API calls.

hellovai:
> Do you have any idea of when this will be merged? We are eagerly awaiting for this release for some use cases regarding custom headers and params for openai API calls.

Sorry about the delay! we had some tests i was running on this! I'll aim to merge this in today.

For headers and custom paramters, you can also do it via .baml files directly if you want see: https://docs.boundaryml.com/docs/snippets/clients/providers/openai#non-forwarded-options

The custom headers field. Would that work?

sxlijin (approved):
integ-tests match docs for python + ts

hellovai:
@accupham we've finally merged this! you can expect this to land in our release tmrw or thursday!

PR #692, opened 2024-06-19
Generic multimodal BAML type

anish-palakurthi:
BAML now supports both image and audio media input for providers that support these inputs.
These can be used from imported BAML, and rendered/tested in the BAML playground.
Fixes: 
Claude can now be prompted with URLs via BAML request layer.
Default roles are configured on a per-provider basis rather than always "system."

Adds:
Integration tests for base64 and URL media inputs, both image and audio.


integ-tests/baml_src/test-files/functions/input/named-args/single/named-audio.baml
hellovai, line 1: split test case vs function into two files so only 1 file is excluded

integ-tests/python/test_functions.py
hellovai, line 1: b64 should be a variable in a special file

hellovai (changes requested):

integ-tests/python/test_functions.py
```diff
@@ -411,6 +452,10 @@ async def test_stream_dynamic_class_output():
     for prop in tb.DynamicOutput.list_properties():
         print(f"Property: {prop}")
 
+    stream = b.stream.MyFunc(
+        input="My name is Harrison. My hair is black and I'm 6 feet tall.",
+        baml_options={"tb": tb},
+    )
     stream = b.stream.MyFunc(
```
aaronvg: whys this called twice?

anish-palakurthi: probably was from a merge. deleted it.

engine/baml-runtime/src/internal/llm_client/primitive/openai/openai_client.rs
```diff
@@ -393,15 +393,18 @@ macro_rules! make_openai_client {
     ($client:ident, $properties:ident) => {
         Ok(Self {
             name: $client.name().into(),
-            properties: $properties,
+            // properties: $properties,
```
aaronvg: delete

engine/baml-runtime/Cargo.toml
```diff
@@ -55,6 +55,8 @@ pin-project-lite = "0.2.14"
 async-trait = "0.1.80"
 cfg-if = "1.0.0"
 include_dir = "0.7.3"
+futures-lite = "2.3.0"
+infer = "0.16.0"
```
aaronvg: can you fill me in on what these are used for?

hellovai (approved):

PR #694, opened 2024-06-19
fix icon in navbar

lukaskf:
Hopefully this will do it


sxlijin:
Ah, silly us- thanks for the fix!

aaronvg:
thanks!

PR #723, opened 2024-06-27
Transparent request builder

anish-palakurthi:
New feature:
See the raw cURL request that BAML generates for you, providing full transparency into BAML's workings.


engine/.turbo/daemon/d1f419c7a2c5692a-turbo.log.2024-06-26
hellovai, line 1: delete file

engine/baml-lib/jinja/Cargo.toml
```diff
@@ -36,6 +36,7 @@ serde_json.workspace = true
 log = "0.4.14"
 strsim = "0.11.1"
 colored = "2.1.0"
+wasm-bindgen = "0.2.92"
```
hellovai: ?

PR #726, opened 2024-06-27
swapped asyncio for thread

anish-palakurthi:
Fixed streaming buffer delay by separating stream handler to another thread.


sxlijin (approved):

sxlijin:
verified this works with tracing

PR #741, opened 2024-07-02
prompt fiddle2

aaronvg:
- **new promptfiddle files**
- **Add prompt fiddle project with all examples**


typescript/fiddle-frontend/public/_examples/all-projects/main.py
```diff
@@ -0,0 +1,13 @@
+from baml_client import baml as b
```
hellovai:
```suggestion
from baml_client import baml as b
```
```suggestion
from baml_client import b
```

PR #742, opened 2024-07-02
Adding better UX for playground

hellovai:
* Better render envvars buttons
* Show docs links in playground
* Use dropdowns to select test cases / functions
* Add jump to definition in playground

Screenshot 2024-07-02 at 10 28 19 PM
Screenshot 2024-07-02 at 10 21 50 PM


typescript/vscode-ext/packages/vscode/src/panels/WebPanelView.ts
```diff
@@ -201,8 +201,9 @@ export class WebPanelView {
           }
           case 'jumpToFile': {
             try {
+              console.log('jumpToFile', message.data)
               const span = message.data as StringSpan
-              const uri = vscode.Uri.file(span.source_file)
+              const uri = vscode.Uri.parse(span.source_file)
```
aaronvg, line 207: What does the file path loom like here? Windows wont work if we use this on a non "file:///" path.

I made this note for ourselves; https://www.notion.so/gloochat/windows-uri-treatment-fe87b22abebb4089945eb8cd1ad050ef?pvs=4

hellovai: it is a file:/// fyi!

PR #749, opened 2024-07-03
fix: release on arm64-linux

sxlijin:
This matters for people using VMs; arm64-linux is important for a lot of users:

- the default linux container on macOS is arm64-linux, because cpu compatibility
- a lot of cloud users use arm64-linux because graviton machines are cheap


sxlijin:
superseded by #751

PR #763, opened 2024-07-08
Bug fixes to cURL rendering and mime_type overriding

anish-palakurthi:
Fixes:
- mime_type is inferred from URL media input for non-URL-supporting providers, we now allow users to override this if they manually specify the media's mime_type
- azure requests cURL rendering now include query parameters
- raw cURL panel lets users toggle between streaming and non-streaming versions of their request
- proxy media URL resolution calls for CORS media blocks 
- style changes for cURL panel

Screenshot 2024-07-08 at 3 34 22 PM


hellovai (commented):
LGTM

PR #773, opened 2024-07-11
bug: correctly restrict requests in the fiddle proxy

sxlijin:


hellovai:
This patches the issue!

PR #790, opened 2024-07-15
Support Vertex AI (Google Cloud SDK)

### LINKED ISSUE #706: Offer Vertex-AI as a provider
Vertex-ai is another commonly offered interface for speaking to LLMs on-top of gemini directly. It has issues in the way we get the authorization token (needs to be refreshed frequently w/ oauth). We can support this if many people require it.

anish-palakurthi:
- Added support for the Vertex AI provider
- Automate Google authorization token fetching on per-request basis with JWT signing from Google Cloud service account credentials
- Added enum for different cases of resolve media -- never, when mime type is absent, and always -- helpful for providing interoperability between multimodal providers
- Added docs and integ-tests for Vertex AI to Python/TS

Improved capabilities of Vertex AI
- supports media input from any URL or base64 input (Google AI does not take URL input)
- more advanced safety settings/generation configurations

Integ-tests pass 

See docs for further information on Vertex AI


engine/baml-runtime/src/internal/llm_client/primitive/mod.rs
```diff
@@ -95,13 +100,18 @@ impl TryFrom<(&ClientProperty, &RuntimeContext)> for LLMPrimitiveProvider {
             "anthropic" => {
                 AnthropicClient::dynamic_new(value, ctx).map(LLMPrimitiveProvider::Anthropic)
             }
-            "google-ai" => GoogleClient::dynamic_new(value, ctx).map(LLMPrimitiveProvider::Google),
+            "google-ai" => {
+                GoogleAIClient::dynamic_new(value, ctx).map(LLMPrimitiveProvider::Google)
+            }
+            // "aws-bedrock" => aws::AwsClient::dynamic_new(value, ctx).map(LLMPrimitiveProvider::Aws),
```
sxlijin: what happened here?

integ-tests/baml_src/test-files/functions/input/named-args/single/named-int.baml
```diff
@@ -1,6 +1,6 @@
 // test for int
 function TestFnNamedArgsSingleInt(myInt: int) -> string {
-  client GPT35
+  client Vertex
```
sxlijin: revert

sxlijin (approved):

engine/baml-runtime/src/internal/llm_client/primitive/mod.rs
anish-palakurthi: The dynamic_new method is not implemented for AWS right now and AWS was never present here. I just added a commented version of it so it was more explicit. I can either remove it entirely or just explain this there.

PR #794, opened 2024-07-16
LLVM install and rebuild script

anish-palakurthi:
Added script to shift developer usage of Apple Clang to LLVM Clang, as Apple Clang does not support all WASM versions of packages. A one-time script, only for use in development of BAML. Will not affect end user as binaries are produced.


engine/llvm_install.sh
sxlijin, line 1: rename to `tools/reinstall_llvm_on_macos.sh`

sxlijin (approved):

PR #797, opened 2024-07-17
feat: implement maps in BAML

### LINKED ISSUE #787: BAML language: maps
BAML currently doesn't support map types. Add support for it.

sxlijin:
Add support for maps in the BAML language. We currently only support maps with string keys.

Here's the work item list that we verified (fortunately, most of these were already pre-implemented):

- integ-tests (python, ts, ruby)
- parser for the compiler
    - grammar
    - schema-ast
- parser for codemirror
    - lezer
- vscode: textmate grammar
- diagnostics
    - not sure what this means concretely
- in-Rust implementation, i.e. BamlValue
- value syntax
    - ie how you define a map in a test
- input layer (lang-native object → BamlValue)
- ctx.output_format
- output layer (BamlValue → lang-native object)
- codegen (python, ts, ruby)

We also added:

- a new entry point for `validate_type_allowed` where we apply refinements to the lexed types
- textmate grammar snapshot testing using vscode-tmgrammar-snap

Fixes #787


engine/baml-lib/jsonish/src/tests/test_maps.rs
```diff
@@ -0,0 +1,45 @@
+use super::*;
```
hellovai, line 3: @sxlijin some more tests would be nice

map with new lines, for keys,

partial maps
nested partial maps
maps with ints as keys to ensure they cast to strings
maps with arrays as key to ensure they cast to strings
maps in json markup
nested maps with new lines
union of map vs class (would the map correctly let the class have it?)

engine/baml-lib/baml/tests/validation_files/class/map_types.baml
```diff
@@ -0,0 +1,64 @@
+// Tests for validating map types
```
hellovai, line 1: nit add a test for return type in a function as well

engine/baml-lib/baml/tests/validation_files/class/secure_types.baml
```diff
@@ -17,207 +17,45 @@ class ComplexTypes {
   o (((int | string) | bool[]), (float, double) | long_long_identifier_123)
 }
 
-// error: Type `apple_pie` does not exist. Did you mean one of these: `ComplexTypes`, `float`, `bool`, `string`, `int`?
+// error: Error validating: This line is not a valid field or attribute definition. A valid class property looks like: 'myProperty string[] @description("This is a description")'
```
hellovai: this error got much worse.

seems we broke the grammar :(

<blocking>

hellovai (commented):
almost, just one grammar bug it seems.

also what does the prompt look like for a map type? I don't see how the code changes for that

integ-tests/baml_src/test-files/functions/input/named-args/single/named-map-string-to-class.baml
```diff
@@ -0,0 +1,19 @@
...
+    Return this value back to me: {{myMap}}
+  "#
+}
+
+test TestFnNamedArgsSingleMapStringToClass {
+  functions [TestFnNamedArgsSingleMapStringToClass]
+  args {
+    myMap {
+      "key" "example string"
+    }
```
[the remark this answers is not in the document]
aaronvg: wait isnt greptile correct here?


myMap {
   "key" { word "example string" }
 }

engine/baml-lib/baml/tests/validation_files/class/secure_types.baml
sxlijin: you reviewed an outdated version of the PR - the grammar wasn't broken, the test was still using the `{key: value}` map syntax

sxlijin:
todo: lex maps of arbitrary parity

PR #801, opened 2024-07-19
Client orchestration graph

anish-palakurthi:
- Added dynamic Client Orchestration Graph visualizer
- allows users to see the execution path of their fallback, retry, and round robin strategies, including the render prompt and raw cURL per node
- uses a DP-based node identifier algorithm to ID the groups that nodes belong to
- recursively generates sizes and positioning of nodes for React-flow component

Screenshot 2024-07-18 at 6 02 37 PM


engine/baml-runtime/src/internal/llm_client/orchestrator/mod.rs
```diff
@@ -7,8 +7,11 @@ use baml_types::BamlValue;
 use internal_baml_core::ir::repr::IntermediateRepr;
 use internal_baml_jinja::RenderedChatMessage;
 use internal_baml_jinja::RenderedPrompt;
+use js_sys;
+use serde::Serialize;
 use std::{collections::HashMap, sync::Arc};
-use web_time::Duration;
+use wasm_bindgen::JsValue;
```
hellovai: too much leakage of JsValue into runtimes (python, ruby, etc. This is a WASM only feature and all things related to wasm_bindgen / JSValue should be behind a cfg for wasm target.

Not doing this can lead to hard to detect bugs in runtimes when mixing and matching WASM with py03, typescript, etc.

engine/Cargo.toml
```diff
@@ -45,7 +45,7 @@ indexmap = { version = "2.1.0", features = ["serde"] }
 indoc = "2.0.1"
 regex = "1.10.4"
 serde_json = { version = "1", features = ["float_roundtrip", "preserve_order"] }
-serde = { version = "1", features = ["derive"] }
+serde = { version = "1", features = ["derive", "rc"] }
```
hellovai, line 47: whats rc?

engine/baml-runtime/src/internal/llm_client/primitive/request.rs
```diff
@@ -45,30 +44,33 @@ pub async fn make_request(
     {
         Ok(req) => req,
         Err(e) => {
+            log::info!("Failed to build request: {:#?}", e);
```
hellovai: bad log messages

hellovai (dismissed):
pretty much done!

1 question: What happens if i select node index 10 in one prompt preview, then switch to a function which only has 1 node? Will it work?

Also do update teh PR description to one more thats more user facing!

engine/Cargo.toml
anish-palakurthi: Rc is a feature that we use to serialize Arc types.

engine/baml-runtime/src/internal/llm_client/orchestrator/mod.rs
anish-palakurthi: Sorry forgot about that.

engine/baml-schema-wasm/src/runtime_wasm/mod.rs
```diff
@@ -1265,7 +1305,9 @@ impl WasmFunction {
         rt: &WasmRuntime,
         params: JsValue,
     ) -> Result<WasmPrompt, wasm_bindgen::JsError> {
-        let params = serde_wasm_bindgen::from_value::<BamlMap<String, BamlValue>>(params)?;
+        let mut params = serde_wasm_bindgen::from_value::<BamlMap<String, BamlValue>>(params)?;
+        let node_index = params.shift_remove("node_index");
```
sxlijin, line 1309: this breaks a user who defines a fn with node_index as an arg

engine/baml-schema-wasm/src/runtime_wasm/mod.rs
```diff
@@ -1346,4 +1409,83 @@ impl WasmFunction {
...
+
+        // Serialize the scopes to JsValue
+        let mut scopes = Vec::new();
+        for scope in graph {
+            scopes.push(WasmScope::from(scope.scope));
+        }
+        Ok(scopes)
+    }
+}
+trait ToJsValue {
```
sxlijin, line 1442: this is a really non-type-safe way to implement this, but i guess it works? it throws away a lot of type system guarantees we have in rust and relies on our ts code hopefully getting the types correct - that works in a single PR but we're definitely going to mistakes with this in the future, esp. e.g. because we won't have auto-complete on the js analogue of ExecutionScope

engine/baml-schema-wasm/src/runtime_wasm/mod.rs
anish-palakurthi: Will fix tmrw morning

engine/baml-schema-wasm/src/runtime_wasm/mod.rs
anish-palakurthi: Yeah...seemed off when I did it but it wasn't letting me add another function otherwise. Could probably find a better alternative.

anish-palakurthi: Will also fix tmrw

PR #802, opened 2024-07-19
Dont drop extra fields in dynamic classes when passing them as inputs to a function

aaronvg:


engine/language_client_python/src/parse_py_type.rs
```diff
@@ -249,25 +249,40 @@ pub fn parse_py_type(
...
-                        })
-                        .collect::<HashMap<_, _>>(),
-                    Err(_) => {
-                        bail!("model_fields is not a dict")
+                        }
                     }
-                };
+                }
+
+                // Log the fields
```
hellovai, line 276: delete comment

hellovai (approved):

PR #812, opened 2024-07-20
Codelenses for test cases

anish-palakurthi:
- added open test and swap test playground rendering
- onclick => if playground is currently open, runs the test. else, opens the playground to that test without running it.
- swaps rendered test based on cursor location. click on a specific function in a testcase's functions list to swap to that implementation.

Screenshot 2024-08-08 at 10 33 29 AM


typescript/vscode-ext/packages/language-server/src/lib/baml_project_manager.ts
```diff
@@ -295,6 +295,10 @@ class Project {
 
     return runtime.list_functions()
   }
+  list_testcases(): BamlWasm.WasmTestCase[] {
+    let runtime = this.runtime()
```
aaronvg, line 301: what happens if the runtime isnt defined like when theres a syntax error? Do we get a language server crash?

typescript/vscode-ext/packages/vscode/src/extension.ts
```diff
@@ -254,6 +254,42 @@ export function activate(context: vscode.ExtensionContext) {
...
+    'baml.runBamlTest',
+    (args?: {
+      projectId?: string
+      functionName?: string
+      implName?: string
+      showTests?: boolean
+      testCaseName?: string
+    }) => {
+      const config = vscode.workspace.getConfiguration()
+      config.update('baml.runBamlTest', true, vscode.ConfigurationTarget.Global)
```
aaronvg: you dont need this

typescript/vscode-ext/packages/language-server/src/lib/baml_project_manager.ts
anish-palakurthi: Yeah you're right, should throw an error

PR #837, opened 2024-07-24
docs: make dynamic clients discoverable

sxlijin:


hellovai:
closing at this is an organizational PR, and we have had too many major submissions since then. we can re-do this rather than merge these changes forward

PR #843, opened 2024-07-29
chore: move chatwoot to the bottom left, not bottom right

sxlijin:
less likely to cover test results or the runtime version badge

https://github.com/orgs/chatwoot/discussions/2145


hellovai:
this can have issues:
- when in test view, you can't see tests or press play buttong

hellovai:
we may want to just create an explicit feedback button instead and map it to whatever they pop up

hellovai (changes requested):
we may want to just create an explicit feedback button instead and map it to whatever they pop up

hellovai:
handled already

PR #857, opened 2024-08-04
AST Restructuring

### LINKED ISSUE #777: First line of an array cannot be a comment
Surface better error or fix Pest grammar to allow this

### LINKED ISSUE #859: bug: compound clients do insufficient compiler validation
User report via Slack connect: they have a `fallback` client where one of the clients doesn't actually exist, which causes prompt rendering to panic in `unreachable!`.

anish-palakurthi:
- Significant rework of the BAML AST, mainly aimed at reducing the complexity of the grammar, parser, and IR validation scheme to yield auto-formatting in the future
- Maintains full feature parity with legacy BAML AST
- Supports multiple key value pairs per line now, with "," delimiter -- tradeoff includes slightly worse errors in map blocks temporarily.
- Addresses minor bugs in BAML LSP


engine/.turbo/daemon/d1f419c7a2c5692a-turbo.log.2024-07-23
hellovai, line 1: log file remove

docs/docs.yml
```diff
@@ -131,12 +131,14 @@ navigation:
...
       - page: Test Cases
         path: docs/snippets/test-cases.mdx
-  - section: Advanced BAML Snippets
+  - section: Advanced BAML Features
     contents:
       - page: Dynamic Types
         path: docs/calling-baml/dynamic-types.mdx
       - page: Client Registry
         path: docs/calling-baml/client-registry.mdx
+      - page: Custom Type Assertions
```
hellovai: bad diff :)

engine/baml-lib/baml/tests/validation_files/class/dependency_cycle.baml
```diff
@@ -8,24 +8,4 @@ class InterfaceOne {
...
-//  4 | 
-//  5 | class InterfaceOne {
-//    | 
-// error: Error validating: These classes form a dependency cycle: InterfaceThree
-//   -->  class/dependency_cycle.baml:9
-//    | 
-//  8 | 
-//  9 | class InterfaceThree {
-//    | 
+}
```
hellovai: this one should have errors

engine/baml-lib/baml/tests/validation_files/class/spelling_error.baml
```diff
@@ -4,22 +4,4 @@ class InterfaceTwo {
 
 class InterfaceOne {
```
hellovai: this file should have errors

hellovai (changes requested):
its close, but the validation files have pretty large diffs. (Lots of old errors are deleted).

is that intentional?

docs/docs/snippets/prompt-syntax/what-is-jinja.mdx
```diff
@@ -20,7 +20,7 @@ When in doubt -- use the BAML VSCode Playground preview. It will show you the fu
 
 Here's how you can iterate over a list of items, accessing each item's attributes:
 
-```jinja
+```jinja Jinja
```
aaronvg, line 23: delete the extra Jinja

engine/baml-lib/baml/tests/validation_files/strings/unquoted_strings.baml
```diff
@@ -8,28 +8,46 @@ client<llm> Hello {
...
+//  3 |   options {
+//    | 
+// error: Error validating: This line is not a valid field or attribute definition.
 //   -->  strings/unquoted_strings.baml:3
 //    | 
 //  2 |   provider baml-openai-chat
 //  3 |   options {
 //  4 |     thing hello'world
 //    | 
-// error: Error validating: This line is not a valid field or attribute definition. A valid property may look like: 'myProperty "some value"' for example, with no colons.
```
aaronvg: feels like the previous error was better because it told you how to fix it. Usuallly people add colons.

engine/baml-lib/baml/tests/validation_files/strings/unquoted_strings.baml
```diff
@@ -8,28 +8,46 @@ client<llm> Hello {
...
+// error: Error validating: This line is not a valid field or attribute definition.
 //   -->  strings/unquoted_strings.baml:4
 //    | 
 //  3 |   options {
 //  4 |     thing hello'world
 //  5 |     banned @helloworld
 //    | 
-// error: Error validating: This line is not a valid field or attribute definition. A valid property may look like: 'myProperty "some value"' for example, with no colons.
+// error: Error validating Client "Hello": This field declaration is invalid. It is either missing a name or a type.
```
aaronvg, line 31: same as above comment for alll these

engine/baml-lib/parser-database/src/walkers/arg.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub fn ast_function(self) -> &'db ast::ValueExprBlock {
+        &self.db.ast[self.id.0]
+    }
+
+    /// The AST node.
+    pub fn ast_arg(self) -> (Option<&'db Identifier>, &'db ast::BlockArg) {
+        match self.id.1 {
+            true => {
+                let args = self.ast_function().input();
+                let res = &args.expect("Expected input args")[self.id.2];
```
aaronvg: seems like the greptile comment here is valid no?

engine/baml-lib/parser-database/src/walkers/arg.rs
```diff
@@ -0,0 +1,74 @@
...
+                (Some(&res.0), &res.1)
+            }
+
+            false => {
+                let output = self.ast_function().output();
+                let res = output.expect("Error: Output is undefined for function ID");
```
aaronvg: ssame thing

integ-tests/python/baml_client/async_client.py
```diff
@@ -1717,7 +1717,7 @@ async def TestVertex(
     
     async def UnionTest_Function(
         self,
-        input: Union[str, bool],
+        input: Union[Union[str], Union[bool]],
```
aaronvg: extra union?

integ-tests/python/baml_client/partial_types.py
```diff
@@ -163,14 +163,14 @@ class Person(BaseModel):
 class Quantity(BaseModel):
     
     
-    amount: Optional[Union[Optional[int], Optional[float]]] = None
+    amount: Optional[Union[Optional[Union[Optional[int]]], Optional[Union[Optional[float]]]]] = None
```
aaronvg: same thing with unions

integ-tests/typescript/baml_client/async_client.ts
```diff
@@ -315,13 +315,13 @@ export class BamlAsyncClient {
   }
   
   async ExtractResume(
-      resume: string,img?: Image | null,
+      resume: string,img: Image | null,
```
aaronvg: why did we remove the undefined? This may break everyone

integ-tests/typescript/baml_client/async_client.ts
```diff
@@ -411,13 +411,13 @@ export class BamlAsyncClient {
...
   async FnNamedArgsSingleStringOptional(
-      myString?: string | null,
+      myString: string | null,
       __baml_options__?: { tb?: TypeBuilder, clientRegistry?: ClientRegistry }
   ): Promise<string> {
     const raw = await this.runtime.callFunction(
       "FnNamedArgsSingleStringOptional",
       {
-        "myString": myString?? null
+        "myString": myString
```
aaronvg: myString ?? null

engine/baml-lib/parser-database/src/walkers/arg.rs
anish-palakurthi: I'm pretty sure this expect is always safe, because if the argument exists, it will be of input or output and be matched directly. If it doesn't exist, then the arg would not have been created in the first place. Prior to this PR, we had a function specific ast node so input/output args were required, but now we use that same ast node structure for other structs as well which have these as optional. However, since this code is specifically for functions, it is safe. Can discuss tomorrow.

hellovai:
LGTM approved

hellovai (approved):

PR #877, opened 2024-08-13
Fix crash in promptfiddle

aaronvg:


aaronvg:
not running orchestration graph on promptfiddle for now, but vscode is good

PR #897, opened 2024-08-24
feat: teach vscode/fiddle to explain when we drop information

sxlijin:
when there's an explanation to be shown, the information icon is clickable and toggles showing the explanation:

image

if there's no explanation, the information icon tooltip just shows "parsing succeeded" and clicking it does nothing

image

### DIFF
--- a/engine/baml-lib/jsonish/src/deserializer/coercer/ir_ref/coerce_class.rs
+++ b/engine/baml-lib/jsonish/src/deserializer/coercer/ir_ref/coerce_class.rs
@@ -239,13 +239,19 @@ impl TypeCoercer for Class {
                 })
                 .collect::<Vec<_>>();
 
-            if !missing_required_fields.is_empty() || !unparsed_required_fields.is_empty() {
+            let unparsed_or_missing = required_values
+                .iter()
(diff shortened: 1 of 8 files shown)


hellovai:
Looks interesting! Let's chat before merge to discuss edge cases!

sxlijin:
use this file to repro it on promptfiddle (and swap out for `openai/gpt-4o-mini` for no errors)

```
// This is a BAML config file, which extends the Jinja2 templating language to write LLM functions.

class Recipe {
  name string
  ingredients Ingredient[]
}
enum IngredientUnit {
  Tbsp
  Lbs
  Grams
  }

class Ingredient {
  name string
  amount int | float
  unit IngredientUnit
}

function ExtractRecipe(subject: string) -> Recipe {
  // see clients.baml
  client "openai/gpt-3.5-turbo"

  // The prompt uses Jinja syntax. Change the models or this text and watch the prompt preview change!
  prompt #"
        {# special macro to print the output instructions. #}
    {{ ctx.output_format }}

    
    Give me a recipe for {{ subject }}
  "#
}

test Test1 {
  functions [ExtractRecipe]
  args {
    subject "beef stew"
  }
}
```

PR #900, opened 2024-08-26
Fix grammar changes introduced by AST changes

hellovai:
We added a new parser in #857 that added a new construct - namely referential properties and also attributes that be attached to types.

There were some minor regressions in the canary build that identified this issue.

We've added some more tests and have ensured that prompts are not modified at all.

Other changes:
* updated parser to improve union parsing


engine/baml-lib/baml-core/src/ir/repr.rs
```diff
@@ -129,6 +129,7 @@ impl IntermediateRepr {
         db: &ParserDatabase,
         configuration: Configuration,
     ) -> Result<IntermediateRepr> {
+        log::debug!("Parser db {:#?}", db.ast());
```
hellovai:
```suggestion
```

PR #912, opened 2024-09-01
activate win32-arm64 build

aaronvg:


hellovai:
@aaronvg should we still do this?

PR #915, opened 2024-09-03
Update build script for ruby to build all platforms

hellovai:


.github/workflows/build-ruby-release.reusable.yaml
```diff
@@ -27,10 +27,14 @@ jobs:
             # https://github.com/rake-compiler/rake-compiler-dock/issues/122
             # for more details.
             rb-sys-dock-setup: ./x86-64_linux-setup.sh
+          - platform: x86_64-linux
+          - platform: x86_64-linux-musl
           - platform: aarch64-linux
-          - platform: x86_64-darwin
```
philip-maina, line 31: Hi @hellovai , 

I noticed that recent baml releases don’t seem to have prebuilt ruby gems for the x86_64-darwin platform, only for arm64-darwin and others. Is this a deliberate choice (e.g., dropping Intel macOS support), or just something that hasn’t been built yet?

Only asking because I’m currently on an Apple Silicon Mac but using an Intel Ruby (Rosetta)

PR #974, opened 2024-09-20
[syntax] Allow lists to contain trailing comma

imalsogreg:
Prior to this change, an expression `[1,2,]` would fail to parse.

This PR updates the pest grammar to allow lists to contain trailing commas.

The new tests in this PR fail if you revert the change to the pest grammar, and pass with the updated grammar.

I also tested this change in baml_runtime, by updating an existing example to take a list of strings as a parameter, and supplying a list with a trailing comma in a test-case. This test fails prior to the grammar update and passes after. I did not check the baml_runtime test in to the codebase.


aaronvg:
looks good to me

PR #978, opened 2024-09-21
Implement literal types

antoniosarosi:
I got something that *compiles* and parses literals into the AST. Lots of `todo!()`s left.

----

> [!IMPORTANT]
> Implement literal types in the codebase, including coercion, validation, and code generation updates for Python, Ruby, and TypeScript.
> 
>   - **Behavior**:
>     - Implement literal types in `FieldType` and `LiteralValue` enums.
>     - Add coercion logic for literals in `coerce_literal.rs`.
>     - Update `to_baml_arg.rs` to handle literal values.
>     - Modify `json_schema.rs` to support literal types in JSON schema.
>     - Update `repr.rs` to include literal types in representation.
>   - **Validation**:
>     - Add validation for literal types in `types.rs`.
>     - Introduce tests for literals in `test_literals.rs`.
>   - **Code Generation**:
>     - Update Python, Ruby, and TypeScript code generation to support literals.
>     - Modify OpenAPI generation to include literal types.
>   - **Misc**:
>     - Add new test files for validation of literal types.
>     - Update existing test files to include literal type scenarios.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for a14b3082d3cee8395dc8ef05b31a82858bd35c46. It will automatically update as commits are pushed.</sup>


integ-tests/openapi/baml_client/openapi.yaml
```diff
@@ -1366,11 +1366,16 @@ components:
             properties:
               email:
                 type: string
+              reason:
+                oneOf:
+                - type: string
+                - type: string
```
sxlijin, line 1372: still need to implement union flattening

antoniosarosi: I did actually but we discussed it yesterday during standup and we concluded that we should't resolve types ourselves and let each language handle such cases themselves. So for now Baml types map one to one to the target language.

sxlijin: So I wasn't there for that conversation, so maybe I'm missing something, but here's the result of changing `Resume.name` from `type: string` to `oneOf: [{type: string}, {type: string}]` in golang:

`type: string`:

```
type Resume struct {
        Name string `json:"name"`
        Email string `json:"email"`
        Phone string `json:"phone"`
        Experience []Education `json:"experience"`
        Education []string `json:"education"`
        Skills []string `json:"skills"`
}
```

`oneOf: [{type: string}, {type: string}]`:

```
type Resume struct {
        Name ResumeName `json:"name"`
        Email string `json:"email"`
        Phone string `json:"phone"`
        Experience []Education `json:"experience"`
        Education []string `json:"education"`
        Skills []string `json:"skills"`
}

// ResumeName - struct for ResumeName
type ResumeName struct {
        String *string
}
```

so accessing `resume_obj.Name` now becomes `*resume_obj.Name.string` (and in the case of Go,  you don't even get a guarantee that the string is non-nil!) - this is something we're OK with?

hellovai: yea we're ok with this for now.

The way to resolve this is to resolve literals in the inner layer to a: `Literal['stringA', 'stringB']` isntead of `Literal['stringA'] | Literal['stringB']`.

The reason is type ordering can matter and may cause subtle bugs.

curious, are we sure literals shouldn't be enums? We agreed here its just not as clean? Should we include some docs or something?

engine/baml-lib/jsonish/src/deserializer/coercer/coerce_literal.rs
```diff
@@ -0,0 +1,85 @@
...
+                return Err(ctx.error_unexpected_null(target));
+            }
+            Some(v) => v,
+        };
+
+        match literal {
+            LiteralValue::Int(literal_int) => {
+                let BamlValueWithFlags::Int(coerced_int) = coerce_int(ctx, target, Some(value))?
+                else {
+                    unreachable!("coerce_int returned a non-integer value");
```
hellovai, line 53: is this actually unreachable?

antoniosarosi: Technically it is, `coerce_int` always returns `Ok(BamlValueWithFlags::Int)` or `Err` which is already propagated upwards with `?`. But I agree it's probably a little risky and it could be turned into an error instead.

integ-tests/openapi/baml_client/openapi.yaml
antoniosarosi: > The reason is type ordering can matter and may cause subtle bugs.

Do you have any concrete example where that is the case?

> curious, are we sure literals shouldn't be enums? We agreed here its just not as clean?

Unions of string literals can be represented as enums, it adds some complexity but most of the code is already there. I don't think base types are cleaner than enums, neither base types or enums are a one to one representation of literals, so it probably doesn't matter which one we choose in terms of how "clean" it is.

> Should we include some docs or something?

Yes, I can write the docs in a separate PR.

PR #979, opened 2024-09-22
docs: Fix Python dynamic types example

lorenzoh:
Stumbled upon this broken example in the documentation. This:

- removes the `const` declaration 
- uses `snake_case` variable names
- imports the async client so that the example runs without modifications

Related but not changed here: if copied as is (with the fixes), the example will throw a `BamlValidationError` since the input `"some user info"` doesn't contain the needed information. Replacing that with a sentence like "Bill (16) plays soccer in Germany" will lead to a valid output. Happy to pull up a patch for that as well if you want.


aaronvg:
thanks for the contribution!

PR #1009, opened 2024-10-04
feat(vertex): allow specifying creds as JSON object

sxlijin:
Allow `credentials` to be a JSON object corresponding to service account credentials downloaded from GCP.

Concretely, `credentials` can now be either a path to a file, a string containing credentials JSON, or a JSON object itself. Only document the first and the last options; we don't really want users passing in a string containing JSON, but it's easiest to do this for internal testing.


hellovai (changes requested):
add integ test plz!

hellovai (approved):

PR #1091, opened 2024-10-24
Constraints cause a runtime bug when functions are unioned with media types

hellovai:
The bug appears to be a bad propagation of attributes on `media | string`

Repro:
cd ./engine/baml-runtime/tests
cargo test --features internal -- internal_tests::test_with_image_union --exact --show-output

Bug:
```
Error: receipt: Failed to evaluate assert: Could not unify Media with Union([Primitive(Media(Image)), Primitive(String)])
```


imalsogreg:
🙏 Thank you for the minimal test case. Very helpful! Fixed by #1096

Issue #1085, opened 2024-10-22
Parser Bug: Enums with shared substrings


hellovai:
When the LLM returns something more "approximate" our parsing algorithm can handle substring'ed aliases better.

For example:

```rust
enum Foo {
   A @alias("car")
   B @alias("car-2")
}
```

Raw LLM response.

```
The answer is not car or car-2!
```

We will currently parse this as`Foo.A` cause technically we find two instances of "car" and one of "car-2".

Fix:
https://github.com/BoundaryML/baml/blob/cd6b141020ec8dfd2514c82ffffaebc5678a025b/engine/baml-lib/jsonish/src/deserializer/coercer/match_string.rs

Change `string_match_strategy` to account for substrings that are counted multiple times to only favor the longest possible one.

hellovai:
For more context the proposed solution is something like:

1. take all matches (same as before)
2. Find all matches that overlap with other matches (new)
3. Reduce to the most strict set (new)
4. Run tie breaker (same as before)

In the example:

All matches:
`car` x2
`car-1` x1

Reduced to:
`car` x1
`car-1` x1

Tie breaker fails to disambiguate.

hellovai:
added a branch with a unit test: https://github.com/BoundaryML/baml/pull/1088

Anyone is free to work on it from here:

```bash
cd $REPO/engine/baml-lib/jsonish/src
RUST_LOG=trace cargo test test_numerical_enum
```

PR #1098, opened 2024-10-25
[Fix] Improve Enum Substring Alias Handling

miguelcsx:
## Summary:

This pull request introduces improvements to the handling of enum substring aliases in the deserializer. The update modifies the string match strategy to favor the longest substring match, enhancing the accuracy and robustness of enum alias recognition.

## Changes

- Updated the string match strategy in the deserializer to prioritize the longest substring match for enum aliases.

## Benefits

- Improved handling of enum aliases, reducing potential mismatches.
- 
- Enhances overall functionality and user experience when working with enums.
    
Related Issue: #1085


engine/baml-lib/jsonish/src/deserializer/coercer/match_string.rs
```diff
@@ -142,57 +142,76 @@ fn string_match_strategy<'c>(
...
-                    .iter()
-                    .map(|(count, _, candidate)| ((*count) as usize, candidate.to_string()))
-                    .collect(),
-            ));
+        // Find all variants with the same count
+        let max_count = count;
+        let ties: Vec<_> = variant_counts
+            .iter()
+            .filter(|(c, _)| c == max_count)
+            .map(|(count, variant)| (*count as usize, variant.to_string()))
```
antoniosarosi: Judging by the code [here](https://github.com/BoundaryML/baml/blob/23e590b0b2fdb51f80e7eced769baabd12b3be22/engine/baml-lib/jsonish/src/deserializer/deserialize_flags.rs#L178-L183) it looks like the first element of the tuple should be the index of the match, not its count:

```rust
Flag::StrMatchOneFromMany(values) => {
    write!(f, "Enum one from many: ")?;
    for (idx, value) in values {
        writeln!(f, "Item {}: {}", idx, value)?;
    }
}
```

I'm fixing this right now.

antoniosarosi: I guess that code is wrong and `idx` is actually `count`, changed it in [this commit](https://github.com/BoundaryML/baml/pull/1098/commits/555fd242fd673e74d0f35c77db2bffd70ab8bdd9). Maybe we should add both the count and all the occurrences to every match so we can return a nicer error in [`deserialize_flags.rs`](https://github.com/BoundaryML/baml/blob/23e590b0b2fdb51f80e7eced769baabd12b3be22/engine/baml-lib/jsonish/src/deserializer/deserialize_flags.rs#L178-L183), but for now this PR looks good to me!

antoniosarosi (approved):
LGTM!

PR #1163, opened 2024-11-12
Implement Type Aliases

antoniosarosi:
Implementation of TS-like type aliases (structural recursion allowed through maps and lists).

Basic aliases:

```baml
type A = int
type B = float
type C = A | B
```

Recursive aliases:

```baml
type RecursiveMap = map<string, RecursiveMap>

type A = A[]

type JsonValue = int | float | bool | string | null | JsonValue[] | map<string, JsonValue> 
```

See also #1207


imalsogreg:
Something strange going on in the playground. These two errors only appear once I add a type alias. The name I chose, `Foo`, is of course maybe problematic (if it overlaps with some other `Foo` in our big `baml_src`). But I see the same behavior when I use a unique name for the type alias.

Screenshot 2024-12-09 at 6 32 35 PM

antoniosarosi:
@imalsogreg Fixed!

imalsogreg:
This is just a thought for later, definitely not something i would suggest implementing now:

Some time in the future, I think `FieldType::Alias` could be removed from the `FieldType` definition, and type aliases should be a form of syntactic sugar, rather than a core part of the IR.

Pros/Cons to this proposal:
 - Pro: `FieldType::Alias` isn't necessary in the IR because a type alias is always just a user-facing name for a type, not a semantically distinct concept. (When `type Foo = int`, the compiler doesn't distinguish between the two - the difference is only present at a surface syntax/UI level).
 - Pro: Having two representations for the same entity causes duplicate code for many non-type-alias related calculations. I.e. many IR operations need to account for `FieldType::Alias` and all its intricacies, when in fact at this compiler stage only the `resolved` field is relevant.
 - Con: `FieldType::Alias` is already implemented and well understood, we would be changing it when it's not broken, under this proposal.
 - Con: Maybe it's not obvious which compiler pass should be responsible for desugaring type aliases.
 - Con: We would need to carry metadata (maybe on IR nodes, maybe on types) to track the syntax provenance of a type, in order to use the correct spelling (raw type spelling or alias spelling) in different contexts. It might not be obvious where we should store that metadata. However, `FieldType` probably deserves the same `FieldTypeWithMeta<T>` treatment that we did for `BamlValue`.

Again - this is just some thought about the future, definitely not a change we should make now. But I'm curious @antoniosarosi what you think about it.

antoniosarosi:
Thought about removing `FieldType::Alias` from the IR yesterday, the only reason I added that was to report errors using either the alias name or target/resolved type depending on the context. For this particular PR it could be removed but for the cycles PR it's still necessary to store the alias name somewhere.  Still thinking about how to refactor it @imalsogreg.

PR #1211, opened 2024-12-04
BAML formatter implementation (Partial)

dymk:
Adds a `baml fmt` subcommand for formatting files.

Formatter is a WIP, and writes out files from `x.baml` to `x.formatted.baml` to avoid clobbering the original file.

Example syntax it can format in its current state - 

```
========= before ==========
class Foo {
}
------------ > ------------
class Foo {}
========== after ==========

========= before ==========
class Foo { field1 string }
------------ > ------------
class Foo {
    field1 string
}
========== after ==========

========= before ==========
class Foo {

    field1 string
}
------------ > ------------
class Foo {
    field1 string
}
========== after ==========

========= before ==========
class Foo {
    field1   string|int
}
------------ > ------------
class Foo {
    field1 string | int
}
========== after ==========

```

----

> [!IMPORTANT]
> Add `baml format` subcommand to CLI for formatting BAML files, outputting to a new ".formatted.baml" file.
> 
>   - **Behavior**:
>     - Adds `baml format` subcommand to format BAML files, outputting to `x.formatted.baml`.
>     - Formats class definitions and field types with specific syntax adjustments.
>   - **Implementation**:
>     - Implements `format_schema()` in `formatter/mod.rs` to handle BAML syntax.
>     - Adds `FormatArgs` in `format.rs` to manage file paths and execution.
>   - **Dependencies**:
>     - Adds `pretty` and `unindent` dependencies in `Cargo.toml` for formatting support.
>   - **CLI Integration**:
>     - Updates `commands.rs` to include `Format` command.
>     - Adds `format.rs` to handle the format command logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for 78c8476fce7fd75abd6ad3b131194ee1d467c21d. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/engine/cli/src/commands.rs
+++ b/engine/cli/src/commands.rs
@@ -35,6 +35,9 @@ pub(crate) enum Commands {
 
     #[command(about = "Deploy a BAML project to Boundary Cloud")]
     Deploy(crate::deploy::DeployArgs),
+
+    #[command(about = "Format BAML source files", name = "fmt", hide = true)]
+    Format(crate::format::FormatArgs),
 }
 
 impl RuntimeCli {
@@ -64,6 +67,7 @@ impl RuntimeCli {
                 args.from = BamlRuntime::parse_baml_src_path(&args.from)?;
                 t.block_on(async { args.run_async().await })
             }
+            Commands::Format(args) => args.run(),
         }
     }
 }
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -121,6 +121,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
+name = "arrayvec"
+version = "0.5.2"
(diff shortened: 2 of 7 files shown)


engine/cli/src/commands.rs
```diff
@@ -35,6 +35,9 @@ pub(crate) enum Commands {
 
     #[command(about = "Deploy a BAML project to Boundary Cloud")]
     Deploy(crate::deploy::DeployArgs),
+
+    #[command(about = "Format BAML source files")]
```
sxlijin:
```suggestion
    #[command(about = "Format BAML source files", name = "fmt", hide = true)]
```

I think we can merge this code as long as we hide this command from users

sxlijin:
amazing! glad to see that it was possible to get a proof-of-concept going so quickly, thanks so much!

dymk:
Made a quick change to not fail on unhandled pest rules when ran via `baml-cli fmt` so it can be used on files without the formatter being fully complete.

PR #1224, opened 2024-12-06
fix: Support parsing primitive values from single-key objects

### LINKED ISSUE #1141: Consider fixing parsing of literals from objects
Sometimes when an LLM responds to a primitive value, it can use an object to contain its response. We can benefit from parsing objects of singular keys as the values themslves and drop they key. we already do this in some places.

This new test fails:

```rust
// file: engine/baml-lib/jsonish/src/tests/test_literals.rs

test_deserializer!(
    test_union_literal_with_multiple_types_from_object,
    EMPTY_FILE,
    r#"{
  "status": 1
}"#,
    FieldType::Union(vec![
        FieldType::Literal(LiteralValue::Int(1)),
        FieldType::Literal(LiteralValue::Bool(true)),
        FieldType::Literal(LiteralValue::String("THREE".into())),
    ]),
    1
);
```

afyef:
# Description
Fixes #1141

This PR adds support for parsing primitive values from single-key objects in unions. When an LLM responds with an object-wrapped primitive (e.g., `{"status": 1}`), we now correctly extract and validate the inner value.

## Changes
- Add support for extracting primitive values (numbers, booleans, strings) from single-key objects
- Track the extracted key name using `ImpliedKey` flag
- Reject invalid cases:
  - Nested objects (e.g., `{"status": {"code": 1}}`)
  - Arrays (e.g., `{"values": [1]}`)
  - Multi-key objects

## Test Plan
Added test cases covering:
- Basic object extraction: `{"status": 1}` → `1`
- Multiple primitive types in union
- Invalid cases (nested objects, arrays)

All tests are passing.

----

> [!IMPORTANT]
> Adds support for parsing primitive values from single-key objects in unions, with tests for valid and invalid cases.
> 
>   - **Behavior**:
>     - Supports parsing primitive values from single-key objects in unions in `coerce_literal.rs`.
>     - Extracts and validates inner values if they are numbers, booleans, or strings.
>     - Rejects nested objects, arrays, and multi-key objects.
>   - **Flags**:
>     - Uses `ImpliedKey` flag to track extracted key names.
>   - **Tests**:
>     - Added tests in `test_literals.rs` for basic object extraction, multiple primitive types, and invalid cases (nested objects, arrays).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for b4bf23f51f42dc9478dd71ba4bc0340bb5f06340. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/engine/baml-lib/jsonish/src/deserializer/coercer/coerce_literal.rs
+++ b/engine/baml-lib/jsonish/src/deserializer/coercer/coerce_literal.rs
@@ -7,6 +7,7 @@ use internal_baml_core::ir::FieldType;
 use crate::{
     deserializer::{
         coercer::{coerce_primitive::coerce_bool, match_string::match_string, TypeCoercer},
+        deserialize_flags::{DeserializerConditions, Flag},
         types::BamlValueWithFlags,
     },
     jsonish,
@@ -45,6 +46,22 @@ impl TypeCoercer for LiteralValue {
             }
             Some(v) => v,
         };
+
+        // If we get an object with a single key-value pair, try to extract the value
+        if let jsonish::Value::Object(obj) = value {
+            if obj.len() == 1 {
+                let (key, inner_value) = obj.iter().next().unwrap();
+                // only extract value if it's a primitive (not an object or array, hoping to god its fixed)
+                match inner_value {
+                    jsonish::Value::Number(_) | jsonish::Value::Boolean(_) | jsonish::Value::String(_) => {
+                        let mut result = self.coerce(ctx, target, Some(inner_value))?;
+                        result.add_flag(Flag::ImpliedKey(key.clone()));
+                        return Ok(result);
+                    }
+                    _ => {}
+                }
+            }
+        }
 
         match literal {
             LiteralValue::Int(literal_int) => {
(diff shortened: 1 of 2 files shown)


engine/baml-lib/jsonish/src/deserializer/coercer/coerce_literal.rs
```diff
@@ -46,6 +47,22 @@ impl TypeCoercer for LiteralValue {
...
+        // If we get an object with a single key-value pair, try to extract the value
+        if let jsonish::Value::Object(obj) = value {
+            if obj.len() == 1 {
+                let (key, inner_value) = obj.iter().next().unwrap();
+                // only extract value if it's a primitive (not an object or array, hoping to god its fixed)
+                match inner_value {
+                    jsonish::Value::Number(_) | jsonish::Value::Boolean(_) | jsonish::Value::String(_) => {
+                        let mut result = self.coerce(ctx, target, Some(inner_value))?;
+                        result.add_flag(Flag::ImpliedKey(key.clone()));
+                        return Ok(result);
```
hellovai, line 59: we should use a different flag here: `ObjectToPrimitive`  instead.

implied key actually means the key doesn't exist, but we were able to assume it does. Sorry no docs about that.

hellovai (changes requested):

hellovai (approved):
LGTM!

PR #1238, opened 2024-12-12
feat(serve): Add graceful Ctrl+C handling with exit message

afyef:
# Graceful Ctrl+C Handling

Implements reliable Ctrl+C handling for BAML's Python interface with a clean shutdown message. This fixes issues with signal handling across the Python/Rust boundary and improves user experience when stopping BAML processes.

## Changes
- Added custom signal handling using the `ctrlc` crate
- Implemented clean "Shutting Down BAML..." message on interrupt
- Bypassed Python's signal handling to ensure reliable behavior
- Added proper exit code (130) following Unix conventions
- Added detailed documentation explaining the implementation

## Testing
- Tested Python CLI server start/stop
- Manual Ctrl+C testing in various scenarios
- Verified clean shutdown message display
- Tested cross-platform signal handling behavior

## Technical Details
- Uses `ctrlc` crate for reliable cross-platform signal handling
- Implements channel-based communication between signal handler and main thread
- Follows Unix conventions for interrupt handling (exit code 130)
- Properly documented implementation rationale and technical decisions

## Notes
- While this is a workaround, it provides reliable interrupt handling without requiring major architectural changes
- The implementation bypasses Python's signal handling entirely to avoid conflicts
- Future improvements could involve deeper integration with BAML's architecture

----

> [!IMPORTANT]
> Enhances BAML's Python interface with reliable Ctrl+C handling and clean shutdown message using the `ctrlc` crate.
> 
>   - **Behavior**:
>     - Implements custom signal handling to bypass Python's signal handlers
>     - Displays "Shutting Down BAML..." message on interrupt
>     - Ensures proper process termination with standard exit codes
>   - **Implementation**:
>     - Uses `ctrlc` crate for cross-platform signal handling
>     - Implements channel-based communication for safe signal handling
>     - Follows Unix conventions with exit code 130 for SIGINT
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for the latest commit. It will automatically update as commits are pushed.</sup>


sxlijin:
Started a discussion thread in https://discord.com/channels/1119368998161752075/1316923453814734959 - I have some questions about how this plays with pyo3 and different architectures (it sounds like we handle this fine in N-API world?)

Thanks for digging into this!

engine/baml-runtime/src/cli/serve/mod.rs
```diff
@@ -340,7 +330,29 @@ Tip: test that the server is up using `curl http://localhost:{}/_debug/ping`
...
+                    // Small delay to ensure message is visible
+                    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+                }
+                Err(err) => {
+                    log::error!("Error handling Ctrl+C: {}", err);
+                }
+            }
+        };
+
+        // Set ctrl-c handler to ignore so Python doesn't intercept it, otherwise python doesnt like it
```
sxlijin: As discussed in Discord, let's try moving this into [`invoke_runtime_cli`](https://github.com/BoundaryML/baml/blob/canary/engine/language_client_python/src/lib.rs#L12-L13)

Everything in `baml-runtime` is shared across all 4 builds of BAML (Python, Node, Ruby, and the internal-only Rust CLI) so I'd rather not muck with default signal handling behavior unless we have to - e.g. right now it's really easy for someone to misunderstand how the current code works, because it's easy to read it as setting up a tokio ctrl-c handler and then uninstalling all signal handlers (even though it's actually the other way around due to how Rust async works)

engine/baml-runtime/src/cli/serve/mod.rs
```diff
@@ -313,21 +317,7 @@ impl Server {
...
-        // if we try to handle this ourselves, see
-        // https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html#caveats
-        //
-        // Namely- we need to ensure resilient delivery of Ctrl-C to everything, and I
-        // suspect we need to do a bit of work to ensure that we handle that bookkeeping
-        // correctly. Shutting down the BAML runtime, tokio runtime, _and_ axum is not
-        // super straightforward, because I don't know how much is handled for us
-        // out of the box.
-        //
-        // .with_graceful_shutdown(signal::ctrl_c());
```
sxlijin: Can we keep these comments around in this file? Even if we push the libc signal handler stuff to language_client_python/src/lib.rs. Mostly just because this is because a good reminder for myself about how careful we should be about touching signal handling code here.

sxlijin (approved):
This is clever! Nice work with this

We could use a oneshot channel for the signal but I really like this - on the one hand spawning an additional thread feels overkill, but it's also a neat way to avoid wiring in `Python::check_signals` everywhere

sxlijin:
Sorry for the delay! I was in a GitHub beta for a new UI and couldn't figure out how to trigger workflows on your PR (maintainers need to explicitly approve PRs from new contributors)

sxlijin:
Hmm - looks like I can't push to your branch to update `Cargo.lock`. Can you commit that change?

afyef:
@sxlijin  Committed.

PR #1249, opened 2024-12-16
feat(jsonish): Add support for curly quotes in JSON parsing

### LINKED ISSUE #1074: Parser Improvement: Support ” along with "
LLMs will sometimes output something like this:
```
Sure! Here is a made-up JSON blob that matches the schema you provided:
\```
{
  "prop1": "example",
  "prop2": {
    "prop1": "value1",
    "prop2": "value2”,
    "inner": {
      "prop2": 42,
      "prop3": 3.14,
    }
  }
}
\```
```

where it would parse if `”` was `"`. We can likely correct for this.

afyef:
## Description
LLMs will sometimes output JSON with curly quotes (U+201C and U+201D) instead of straight quotes (U+0022). This PR adds automatic normalization of these quotes during parsing, making BAML more robust when handling LLM outputs.

## Changes
- Added `normalize_quotes()` function that converts:
  - Left double quotation mark (U+201C) → Basic quotation mark (U+0022)
  - Right double quotation mark (U+201D) → Basic quotation mark (U+0022)
- Applied normalization before all parsing attempts:
  - Standard JSON parsing (serde_json)
  - Markdown-embedded JSON
  - Multiple JSON objects
  - JSON fixing parser
- Preserved original string for error messages and output

Fixes #1074

----

> [!IMPORTANT]
> Add support for curly quotes in JSON parsing and implement custom SIGINT handling for graceful shutdowns.
> 
>   - **JSON Parsing**:
>     - Added `normalize_quotes()` in `entry.rs` to convert curly quotes (U+201C, U+201D) to standard quotes (U+0022).
>     - Applied normalization in `parse()` before parsing JSON, Markdown-embedded JSON, multiple JSON objects, and JSON fixing parser.
>   - **Signal Handling**:
>     - Implemented custom SIGINT handling in `lib.rs` using `ctrlc` crate to ensure graceful shutdown across Python/Rust boundary.
>     - Added dependencies `libc`, `ctrlc`, and `tokio-util` in `Cargo.toml` for signal handling support.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for 0513ae74ac18e9a8b19151ab7242dbe4369c166b. It will automatically update as commits are pushed.</sup>


imalsogreg:
@revidious Thanks for this PR, it looks great!

Regarding curly quotes, do you think it's better to do normalization as a preprocessing step, as you have done in the PR? Or should we add curly quotes as another type of quotation mark, alongside the parsers for quoted string, single-quoted string, and backtick-string, here: https://github.com/BoundaryML/baml/blob/canary/engine/baml-lib/jsonish/src/jsonish/parser/fixing_parser/json_parse_state.rs#L540 ?

hellovai (changes requested):
@revidious regarding gregs comment, I would highly recommend doing it with that approach rather than a preprocessing step. Additoinally use the `AnyOf` capability in the preprocessor as well, since it could very intentionally be an actual curly quote as intended by the user.

PR #1258, opened 2024-12-17
feat(python): Implement TypeBuilder __str__ and __repr__ for better debugging

### LINKED ISSUE #1253: TypeBuilder `_repr_` should be implemented in python for better debugging
When users print out the Typebuilder instance in python, they see a rust pointer instead of the actual type information.

afyef:
## Description
This PR adds a human-readable string representation to the TypeBuilder class to improve debugging experience. When users print a TypeBuilder instance, they now see a clear list of available classes and enums instead of a Rust pointer.

## Changes
- Added `__str__` method to format TypeBuilder contents
- Made `__repr__` use the same implementation for consistency
- Added comprehensive docstrings explaining the format

## Examples
```python
tb = TypeBuilder()
print(tb)  # TypeBuilder(empty)

tb.add_class('User')
tb.add_enum('Status')
print(tb)  # TypeBuilder(Classes: ['User'], Enums: ['Status'])
```

## Testing
Tested with various combinations:
- Empty TypeBuilder
- TypeBuilder with only classes
- TypeBuilder with only enums
- TypeBuilder with both classes and enums
- Dynamic addition of types

Fixes #1253

----

> [!IMPORTANT]
> Enhance `TypeBuilder` in Python with `__str__` and `__repr__` for better debugging and update Rust code to handle optional types in JSON schema and parsing.
> 
>   - **Python `TypeBuilder`**:
>     - Added `__str__` and `__repr__` methods to `TypeBuilder` for human-readable string representation.
>     - Formats include empty, classes only, enums only, and both classes and enums.
>   - **Rust JSON Schema**:
>     - Updated `FieldType::List` and `FieldType::Map` in `json_schema.rs` to handle optional types, allowing `null` values.
>     - Added default `null` for optional arrays and maps.
>   - **Parsing Enhancements**:
>     - Updated `datamodel.pest` and `parse_types.rs` to support optional arrays and maps with `?` suffix.
>     - Added tests for optional types in `parse_types.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for aaaed1d76ad7ae529d2691c7a33dd2e6a5c6a69a. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/engine/baml-lib/baml-core/src/ir/json_schema.rs
+++ b/engine/baml-lib/baml-core/src/ir/json_schema.rs
@@ -186,16 +186,40 @@ impl WithJsonSchema for FieldType {
                     "required": ["url"],
                 }),
             },
-            FieldType::List(item) => json!({
-                "type": "array",
-                "items": (*item).json_schema()
-            }),
-            FieldType::Map(_k, v) => json!({
-                "type": "object",
-                "additionalProperties": {
-                    "type": v.json_schema(),
+            // Handle list types (arrays) with optional support
+            // For example: string[]? generates a schema that allows both array and null
+            FieldType::List(item) => {
+                let mut schema = json!({
+                    "type": "array",
+                    "items": (*item).json_schema()
+                });
+                // If the list itself is optional (marked with ?),
+                // modify the schema to accept either an array or null
+                if self.is_optional() {
+                    schema["type"] = json!(["array", "null"]);
+                    // Add default null value for optional arrays
+                    schema["default"] = serde_json::Value::Null;
                 }
-            }),
+                schema
+            },
+            // Handle map types with optional support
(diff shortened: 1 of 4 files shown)


aaronvg:
your diff seems to include the code for a different PR -- it may be better to isolate these

aaronvg:
mind also outputting each of the properties in a class, or the values in an enum?

There is also aliases and descriptions attached that we could print out

afyef:
> your diff seems to include the code for a different PR -- it may be better to isolate these

yeah, my bad will get that checked

afyef:
> mind also outputting each of the properties in a class, or the values in an enum?
> 
> There is also aliases and descriptions attached that we could print out

will look into it

PR #1346, opened 2025-01-17
NextJS Codegen

seawatts:
This adds native integration for next.js code generation.


engine/language_client_codegen/baml_src/baml_client/__init__.py
aaronvg, line 1: should this baml_client be here? in language_client_codegen?

integ-tests/react/src/app/api/baml/[baml_function]/route.ts
```diff
@@ -0,0 +1,88 @@
+import { NextRequest, NextResponse } from 'next/server'
```
Comment (aaronvg) [inline integ-tests/react/src/app/api/baml/[baml_function]/route.ts]: is this a file that users need to write / copy paste? Or does the generator try to put this route here?

integ-tests/react/src/app/actions.ts
```diff
@@ -0,0 +1,33 @@
+
```
aaronvg: is this also written by the generator? Outside of baml_client and directly into your app/ ?

engine/language_client_codegen/baml_src/baml_client/__init__.py
seawatts: Good catch, not sure how that got put in there.

integ-tests/react/src/app/api/baml/[baml_function]/route.ts
Comment (seawatts) [reply integ-tests/react/src/app/api/baml/[baml_function]/route.ts]: Oops, we aren't using this. I forgot to delete it.

integ-tests/react/src/app/actions.ts
seawatts: Also not being used, deleting.

engine/language_client_codegen/src/openapi.rs
```diff
@@ -326,22 +326,26 @@ impl<'ir> TryFrom<(&'ir IntermediateRepr, &'_ crate::GeneratorArgs)> for OpenApi
...
+        let mut paths: Vec<OpenApiMethodDef> = ir
+            .walk_functions()
+            .map(|f| {
+                f.try_into().context(format!(
+                    "Failed to convert BAML function {} to OpenAPI method",
+                    f.item.elem.name()
+                ))
+            })
+            .collect::<Result<_>>()?;
+        paths.sort_by(|a, b| a.function_name.cmp(b.function_name));
```
imalsogreg, line 338: Thank you :)

seawatts: 🥳 

I was having an issue that I’m surprised we hadn’t seen before, where the open API and typescript generators were producing slightly different list ordering for types, functions, classes, etc. It was only intermittently, though. I added some updates to the rust generators to sort the lists by name beforehand so it will be consistent every time. I haven’t updated Python or Ruby yet, though.

engine/language_client_codegen/src/typescript/mod.rs
```diff
@@ -1,20 +1,82 @@
...
+        None,
+        React
+    }
+
+    impl TypescriptFramework {
+        pub fn from_generator_type(output_type: Option<GeneratorOutputType>) -> Self {
+            match output_type {
+                Some(GeneratorOutputType::TypescriptReact) => Self::React,
+                Some(GeneratorOutputType::Typescript) | None => Self::None,
+                Some(_) => panic!("Invalid generator type for TypeScript framework"),
```
imalsogreg: Consider returning `Option<Self>`, with `None` for the wildcard case. If `Some(_)` is truly unreachable, the callsite is a clearer place to indicate that.

It's also better to explicitly match every possible `GeneratorOutputType`. Wildcard patterns (`_`) make it easy to accidentally add a variant, then forget to update the `match`. e.g. if we added `GeneratorOutputType::TypescriptSveltekit` to the `GeneratorOutputType`, `from_generator_type` would silently start to panic. But if we never use a wildcard pattern, we get a compiler error that points us to the function we need to update.

engine/language_client_codegen/src/typescript/templates/types.ts.j2
```diff
@@ -1,4 +1,8 @@
-import { Image } from "@boundaryml/baml"
+import type { Image, Audio } from "@boundaryml/baml"
+
+export type RecursivePartialNull<T> = T extends object
```
imalsogreg, line 9: Do we still use/need `RecursivePartialNull`?

engine/language_client_typescript/artifacts/async_context_vars.d.ts.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"async_context_vars.d.ts","sourceRoot":"","sources":["../typescript_src/async_context_vars.ts"],"names":[],"mappings":"AAAA,OAAO,EAAE,QAAQ,EAAE,qBAAqB,EAAE,WAAW,EAAE,YAAY,EAAE,MAAM,UAAU,CAAA;AAGrF,qBAAa,cAAc;IACzB,OAAO,CAAC,EAAE,CAAa;IACvB,OAAO,CAAC,GAAG,CAA0C;gBAEzC,EAAE,EAAE,WAAW;IAS3B,WAAW,IAAI,OAAO;IAYtB,KAAK,IAAI,IAAI;IAKb,UAAU,CAAC,IAAI,EAAE,MAAM,CAAC,MAAM,EAAE,MAAM,CAAC,GAAG,IAAI;IAK9C,YAAY,IAAI,qBAAqB;IASrC,UAAU,CAAC,IAAI,EAAE,MAAM,EAAE,IAAI,EAAE,MAAM,CAAC,MAAM,EAAE,GAAG,CAAC,GAAG,CAAC,qBAAqB,EAAE,QAAQ,CAAC;IAKtF,QAAQ,CAAC,IAAI,EAAE,QAAQ,EAAE,QAAQ,EAAE,GAAG,GAAG,IAAI;IAa7C,KAAK,IAAI,IAAI;IAIb,UAAU,CAAC,QAAQ,EAAE,CAAC,CAAC,KAAK,EAAE,YAAY,KAAK,IAAI,CAAC,GAAG,SAAS,GAAG,IAAI;IAYvE,WAAW,CAAC,UAAU,EAAE,CAAC,SAAS,CAAC,GAAG,IAAI,EAAE,GAAG,EAAE,KAAK,UAAU,EAAE,IAAI,EAAE,MAAM,EAAE,IAAI,EAAE,CAAC,GAAG,CAAC;IAuB3F,YAAY,CAAC,UAAU,EAAE,CAAC,SAAS,CAAC,GAAG,IAAI,EAAE,GAAG,EAAE,KAAK,OAAO,CAAC,UAAU,CAAC,EAAE,IAAI,EAAE,MAAM,EAAE,IAAI,EAAE,CAAC,GAAG,CAAC;CAuBtG"}
```
imalsogreg: Did we mean to check in this and `async_context_vars.d.ts`?  This file looks like it wast meant to be gitignored?

engine/language_client_typescript/artifacts/native.d.ts
```diff
@@ -0,0 +1,133 @@
+/* auto-generated by NAPI-RS */
```
imalsogreg: Looks like lots of files under `artifacts` get checked in. Did we mean to exclude them from VCS?

engine/language_client_typescript/stream.js
```diff
@@ -57,5 +57,28 @@ class BamlStream {
         const final = await this.driveToCompletionInBg();
         return this.finalCoerce(final.parsed(false));
     }
+    /**
+     * Converts the BAML stream to a Next.js compatible stream.
+     * This is used for server-side streaming in Next.js API routes and Server Actions.
+     */
+    toStreamable() {
+        const stream = this;
+        return new ReadableStream({
```
imalsogreg: Where does `ReadableStream` come from? Does this break users who don't have next or react installed?

fern/.cursorrules
```diff
@@ -0,0 +1,320 @@
+You are the world's best documentation writer, renowned for your clarity, precision, and engaging style. Every piece of documentation you produce is:
```
imalsogreg, line 1: Just making sure - we want to check this in?

fern/docs.yml
```diff
@@ -83,10 +83,36 @@ navigation:
...
-
-          - page: NextJS
+      - section: Framework Integration
+        icon: fa-solid fa-plug
+        contents:
+          - section: Next.js
             icon: fa-brands fa-react
-            path: 01-guide/08-integrations/nextjs.mdx
+            contents:
+              # - page: Overview
```
imalsogreg: Commented code probably deletable.

integ-tests/README.md
```diff
@@ -172,7 +172,13 @@ b. **Using .env File (Recommended for open source contributors)**
       ```bash
       # Build the debug version of the client
       cd typescript && pnpm run build:debug
+      ```
+
+      <Note>
+      For local development, use `pnpm run build:debug` which is much faster. The full `pnpm run build` can take several minutes as it creates an optimized production build.
```
imalsogreg, line 178: Thank you :)

imalsogreg (approved):
Left a couple comments, none of them critical.

Looks pretty great!

Note - I haven't run integ tests myself.

I did run `pnpm generate` but didn't carefully examine the results.

Happy to do the above things soon, just wanted to give my initial 👍 in case you are trying to merge quickly.

engine/language_client_typescript/artifacts/native.d.ts
seawatts: Ah ha, yes we can delete these now. It gets auto generated during build time. This allows us to do debugging locally a bit easier

engine/language_client_codegen/src/typescript/templates/types.ts.j2
seawatts: This is still being exported for people who might have been importing this in their projects. We are not using it anywhere else in the generated code. We could add a `@depreacted` flag. Thoughts?

engine/language_client_typescript/artifacts/async_context_vars.d.ts.map
seawatts: No, will delete

engine/language_client_typescript/stream.js
seawatts: That is built into Node, we don't need to import it from anywhere. It's available in the bowser environment and node environment. Not react specific

fern/.cursorrules
seawatts: I copied this from their fern llm.txt file. When we use cursor for files in that directory it can help cursor get the correct results. 

https://buildwithfern.com/learn/docs/developer-tools/llms-txt

Cursor just did an update where you can have multiple .cursorrules files in a repo and it takes precedence based on the file you talking about.

fern/docs.yml
seawatts: Done!

engine/language_client_codegen/src/typescript/mod.rs
seawatts: Fixed!

engine/language_client_codegen/src/typescript/templates/types.ts.j2
seawatts: Added one.

engine/language_client_codegen/src/typescript/templates/partial_types.ts.j2
```diff
@@ -1,6 +1,7 @@
-import { Image } from "@boundaryml/baml"
-
-import * as types from "./types"
+import type { Image, Audio } from "@boundaryml/baml"
+import type { Checked, Check } from "./types"
+import type { {% for t in types %} {{ t }}{% if !loop.last %}, {% endif %}{% endfor %} } from "./types"
```
sxlijin, line 3: I think this is a name collision, that users can't use "Check" for their own type name - or is that a decision we already made? (Same also applies to "Image" and "Audio", but I think that one is less egregious.)

cc @imalsogreg

engine/language_client_typescript/cli.js
```diff
@@ -5,6 +5,6 @@ if (require.main === module) {
     process.env.BAML_LOG = 'info'
   }
 
-  const baml = require('./native')
+  const baml = require('./artifacts/native')
```
sxlijin, line 8: Can we split this change out of this PR?

I tried to do something along these lines a while back and screwed up the TS release in a hard-to-understand way. We can try another PR to push this forward but I'm uncomfortable doing that in this PR.

fern/01-guide/08-frameworks/01-react-nextjs/01-app-router.mdx
```diff
@@ -0,0 +1,452 @@
...
+- Configure other options as needed for your project
+
+### Install Dependencies
+
+Next, install BAML and its dependencies:
+
+<CodeBlocks>
+```bash npm
+npm install @boundaryml/baml
+npm install @boundaryml/baml-nextjs-plugin
```
sxlijin:
```suggestion
npm install @boundaryml/baml @boundaryml/baml-nextjs-plugin
```

and I think we can do the same for the others too?

sxlijin (commented):
reading docs now

engine/language_client_codegen/src/typescript/templates/partial_types.ts.j2
seawatts: That's a good point. We have been doing that in other places it seems like. I'd say let's punt this to a new PR

engine/language_client_typescript/cli.js
seawatts: Okay, sounds good. I did it in order to do some testing locally where I need my integ-test/react project to reference the node wasm bindings and this is how I got it to work.

fern/01-guide/08-frameworks/01-react-nextjs/01-app-router.mdx
seawatts: Done

imalsogreg (approved):
🚀

PR #1465, opened 2025-02-17
New LSP server

imalsogreg:
The server is meant to replace much of the existing `server.ts`, which implements an LSP for VSCode. The LSP in this PR is a native binary that should work with any LSP client (VSCode, emacs, vim, etc).

Features
 - [x] Push diagnostics
 - [x] Respond to basic file requests
   - [x] did_open
   - [x] did_change
   - [ ] did_close
   - [x] did_change_watched_files
 - [ ] Request parity with `server.ts`
   - [x] Hover
   - [x] CodeLens
   - [x] Completion
   - [x] GotoDefinition
   - [x] Format
 - [x] Watch files in project
 - [x] Test that unsaved changes are reflected in playground tests
 - [x] Test Windows
 - [x] Test Linux
 - [x] Handle projects with nested directories
 - [x] Run generation
 - [ ] Manage configuration
 - [x] Toast notificitons on generation
 - [ ] Handle multiple compiler versions
 - [x] Test with neovim
 - [x] Test with emaics
 - [x] Test with VSCode
 - [x] Test baml_src/other/baml_src (it should swap workspaces the nearest baml_src parent).

----

> [!IMPORTANT]
> This PR introduces a new Rust-based LSP server with enhanced features and updates the build and release workflows accordingly.
> 
>   - **LSP Server**:
>     - Introduces a new Rust-based LSP server to replace the existing TypeScript server.
>     - Supports diagnostics, file watching, and LSP requests like `did_open`, `did_change`, `hover`, `completion`, and `goto_definition`.
>     - Implements `baml-lsp-types` and `language-server` crates.
>   - **Build and Release**:
>     - Updates GitHub Actions workflows for building and releasing the new LSP server.
>     - Adds `build-cli-release.reusable.yaml` and `release-cli.yaml` for CLI builds.
>     - Modifies `build-vscode-release.reusable.yaml` for VSCode extension packaging.
>   - **Miscellaneous**:
>     - Updates `Cargo.lock` and `Cargo.toml` for new dependencies.
>     - Adds test cases for the LSP server in `tests.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for d2fbcad51f2b6df7ff575ae9642f7f595f513d45. It will automatically update as commits are pushed.</sup>


.github/workflows/primary.yml
```diff
@@ -160,14 +165,85 @@ jobs:
...
         working-directory: engine
       - name: Upload CLI artifact
         uses: actions/upload-artifact@v4
         with:
           name: baml-cli-${{ matrix.target }}
-          path: engine/target/release/baml-cli*
\ No newline at end of file
+          path: engine/target/${{ matrix.target }}/release/baml-cli*
+
+  build-vscode:
```
aaronvg: we also have a build-vscode-release.reusable.yaml, which is the one that's called on release script.

Perhaps we need a build-cli.reusable.yaml?

`primary` i believe runs on each push to canary, while the `release.yaml` runs on the actual push of a 0.x.x tag.

We likely want the cli to be built on every push to canary (or even on every PR we run one architecture?), as well as on release.


Some other notes:
we probably need this env:
```
env:
  MACOSX_DEPLOYMENT_TARGET: "10.13"
```

and change all the hosts to ubuntu-20.04, and macos host from `latest` to `macos-14` to build on the oldest platform for max compatibility.

engine/.cargo/config.toml
```diff
@@ -4,6 +4,6 @@ rustflags = ["--cfg", "tracing_unstable"]
 
 # https://github.com/rust-lang/cargo/issues/8607
 [target.x86_64-unknown-linux-musl]
-rustflags = ["-C", "target-feature=-crt-static", "--cfg", "tracing_unstable"]
+rustflags = ["-C", "link-arg=-lm", "--cfg", "tracing_unstable"]
```
aaronvg: curious -- what's this do?

engine/language_server/src/baml_project/watch.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    pub const fn is_deleted(&self) -> bool {
+        matches!(self, ChangeEvent::Deleted { .. })
+    }
+}
+
+/// Classification of an event that creates a new path.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum CreatedKind {
```
aaronvg: these types dont come from lsp-types?

engine/language_server/src/edit.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::{Path, PathBuf};
+    use url::Url;
+
+    #[test]
+    fn parse_windows_path() {
```
aaronvg, line 131: nice

engine/language_server/src/edit/text_document.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// The latest version of the document, set by the LSP client. The server will panic in
+    /// debug mode if we attempt to update the document with an 'older' version.
+    version: DocumentVersion,
+    /// The language ID of the document as provided by the client.
+    language_id: Option<LanguageId>,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum LanguageId {
+    Python,
```
aaronvg: should this say Baml? I'm guessing this is used right?

engine/language_server/src/server/api/requests/completion.rs
```diff
@@ -0,0 +1,72 @@
...
+            .get(&document_key)
+            .ok_or(anyhow::anyhow!(
+                "File {} was not present in the project",
+                document_key
+            ))
+            .internal_error()?;
+        let word = get_word_at_position(&doc.contents, &params.text_document_position.position);
+        let cleaned_word = trim_line(&word);
+        // let cleaned_word = word;
+        let completions = match cleaned_word.as_str() {
```
aaronvg, line 48: ooo nice

engine/language_server/src/server/schedule/thread/pool.rs
```diff
@@ -0,0 +1,113 @@
...
+
+        // Channel buffer capacity is between 2 and 4, depending on the pool size.
+        let (job_sender, job_receiver) = crossbeam::channel::bounded(std::cmp::min(threads * 2, 4));
+        let extant_tasks = Arc::new(AtomicUsize::new(0));
+
+        let mut handles = Vec::with_capacity(threads);
+        for i in 0..threads {
+            let handle = Builder::new(INITIAL_PRIORITY)
+                .stack_size(STACK_SIZE)
+                .name(format!("ruff:worker:{i}"))
```
aaronvg: baml:worker

engine/language_server/src/tests.rs
```diff
@@ -0,0 +1,380 @@
...
+  "#
+}
+
+test TestSucc {
+  functions [Succ]
+  args { inp 1 }
+}
+"##;
+
+    // #[test]
```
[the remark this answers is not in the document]
aaronvg: is this supposed to work?

.github/workflows/primary.yml
aaronvg: to further comment on this -- it seems like we need some github script that uploads the cli artifacts to the release that's created. We used to do this a while back.. let me see if i can find it.

aaronvg: here is an example that attached the artifacts: https://github.com/BoundaryML/baml/blob/7ab83bc9160ec9bb1e1508d361cb7fea4793e6f2/.github/workflows/cli.yml#L140

engine/language_server/src/edit/text_document.rs
imalsogreg: It's some language-tracking stuff left over from Ruff. We don't track multiple languages, so I'll delete this whole type and all the fields that depend on it.

engine/language_server/src/baml_project/watch.rs
imalsogreg: They don't exist in lsp-types or lsp-server by the name `CreatedKind`. TBH I'm not sure if we need this module at all, since our LSP server relies on messages from the client about file changes. I'll try deleting and retesting.

imalsogreg: It's currently unused. Deleted the whole module.

engine/.cargo/config.toml
imalsogreg: The previous `-ctr-static` flag was disabling static linking of musl into the binaries. Removing it makes the linker fully statically link musl.

This change was necessary for the binary to work on my linux machine, which doesn't have a globally shared glibc, and probably for other users machines if they depend on static linking for various reasons.

Rereading the linked github thread, I worry that this change might break building of the baml shared library on linux machines - I'll have to check this.

imalsogreg: `link-arg=-lm` was a needed on my linux machine for the compiler to find the math lib during static linking. It may be useful to remove that arg and see if CI can work without it.

imalsogreg: @aaronvg integ tests on an ubuntu-20.04 machine pass. IIUC that means this change hasn't broken static linking for the cdylib used in baml clients.

Issue #1403, opened 2025-02-01
Expose prompt and parser separately


aaronvg:
Some folks want a more modular approach to using BAML, where they may modify the api request themselves.

This issue tracks exposing the api request baml makes from a function, as well as the parser

hellovai:
Specifically we could expose this:

```python
# get the raw request BAML would have fired.
prompt = b.prompt.SomeFunction(..)

str = call_llm_yourself(prompt)

# parse into the type using SAP
T = b.parse.SomeFunction(llm_response: str)

# or for streaming
Partial_T = b.stream.parse.SomeFunction(llm_response: str)
```

This would also enable support for RealTimeAPI and Batch APIs meanwhile while we wait (or any other providers as well).

PGCodehub:
Yes this would be amazing, make the library more modular

Elijas:
+1 looking forward to this

Elijas:
+1 for this feature

I have implemented a temporary workaround for myself until this is implemented, maybe others will find it useful

[Details and Source Code](https://github.com/Elijas/baml-parser)

### Example 1
```python
# We have the class defined in the food.baml file:
#
#    class Food {
#        count int
#        items string[]
#    }

llm_response = """
    Of course, I can help you with that. Here is a list of fruits:
    { count: 2.0,, "items": [banana, 'cherry'] }
    Let me know if you need anything else!
"""

b.ParseFood(llm_response)
# returns Food(count=2, items=['banana', 'cherry'])
```

### Example 2

```python
tb = TypeBuilder()
tb.add_baml("""
   dynamic class Result {
       price float
       tags string[]
   }
""")

llm_response = """
    Here is your answer:
    { price: 42,, "tags": new! }
    Let me know if you need anything else!
"""

b.ParseCompletion(llm_response, {"tb": tb})
# returns Result(price=42.0, tags=['new!'])
```

aaronvg:
https://docs.boundaryml.com/guide/baml-advanced/modular-api

YourTechBud:
@aaronvg I wouldn't call this necessarily done as it still requires you to write BAML files. Would be great if we could use the schema language and parser without any codegen.

aaronvg:
it seems like that is definitely related, but the task would be "Make BAML usable without .baml files", and perhaps that task outlines all the things that are missing. If you have a summary of the things you'd love out of that runtime API do list them and I can create a separate issue.

YourTechBud:
[...]

PR #1505, opened 2025-02-22
Expose HTTP request and LLM response parser separately

antoniosarosi:
#1403


engine/language_client_codegen/src/typescript/templates/sync_client.ts.j2
```diff
@@ -24,18 +26,22 @@ type BamlCallOptions = {
...
+    this.httpRequest = new HttpRequest(runtime, ctxManager)
+    this.httpStreamRequest = new HttpStreamRequest(runtime, ctxManager)
+    this.llmResponseParser = new LlmResponseParser(runtime, ctxManager)
+    this.llmStreamParser = new LlmStreamParser(runtime, ctxManager)
+    this.bamlOptions = bamlOptions || {}
   }
 
-  withOptions(baml_options: BamlCallOptions) {
-    return new BamlSyncClient(this.runtime, this.ctx_manager, baml_options)
+  withOptions(bamlOptions: BamlCallOptions) {
```
aaronvg, line 43: nice catch with the name

aaronvg (commented):
Nice work with the giant PR

Can you also run the with_options test for all 3 langs, and collector tests in ruby, and ts? They should all pass.

Some nits on naming -- e.g. build_request in the internal pyo3 interface / other ones could be renamed to get_http_request_body -- as that is exactly what we return -- it's not headers or anything.

The external interface can remain `b.request` 


In python integ tests there is a script that runs some tests on CI. You'll want to ignore your new test file there. This is why CI fails

engine/language_client_codegen/src/typescript/templates/sync_client.ts.j2
antoniosarosi: @aaronvg build_request returns the entire HTTP request type (including headers, URL, method) not just the body.

PR #1520, opened 2025-02-25
hellovai/go

hellovai:
- **compile for go!**
- **gitignore**


engine/language_client_go/main.go
```diff
@@ -0,0 +1,109 @@
...
+	callback_id, callback := create_unique_id()
+
+	C.call_function_from_c(runtime, function_name, &kwargs, callback_id)
+
+	return_channel := make(chan DoSomethingStreamResult)
+	go func() {
+		for result := range callback {
+			// TODO: Handle the result
+			// error handling, type checking, etc.
+			return_channel <- DoSomethingStreamResult{StreamResult: StreamResult[string, string]{partial: &result.data}}
```
YourTechBud, line 71: This will block this goroutine and hence the trigger callback goroutine untill the user reads the message from this channel. This can cause the rust thread to block as well. I suggest you make the channel  on line 65 a buffered channel to avoid blocking on this line

engine/language_client_go/callbacks.go
```diff
@@ -0,0 +1,56 @@
...
+		if isDone {
+			close(callback)
+			callbackMutex.Lock()
+			defer callbackMutex.Unlock()
+			delete(dynamicCallbacks, id_uint)
+		}
+	}
+}
+
+func create_unique_id() (C.uint32_t, chan ResultCallback) {
```
YourTechBud, line 45: Why not return a clean go type here? Convert to corresponding C type where it's being consumed

engine/language_client_go/main.go
```diff
@@ -0,0 +1,109 @@
...
+extern void trigger_callback(uint32_t, bool, const char *);
+*/
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+type StreamResult[Partial any, Final any] struct {
```
YourTechBud, line 19: We should confirm if gophers apart from me are comfortable to get a stream of such an object

engine/language_client_go/main.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (result *StreamResult[Partial, Final]) IsPartial() bool {
+	return result.partial != nil
+}
+
+func (result *StreamResult[Partial, Final]) Error() error {
+	return result.error
+}
+
+type DoSomethingStreamResult struct {
```
YourTechBud, line 45: Wr can honestly dump this. Its not really adding any value.

engine/language_client_go/main.go
```diff
@@ -0,0 +1,109 @@
...
+	StreamResult[string, string]
+}
+
+// UpperCase in Go only!
+func DoSomethingStream(arg string) <-chan DoSomethingStreamResult {
```
YourTechBud, line 50: I'd be cool if we returned StreamResult directly here. Finalize the generic types in the declaration here itself.

engine/language_client_go/main.go
```diff
@@ -0,0 +1,109 @@
...
+	if runtime != nil {
+		C.destroy_baml_runtime(runtime)
+	}
+}
+
+func init() {
+	C.register_callback((C.callback_fcn)(C.trigger_callback))
+	runtime = C.create_baml_runtime()
+	if runtime == nil {
+		panic("Failed to create Baml runtime")
```
YourTechBud, line 90: This would have to be documented. I really dont mind baml exposing a `NewClient` method which holds the instance of the runtime within it. All Baml functions could be methods of this object

engine/language_client_go/main.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+func main() {
+	channel := DoSomethingStream("Hello, world!")
+	for result := range channel {
+		if result.IsPartial() {
```
YourTechBud: We just need one of IsPartial and IsFinal. Don't need both of em. Also, i know this is a sample, but please do the error check first. Lol.

YourTechBud (commented):
Had some thoughts after revisiting this code.

engine/language_client_go/main.go
tberman: What would the other option be?

engine/language_client_go/main.go
hellovai: it doesn't look like its blocking! ran a test and it runs all 3 in parallel from what i see

YourTechBud: The user may not read the channel at the pace you are. What if the user decides to do something compute intensive in the range loop on the receiving end of this channel?

engine/language_client_go/main.go
YourTechBud: That's what I'd like to know as well. I recommend this way myself.

engine/language_client_go/main.go
tberman: Can confirm with an unbuffered channel this is a blocking send until the consumer reads.

engine/language_client_go/main.go
tberman: I think regardless of channel vs callback this is definitely the 'right' data structure.

Curious though, and this may be my lack of knowledge w/ baml streaming. When would Partial and Final be a different type?

engine/language_client_go/main.go
tberman: Yes please, panic on init would be less than desirable!

Plus doing it as @YourTechBud mentions would allow you to have multiple baml runtimes (if that ever is helpful) and let you use `runtime.AddCleanup` to destroy the underlying runtime

hellovai:
closing in favor of hellovai-go branch

PR #1580, opened 2025-03-08
identified bug.

hellovai:
need to plumb through types into streaming code


hellovai:
fixed in a different PR

PR #1591, opened 2025-03-11
Add version checking and safe import mechanism for Python generators

hellovai:
- Introduce version checking in Python generators to ensure compatibility
- Add `safe_import.py` with `EnsureBamlPyImport` context manager
- Implement `get_version()` in Rust to expose package version
- Update `__init__.py` templates to include version validation

Example error message:
Screenshot 2025-03-11 at 1 16 45 PM

----

> [!IMPORTANT]
> Add version checking and safe import mechanism for Python and TypeScript generators to ensure compatibility.
> 
>   - **Version Checking**:
>     - Implement `get_version()` in `lib.rs` for both Python and TypeScript to expose package version.
>     - Add `EnsureBamlPyImport` in `safe_import.py` for Python to check version compatibility.
>     - Add `ThrowIfVersionMismatch` in `safe_imports.ts` for TypeScript to check version compatibility.
>   - **Templates**:
>     - Update `__init__.py.j2` and `index.ts.j2` to include version validation and error messages for mismatches.
>   - **Integration Tests**:
>     - Update `__init__.py` and `index.ts` in integration tests to include version checking logic.
>   - **Misc**:
>     - Update `Cargo.toml` and `package.json` to reflect version changes.
>     - Add error messages for version mismatches in both Python and TypeScript.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for 0794f0eef7824a8cb4a4f3ad99db9da1b20e2848. It will automatically update as commits are pushed.</sup>


engine/Cargo.lock
```diff
@@ -913,7 +913,7 @@ dependencies = [
 
 [[package]]
 name = "baml-python-ffi"
-version = "0.1.0"
+version = "0.78.0"
```
imalsogreg, line 916: Possible to pull this version from the `engine` workspace so it stays in sync?

hellovai: i already did :) checkout the update to cargo.toml

imalsogreg:
Writing out the failure mode:
1. User is on baml-py v0, generators file says "v0", playground is at v0.
2. Playground automatically updates to v1
3. Playground refuses to generate code until generators file is updated
4. User updates generators file because the error message tells them to
5. Generated code knows that it was created by v1, but it encounters baml-py v0.

engine/language_client_codegen/src/python/mod.rs
```diff
@@ -105,6 +106,8 @@ impl TryFrom<(&'_ IntermediateRepr, &'_ crate::GeneratorArgs)> for PythonInit {
     fn try_from((_, gen): (&'_ IntermediateRepr, &'_ crate::GeneratorArgs)) -> Result<Self> {
         Ok(PythonInit {
             default_client_mode: gen.default_client_mode.clone(),
+            // TODO: Should we use gen.version instead?
```
imalsogreg, line 109: Seems better to me to grab the version from generators file, because that's what the error message implies.

I think it might be impossible for the generators file to be out of sync with CARGO_PKG_VERSION by the time codegen happens, meaning that there's no difference between (a) grabbing the version from CARGO_PKG_VERSION or (b) taking it from generators block. But it may be better not to rely on that fact. For example, discrepancies would exist between the two if the user is on a higher patch-version of the playground.

To say it another way - it would be confusing if the error messages said "your generators files says '0.1.0'" if their generators file actually says '0.1.1'.

hellovai: resolved with better names!

PR #1722, opened 2025-04-03
LLM function composition

### LINKED ISSUE #1707: Unify expression values and types with BAML values and types
The expression language demo has its own `Expr` and `ExprType` definitions that wrap and extend `BamlValue` and `FieldType`. This made the demo easy to create, but it isn't suitable for a full implementation because it prevents free intermixing of `Expr` forms (like function application) and BAML forms (like lists and classes).

More details are available in Notion: [https://www.notion.so/gloochat/Expr-BAML-Unification-1c8bb2d2621680fbb95dc57239800fa1](https://www.notion.so/gloochat/Expr-BAML-Unification-1c8bb2d2621680fbb95dc57239800fa1)

At the end of this work:

* Users can write exprs in lists: `[ DoSomething(1), DoSomething(2) ]`
* Users can construct classes with exprs: `MyCharacter { name: GenerateName("Victorian") }`

When this task is completed, we should quickly merge into `canary`, since the changes may otherwise quickly bitrot. (bitrot was not a problem earlier because the wrapper types were very uncoupled from everything else in BAML. After unification, expression language will be more tightly coupled).

imalsogreg:
Add several new syntactic forms to BAML, allowing the definition of functions whose bodies are expressions that may compose calls to other expression-bodied functions as well as the existing LLM functions.

The changes to the language in this PR are not meant to be used in production BAML apps. This is foundation work for a future, user-facing release. All syntax introduced here is very likely to change.

Closes https://github.com/BoundaryML/baml/issues/1707

Needs cleanup and more tests.

----

> [!IMPORTANT]
> Introduces function composition in BAML, adding expression-bodied functions and updating the type system, code generation, and testing framework to support this foundational change.
> 
>   - **Behavior**:
>     - Introduces function composition in BAML, allowing expression-bodied functions to compose calls to other functions.
>     - Not intended for production; syntax likely to change.
>   - **IR and Helpers**:
>     - Adds `ExprFunctionNode` and `ExprFunctionWalker` in `ir_helpers/mod.rs`.
>     - Updates `IRHelper` trait with `find_expr_fn` and `find_expr_fn_test` methods.
>   - **Type System**:
>     - Adds `Arrow` type to `FieldType` in `field_type/mod.rs`.
>     - Updates type coercion and checking to handle `Arrow` types.
>   - **Code Generation**:
>     - Updates Python, Ruby, and TypeScript code generation to handle new function types.
>     - Adds handling for `Arrow` types in generated type definitions.
>   - **Testing and Validation**:
>     - Adds tests for expression functions in `validation_files/expr`.
>     - Updates validation pipeline to include expression function validation and type checking.
>   - **Miscellaneous**:
>     - Updates `flake.nix` and `flake.lock` for dependency management.
>     - Enhances VSCode extension with new commands and event listeners for function testing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for 0b1fa017c1b4f7cf2d51781c55f6db39451706b4. It will automatically update as commits are pushed.</sup>


engine/baml-lib/baml-types/src/expr.rs
```diff
@@ -0,0 +1,286 @@
...
+        meta: T,
+    },
+
+    LLMFunction(Name, Vec<Name>, T),
+    Var(Name, T),
+    Lambda(Vec<Name>, Arc<Expr<T>>, T),
+    App(Arc<Expr<T>>, Arc<Expr<T>>, T),
+    Let(Name, Arc<Expr<T>>, Arc<Expr<T>>, T), // let name = expr in body
+    ArgsTuple(Vec<Expr<T>>, T),
+}
```
sxlijin, line 30: some questions:

- does `Expr::List` allow lists of heterogeneous type? if yes, what is the type of the list; if no, what do we use as our signal for whether or not `Expr::List` has been type-checked? same for maps
- what is the distinction between `Var` and `Let`?
- is `Lambda` how we bind function definitions?
- how are scopes modelled? or i guess this isn't a thing we have yet?

engine/baml-lib/baml-types/src/field_type/mod.rs
```diff
@@ -86,13 +86,20 @@ pub enum FieldType {
     Tuple(Vec<FieldType>),
     Optional(Box<FieldType>),
     RecursiveTypeAlias(String),
+    Arrow(Box<Arrow>),
```
sxlijin, line 91: is this for lambdas?

engine/baml-runtime/src/eval_expr.rs
```diff
@@ -0,0 +1,488 @@
...
+        _ => Err(anyhow::anyhow!("Not an application: {:?}", expr)),
+    }
+}
+
+/// Fully evaluate an expression to a value.
+pub async fn eval_to_value<'a>(
+    env: &EvalEnv<'a>,
+    expr: &Expr<ExprMetadata>,
+) -> anyhow::Result<Option<BamlValueWithMeta<()>>> {
+    let max_steps = 1000;
```
sxlijin: nit: lift out as a const

engine/baml-runtime/src/eval_expr.rs
```diff
@@ -0,0 +1,488 @@
...
+    use baml_types::{BamlMap, BamlValue};
+    use futures::channel::mpsc;
+    use internal_baml_core::ir::repr::make_test_ir;
+
+    use super::*;
+    use crate::BamlRuntime;
+
+    // Make a testing runtime. It assumes the presence of
+    // OPENAI_API_KEY environment variable.
+    fn runtime(content: &str) -> BamlRuntime {
```
sxlijin, line 340: does this mean `cargo test` will now fail by default?

engine/baml-runtime/src/eval_expr.rs
```diff
@@ -0,0 +1,488 @@
...
+use crate::{BamlRuntime, FunctionResult};
+use baml_types::expr::{Expr, ExprMetadata, Name};
+use baml_types::Arrow;
+use baml_types::{BamlMap, BamlValue, BamlValueWithMeta};
+use internal_baml_core::ir::repr::IntermediateRepr;
+
+pub struct EvalEnv<'a> {
+    pub context: HashMap<Name, Expr<ExprMetadata>>,
+    pub runtime: &'a BamlRuntime,
+    pub expr_tx: Option<mpsc::UnboundedSender<Vec<SerializedSpan>>>,
```
sxlijin, line 19: is this how the incremental progress is recorded and pushed out? or is that in a different PR?

engine/baml-runtime/src/eval_expr.rs
```diff
@@ -0,0 +1,488 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use crate::{BamlRuntime, FunctionResult};
+use baml_types::expr::{Expr, ExprMetadata, Name};
+use baml_types::Arrow;
+use baml_types::{BamlMap, BamlValue, BamlValueWithMeta};
+use internal_baml_core::ir::repr::IntermediateRepr;
+
+pub struct EvalEnv<'a> {
```
sxlijin, line 16: i assume this is $$\Gamma$$ aka the typing context - but how are typing contexts stacked?

engine/baml-runtime/src/eval_expr.rs
sxlijin, line 1: two questions:

- how does execution _scheduling_ work here? i didn't read super closely, but it seems like the beta reduction is all happening here, and i don't quite understand how this code maps onto the demo you showed.
- i would rather merge this as-is rather than block, but what do you think about defining our own error/result types in the evaluator instead of using anyhow? we have ~ok error semantics in LLMFunctionResult but i'm mildly concerned that using anyhow here will force us to use anyhow error downcast, typescript/python-style, to disambiguate different errors, and i don't love the idea of that

engine/baml-lib/baml-types/src/field_type/mod.rs
imalsogreg: Any function `T -> T`, including expr functions and LLM functions. If it's a function, it has type `Arrow { param_types, return_type }`.

engine/baml-lib/baml-types/src/expr.rs
imalsogreg: - `Expr::List` can syntactically hold heterogeneous lists, but the BAML typesystem only supports homogeneous lists, so a list that somehow contained elements of different types would be ill-typed and should fail the typechecker. In the case of a list of unions, the list elements would each have the same union type, although it would be fine for them to be different variants (all of the same union).
- `Var` is a bare variable. `Let` is a three-part form that binds a variable to some value in the scope of another expression. `x` parses to `Expr::Var("x")`. `{ let x = 1; (x + 1) }` parses to `Let( "x", Expr::Const(1), BinOp("+", Expr::Var("x"), Expr::Const(1))`. (We don't have binary operators yet, BinOp is just an imaginary expr that I thought makes this example more clear).
- Yep! Expression-functions and LLM functions are both modeled as lambda abstractions.
- I haven't implemented scoping for name resolution, so we don't have capture-avoiding substitution, and it's likely possible to write an expression where a variable is interpreted incorrectly.

engine/baml-runtime/src/eval_expr.rs
imalsogreg: If any unit test actually uses this function, yes, `cargo test` would fail in an empty environment. Right now it's just a convenience thing to use during development. I tweak the inline BAML code starting on line 386, then uncomment line 382 to enable the test.

Probably should come up with something better. Integ test would make sense, and I think we'd be able to do that as soon as we wire expr functions up to codegen.

engine/baml-runtime/src/eval_expr.rs
imalsogreg: This is the send end of a channel used for publishing the spans of any LLM functions being evaluated, yep :)

engine/baml-runtime/src/eval_expr.rs
imalsogreg: You got it. The `context` field is  Γ . There is no single `EvalEnv` that holds a stack of scoped contexts. Instead every sub expression gets a new `EvalEnv` with a modified (flat) `context` that's appropriate for that subexpression.

For example, typechecking `Let( "x", 1, Expr::IsZero(1) )` in the empty context would involve first typechecking `1` in the empty context, then constructing a new context containing `("x", Number)`, and using that context to typecheck `Expr::IsZero(1)`).

engine/baml-runtime/src/eval_expr.rs
imalsogreg: scheduling:
Roughly call-by-need. The top-level driver is "eval_to_value", which then does beta-reduction and substitution as needed, depending on what expressions it encounters. The scheduling is implicit in each branch of `beta_reduce`. 

Specifically there's decision you can make when you beta-reduce a function - (a) fully evaluate all the args and substutute the results in as the fn parameters or (b) substitute the arguments directly as expressions into the function body. (a) gives you call-by-value or call-by-reference, (b) gives you laziness. Laziness seemed like a wonderful property to have because it lets us trivially have top-level definitions that don't get evaluated unless they're used, or define structs with expensive fields that only get evaluated if their need is justified by some condition that can only determined at runtime.

But the evaluation schedule is something we still have to design - what I have here was just thrown together for the demo and wouldn't work in prod. There's still some design work to do, too, because I think our error handling story will interact with evaluation order. And we want to get some parallelism in the LLM calls where possible. So, this code will be the focus of some interesting work :)

errors:
Switching from `anyhow` to something like `thiserror` makes a lot of sense if you plan to consume the errors - pattern match on them for something. So thinking about it - we may want to pattern match on them so that our runtime can differentiate between different types of errors for retrying/not-retrying, doing a fallback, sending a special instruction to the playground, etc. And also having structured errors, we may want to expose that structure to the user, in the logs (one numerical error code per constructor) or directly in BAML code (they could write their own recovery functions). So, yeah I think it would be a great idea both in Expression Language and other parts of the runtime, to design a hierarchy of error types with `thiserror`! Exactly what the errors should look like is a matter for design, so I agree we probably shouldn't block this PR on that. We should definitely keep it in mind when we circle back to user-facing error handling (if not earlier).

engine/baml-runtime/src/eval_expr.rs
sxlijin: my question the other day at standup around `cargo test` only supporting pass/fail and not more complex statuses was basically about this

a bunch of the `baml-cli serve` tests have this same issue right now

imalsogreg: Heh, agreed!

engine/baml-runtime/src/eval_expr.rs
```diff
@@ -0,0 +1,503 @@
...
+    let max_steps = 1000;
+    let mut current_expr = expr.clone();
+
+    for steps in 0..max_steps {
+        match current_expr {
+            Expr::Atom(value) => return Ok(Some(value.clone().map_meta(|_| ()))),
+            Expr::List(items, meta) => {
+                let mut new_items = Vec::new();
+                for item in items {
+                    let val = Box::pin(eval_to_value(env, &item)).await?;
```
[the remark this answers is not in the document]
imalsogreg: We need the Pins because we are awaiting a recursive call. Without these, the state machine is apparently infinitely sized.

PR #1865, opened 2025-04-28
Go Encode/Decode fixes

tberman:
This PR does a couple things.

1. Adds some cffi roundtripping tests for go
2. Uses those tests to make sure that classes encode/decode properly
3. Uses those tests to make sure that map[string]primitive and []primitive work if they are return values or input parameters
4. Remove passing the typeMap recursively since its statically available inside the baml code (you will notice the type signature is the only change)
5. Added `BamlTypeName()` to the `BamlSerializer` class which allows us to get a baml type name from classes/unions/enums.
6. Use that to support classes as values in maps

Still needs a test with a Union for map/list


hellovai (approved):
temporarily approved so we can get this out in the release for @tberman

PR #1889, opened 2025-05-01
Fix optional struct fields

tberman:
Now optional struct fields can be encoded and decoded properly

I also made it so that gofmt/goimports runs on generate so this is the last time the diff should be this annoying!

----

> [!IMPORTANT]
> Fixes encoding/decoding of optional struct fields and adds `gofmt` to Go code generation.
> 
>   - **Behavior**:
>     - Fixes encoding and decoding of optional struct fields in `generate_types.rs`.
>     - Updates `Decode` functions in `types.go` to handle nil values for optional fields.
>   - **Code Generation**:
>     - Adds `gofmt` execution to `generators.baml` and `primary.yml` to format Go code on generation.
>   - **Misc**:
>     - Updates `generators.baml` in multiple `inlinedbaml` files to include `gofmt` command.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for 2c08b828611884ab0b0186dd28b5d1e7b08bf9aa. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


integ-tests/baml_src/generators.baml
```diff
@@ -43,6 +43,7 @@ generator lang_go {
   output_dir "../go"
   version "0.86.1"
   client_package_name "example.com/integ-tests"
+  on_generate "gofmt -w . && goimports -w ."
```
aaronvg: neat, reminding myself that this could probably go on the docs, and on the baml init sample generator.

PR #1976, opened 2025-05-22
[Go] Simple Collector

tberman:
Looking for feedback on the approach. I think it can be macro'd and made better but curious what y'all think.

If it looks good I will finish out the interface and get it ready to actually be merged.

----

> [!IMPORTANT]
> Introduces a collector mechanism in the Go client for the BAML engine, enhancing function call tracking and logging.
> 
>   - **Behavior**:
>     - Added `call_collector_function` to `lib.rs` for invoking collector functions.
>     - Updated `call_function_from_c` in `lib.rs` to accept collectors.
>     - Modified `AaaSamOutputFormat` in `client.go.j2` to use collectors.
>   - **CFFI**:
>     - Added `CallCollectorFunction` in `baml_cffi_wrapper.c` and `baml_cffi_wrapper.h`.
>     - Updated `CallFunctionFromC` in `baml_cffi_wrapper.c` and `baml_cffi_wrapper.h` to include collectors.
>   - **Go Client**:
>     - Added `collector.go` for managing collectors and their interactions.
>     - Updated `runtime.go` to pass collectors to `CallFunction`.
>     - Modified `exports.go` to handle collector functions.
>   - **Integration Tests**:
>     - Updated `test.go` to test collector functionality with `AaaSamOutputFormat`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for b0a3e20589e57a0ec6fddb7bd1f6300692058f53. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


engine/language_client_cffi/src/lib.rs
```diff
@@ -193,6 +206,22 @@ fn call_function_from_c_inner(
...
+    let collector_ptrs = collectors as *const *const libc::c_void;
+    let collectors = match collectors_length {
+        0 => None,
+        _ => Some(
+            unsafe { std::slice::from_raw_parts(collector_ptrs, collectors_length) }
+                .iter()
+                .map(|c| unsafe {
+                    let collector = Arc::from_raw(*c as *const Collector);
+                    let clone = collector.clone();
+                    let _ = Arc::into_raw(clone);
```
tberman: This can't be real.

There has to be a better way.

The docs implied that Arc::from_raw basically 'consumes' the reference so we need to keep leaking it. ChatGPT agreed. I can't imagine this is right.

tberman: Maybe I could Box an Arc pointer? and then use the box to go across the go/rust boundary. is that real?

engine/language_client_cffi/src/lib.rs
```diff
@@ -279,3 +308,96 @@ fn call_function_stream_from_c_inner(
...
+                }
+                _ => Err(anyhow::anyhow!(
+                    "Failed to call function: {} on object type: {}",
+                    function_name,
+                    object_type
+                )),
+            }
+        }
+        "usage" => {
+            let usage = unsafe { Box::from_raw(object as *mut Usage) };
```
tberman: So this has the same problem as Arc where it is now 'owned' by this Box, and gets dropped when the Box is destroyed, which I need to figure out how to handle.

engine/language_client_cffi/src/lib.rs
tberman: I think it is, and its probably the best way, gonna try it!

tberman: Couldn't quite get it to work, so back to the drawing board (Arc pointers are not copyable for obvious reasons)

hellovai: you gotta do from box. will send over the code!

Issue #1887, opened 2025-05-01
[feat] expose tokens/timings through REST API calls


Elijas:
## Use case
Use case use REST APIs to get token counts, timings etc. associated with the request, but keep REST API stateless

## Solution Idea

There could be a middle-ground API across all clients, including API clients

pseudocode of /call_with_metrics implementation:
```
function b.with_stats.Any(...):
    c = Collector()
    r = b.Any(..., collector=c)
    return {result: r, stats: c.stats}
```

DomWilliamsEE:
I made an initial PR (#1989) that shows it is technically possible to add a Collector for a `/call_with_metrics` endpoint.
@hellovai pointed out we should document and decide on what the usage should look like.

I think dumping the raw Collector state in the response is messy and has unwanted fields. I would suggest something like this:
* `response`: the llm response, i.e. the raw response of `/call`
* `metrics`: has `input_tokens`, `output_tokens` and `duration_ms`, the bare minimum of metrics that are useful
* `logs`: the terminal output of the full function call is useful, but its inaccessible to the caller of the `/call` endpoint. I would suggest returning it as a free-form string here. Alternatively it could be the full nested structure of all the calls and tool usages.

So as an example:
```POST /call_with_metrics/:function_name```: Call a BAML function with extended metrics about the function call
returned.

```bash curl
curl \
  -X POST \
  "http://localhost:2024/call_with_metrics/MyFunctionName" \
  -H "Content-Type: application/json" \
  -d '{"arg1": "value1", "arg2": "value2"}'
```

Example response:
```json
{
  "response": "...",
  "metrics": {
    "input_tokens": 101,
    "output_tokens": 291,
    "duration_ms": 2810
  },
  "logs": "Function MyFunctionName\nClient: GPT4o (gpt-4o-2024-08-06) - 2810ms. StopReason: stop. Tokens(in/out): 101/291\n---PROMPT---..."
}
```

PR #1989, opened 2025-05-27
Add call_with_metrics endpoint to `baml-cli serve`

DomWilliamsEE:
My initial attempt at implementing #1887, comments welcome

----

> [!IMPORTANT]
> Add `/call_with_metrics/:msg` endpoint to BAML CLI server to handle calls with metrics using `Collector`.
> 
>   - **Behavior**:
>     - Adds `/call_with_metrics/:msg` endpoint in `mod.rs` to handle calls with metrics.
>     - Uses `Collector` to gather metrics and include them in the response.
>   - **Functions**:
>     - Adds `baml_call_with_metrics_axum()` in `mod.rs` to process requests to the new endpoint.
>     - Modifies `baml_call()` to accept `BamlCallCollector` and include metrics in the response if applicable.
>   - **Dependencies**:
>     - Updates `either` crate in `Cargo.toml` and `Cargo.lock` to include `serde` feature.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for cd11c9b698d4ffc2ece6196e5c7e357d1655551f. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -2220,6 +2220,9 @@ name = "either"
 version = "1.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "60b1af1c220855b6ceac025d3f6ecdd2b7c4894bfe9cd9bda4fbb4bc7c0d4cf0"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "encode_unicode"
--- a/engine/Cargo.toml
+++ b/engine/Cargo.toml
@@ -59,7 +59,7 @@ dashmap = "5.5.3"
 derive-new = "0.7"
 derive_builder = "0.20.0"
 derive_more = { version = "0.99.18", features = ["constructor"] }
-either = "1.8.1"
+either = { version = "1.8.1", features = ["serde"] }
 env_logger = "0.11.3"
 futures = { version = "0.3.30", features = ["executor"] }
 http = "1.1.0"
--- a/engine/baml-runtime/src/cli/serve/mod.rs
+++ b/engine/baml-runtime/src/cli/serve/mod.rs
@@ -36,6 +36,7 @@ use std::{path::PathBuf, sync::Arc, task::Poll};
 use tokio::{net::TcpListener, sync::RwLock};
 use tokio_stream::StreamExt;
 
+use crate::tracingv2::storage::storage::Collector;
 use crate::{
     client_registry::ClientRegistry, errors::ExposedError, internal::llm_client::LLMResponse,
     BamlRuntime, FunctionResult, RuntimeContextManager,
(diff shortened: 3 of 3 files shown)


hellovai (changes requested):
Appreciate the contribution here! But I'm not sure if this will work. 

Currently the collector lives in memory so sending it across the wire won't solve the issue as it can't be serialized. 

That said if you are able to add some tests and show this working happy to merge it in!

You also likely need to update the generated code we have.

DomWilliamsEE:
Those types do derive Serialize, I tested locally that it does work with `baml-cli serve`:
```
$ curl   -X POST   "http://localhost:2024/call_with_metrics/ExtractResume"   -H "Content-Type: application/json"   -d '{"resume": "      Vaibhav Gupta vbv@boundaryml.com Experience: - Founder at BoundaryML - CV Engineer at Google - CV Engineer at Microsoft Skills: - Rust - C++"}'
{
  "Right": {
    "response": {
      "name": "Vaibhav Gupta",
      "email": "vbv@boundaryml.com",
      "experience": [
        "Founder at BoundaryML",
        "CV Engineer at Google",
        "CV Engineer at Microsoft"
      ],
      "skills": [
        "Rust",
        "C++"
      ]
    },
    "metrics": {
      "usage": {
        "input_tokens": 101,
        "output_tokens": 75
      },
      "timing": {
        "start_time_utc_ms": 1748353793893,
        "duration_ms": 1910,
        "time_to_first_parsed_ms": null
      }
    }
  }
}
```

I will try to add a test for it

hellovai:
sorry let me rephrase! I was on my phone earlier!

What i meant was we should likely write a small design doc for what things we expose (raw prompt, raw http response, tokens, etc) before we go about this change and then update both:

1. the API endpoint we expose
2. the codegen we do in https://github.com/BoundaryML/baml/blob/canary/engine/language_client_codegen/src/openapi.rs to provide an interface

I think the idea of exposing metrics alone is interesting, but i wonder if we should do the design here such that we expose all of it.

additionally: the current API returns it nested in a key called: `Right`. we likely wouldn't want that. 

If you're down to write a brief spec for what you're thinking the end user behavior should be that would be great!

Our general take as a team is that before we write end facing code, we write the end user docs for how someone would interface with it!

That said, we do appreciate the proactiveness here, just that we have a language and every new thing we add becomes another thing people have to learn, so writing the docs first helps us ensure we are on track with establishing clarity.

The docs are in the `fern` folder. to run them locally you can do this:

```
cd fern
pnpm install
pnpm dev
```

DomWilliamsEE:
Ah right, makes sense, thanks for the detail. I will write some docs about how it could look and be used. This first version and its (lack of) design was mainly to get the ball rolling and determine if its technically possible. 
Big derp on the "Right" in the response!

DomWilliamsEE:
I wrote up a little proposal in the original issue :ok_hand:

sxlijin:
Closing this because we haven't come back to this in many months, but we should definitely look into this again at some point

PR #2103, opened 2025-07-03
feat: responses API integration

ba11b0y:
- This PR adds support for the [Responses API](https://platform.openai.com/docs/api-reference/responses) by OpenAI by adding a new provider called `openai-responses`


engine/baml-lib/llm-client/src/clientspec.rs
```diff
@@ -61,6 +61,8 @@ pub enum OpenAIClientProviderVariant {
     Ollama,
     /// The Azure client provider variant
     Azure,
+    /// The OpenAI Responses API variant
+    Responses,
```
aaronvg, line 65: name this OpenAIResponses

hellovai: dont need to cause this is called `OpenAIClientProviderVariant` enum

engine/baml-runtime/src/internal/llm_client/primitive/openai/types.rs
```diff
@@ -6,6 +6,212 @@ use serde::{
 pub type CompletionResponse = ChatCompletionGeneric<CompletionChoice>;
 pub type ChatCompletionResponse = ChatCompletionGeneric<ChatCompletionChoice>;
 
+/// OpenAI Responses API response structure
```
hellovai, line 9: advice, keep only the bare minimum of what we need here. otherwise if they change the API we will break due to failure to parse. so the best thing we can do is ignore all data that we don't need instead of maintaining a 1:1

ba11b0y: ah makes sense!
resolved via https://github.com/BoundaryML/baml/pull/2103/commits/49f284c336db436ac669e168d47b260c2f772572

PR #2117, opened 2025-07-09
for loops implementation

ba11b0y:
TODO: nested for loops don't work right now.

----

> [!IMPORTANT]
> Implement for loop support in Baml compiler and VM, adding iterator instructions and tests.
> 
>   - **Behavior**:
>     - Implement for loop compilation in `lib.rs`, handling iterator creation and loop execution.
>     - Add `CreateIterator` and `IterNext` instructions in `bytecode.rs` for iterator operations.
>     - Update VM execution in `vm.rs` to handle new instructions and manage iterators.
>   - **Testing**:
>     - Add tests for for loops and iterator instructions in `vm.rs` and `tests/vm.rs`.
>     - Ensure correct execution of simple and nested for loops, and iterator operations.
>   - **Misc**:
>     - Update `debug.rs` to display new instructions.
>     - Add comments and TODOs for future improvements and nested loop support.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for e5bd0bb1818b16430ed46fa7e4e9ca4819756bff. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


engine/baml-compiler/src/lib.rs
```diff
@@ -759,6 +824,109 @@ mod tests {
...
+                function: 0,
+                instruction_ptr: 0,
+                locals_offset: 0,
+            }],
+            stack: vec![Value::Object(0)],
+            objects,
+            globals,
+        };
+
+        let result = vm.exec()?;
```
antoniosarosi: This is the compiler crate, tests here should only check that the compiler emits the correct instructions. Tests that actually need to run the VM itself should be in the VM crate.

engine/baml-compiler/src/lib.rs
```diff
@@ -759,6 +824,109 @@ mod tests {
...
+            {
+                assert_eq!(*iter_index, 0); // Should start at index 0
+                                            // Array should be at index 1 (after function at index 0)
+                assert_eq!(*array, 1);
+            } else {
+                panic!("Expected Iterator object");
+            }
+        } else {
+            panic!("Expected Object value");
+        }
```
antoniosarosi: When you have this big if blocks with a tiny else statement at the end, use Rust's `let-else` construct:

```rust
let Value::Object(index) = result else {
    panic!("Expected Object value");
};

let Object::Iterator { array, index: iter_index } = &vm.objects[index] else {
    panic!("Expected Iterator object");
};

assert_eq(*iter_index, 0);
assert_eq(*array, 1);
```

That way it's much easier to read each case and you don't have a lot of nesting. It's similar to early returns in functions.

antoniosarosi (commented):
@ba11b0y Check out [`let-else`](https://doc.rust-lang.org/book/ch06-03-if-let.html#staying-on-the-happy-path-with-letelse) in Rust

engine/baml-lib/ast/src/ast/stmt.rs
```diff
@@ -69,10 +95,31 @@ impl Stmt {
...
+            Stmt::ForLoop(stmt) => &stmt.iterator,
+        }
+    }
+
+    pub fn statements(&self) -> &Vec<Stmt> {
+        match self {
+            Stmt::Let(_) => panic!("Let statement does not have statements"),
+            Stmt::ForLoop(stmt) => &stmt.body.stmts,
+        }
+    }
```
antoniosarosi: You don't need any of these functions, notice how all of them `panic!` if called with the wrong variant of `Stmt`. Usually you'll match over a `Stmt` and at that point the variant is narrowed down, so you'll never check that a let statement has an iterator. You'll do:

```rust
match statement {
    Stmt::ForLoop(stmt) => {
        stmt.iterator // This exists here because we know the type.
    }

    Stmt::Let(stmt) => {
        // Again we know the type of stmt so we can access fields.
    }
}
```

Plus, if you match over a statement and then need to match over the same statement again to grab a field though a getter, it's just redundant code and redundant checks at runtime.

```rust
// Only one variant check
match statement {
    Stmt::ForLoop(stmt) => stmt.iterator
}

// This is 2 checks for no reason
match statement {
    Stmt::ForLoop(_) => statement.iterator()
}
// Because iterator is defined as:
fn iterator(&self) {
    match self {
        Stmt::Let(_) => panic!("Let statemetn does not have an iterator"
        Stmt::ForLoop(_) => &stmt.iterator
    }
}
```

Don't be afraid of public fields in structs. You don't need getters everywhere a la Java. Private fields with getters are nice if you're exposing some user facing API in a library. Other than that, in our internal code, we control how fields are accessed so no need for boilerplate @ba11b0y .

ba11b0y: ah makes sense. My bad, I come from a world guarding access patterns.

PR #2132, opened 2025-07-14
add source line numbers when displaying bytecode

ba11b0y:
> [!IMPORTANT]
> Add source line number tracking to bytecode generation and display for improved debugging context.
> 
>   - **Behavior**:
>     - Track source line numbers in `Compiler` in `lib.rs` using `current_source_line` and `set_source_line_from_span()`.
>     - Update `emit()` in `lib.rs` to push `current_source_line` to `source_lines` in `Bytecode`.
>     - Modify `display_bytecode()` in `debug.rs` to include source line numbers in output.
>   - **Data Structures**:
>     - Add `source_lines: Vec<usize>` to `Bytecode` in `bytecode.rs` to map instructions to source lines.
>   - **Misc**:
>     - Remove unused span variables in `compile_expression()` in `lib.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for d0aa17a6d62106758b74fa242e56cb87a1ac18ea. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


engine/baml-compiler/src/lib.rs
```diff
@@ -233,12 +246,14 @@ impl<'g> Compiler<'g> {
     fn compile_expression(&mut self, expression: &Expression) {
         match expression {
             // Constants.
-            Expression::BoolValue(bool, _span) => {
+            Expression::BoolValue(bool, span) => {
+                self.set_source_line_from_span(span);
```
antoniosarosi: Looks like this line is present at the beginning of each match case. We can put it above the match statement only once:

```rust
self.set_source_line_from_span(span);

match expression {
    // ...
}
```

engine/baml-vm/src/bytecode.rs
```diff
@@ -180,13 +180,20 @@ pub struct Bytecode {
 
     /// Constant pool.
     pub constants: Vec<Value>,
+
+    /// Source line mapping.
+    ///
+    /// Maps instruction indices to their source line numbers.
+    /// Each element corresponds to an instruction at the same index.
+    pub source_lines: Vec<Option<usize>>,
```
antoniosarosi: Why is it `Option<usize>`? Can a bytecode instruction not map to any line in the source? Looks like this is never set to `None` either way.

antoniosarosi (changes requested):

engine/baml-vm/src/bytecode.rs
ba11b0y: that's true, I started with an `Option` in mind to be safe, but yeah now I realize it's not needed. Will make the change.

engine/baml-compiler/src/lib.rs
ba11b0y: thanks!

engine/baml-vm/src/bytecode.rs
ba11b0y: It turns out even a 0 is fine since line numbers always begin with `1`
Resolved via https://github.com/BoundaryML/baml/pull/2132/commits/d0aa17a6d62106758b74fa242e56cb87a1ac18ea

engine/baml-compiler/src/lib.rs
ba11b0y: resolved via https://github.com/BoundaryML/baml/pull/2132/commits/d0aa17a6d62106758b74fa242e56cb87a1ac18ea

antoniosarosi (approved):

PR #2139, opened 2025-07-15
pretty print bytecode

ba11b0y:
100% vibe coded to pretty print instructions based on their type.

image

image

----

> [!IMPORTANT]
> Add color-coded pretty printing for bytecode instructions in BAML VM using the `colored` crate.
> 
>   - **Behavior**:
>     - `display_bytecode()` in `debug.rs` now pretty prints bytecode with colors based on instruction type.
>     - Uses `colored` crate to apply colors if output is to a TTY.
>   - **Functions**:
>     - Adds `get_instruction_color()` in `debug.rs` to map instructions to colors.
>   - **Dependencies**:
>     - Adds `colored = "2.1.0"` to `Cargo.toml`.
>   - **Versioning**:
>     - Updates `baml-vm` and `baml-compiler` versions in `Cargo.lock`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for ef0124edb45eef86f20ac467c3df9e99a84e1f7d. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/engine/baml-vm/src/debug.rs
+++ b/engine/baml-vm/src/debug.rs
@@ -22,6 +22,8 @@
 //!
 //! ```
 
+use colored::*;
+
 use crate::{Function, Instruction, Object, Value};
 
 /// Context aware instruction display.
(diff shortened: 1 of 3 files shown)


engine/baml-vm/src/debug.rs
```diff
@@ -22,6 +22,10 @@
 //!
 //! ```
 
+use std::io::IsTerminal;
+
+use colored::*;
```
antoniosarosi, line 27: Nit 1: Prefer explicit imports over wildcard unless this is a prelude:

```rust
use colored::{Color, Colorize};
```

Otherwise it's hard to see _what_ exactly we're importing and can cause name collisions.

engine/baml-vm/src/debug.rs
```diff
@@ -145,6 +149,24 @@ pub fn display_value(value: &Value, objects: &[Object]) -> String {
 /// See [`display_bytecode`] for more information.
 const COLUMN_MARGIN: usize = 3;
 
+/// Get color for instruction based on its type
+fn get_instruction_color(instruction: &Instruction) -> Color {
```
antoniosarosi, line 153: Nit 2: In Rust, the [naming conventions](https://rust-lang.github.io/api-guidelines/naming.html) don't really have _getters_, so in most codebases you won't see `get_foo` functions, they are just named `foo`. In this case id would be

```rust
fn instruction_color(instruction: &Instruction) -> Color
```

Or you could impl it:

```rust
impl Instruction {
    fn color(&self) -> Color
}
```

But that would make it available throughout the entire codebase while it's only needed in the debug module, so having it as a separate function is probably better.

PR #2194, opened 2025-07-27
Allow optional cache_control arg in jinja role fn

imalsogreg:
- Closes https://discord.com/channels/1119368998161752075/1253172277449855029/1399068983184855100

----

> [!IMPORTANT]
> Allow optional `cache_control` argument in `role` function within `baml::Chat` and add test case in `role.baml`.
> 
>   - **Behavior**:
>     - Allow optional `cache_control` argument in `role` function within `baml::Chat` in `types.rs`.
>     - `cache_control` can be a `Map` of `String` to `Unknown` or `None`.
>   - **Tests**:
>     - Add test case in `role.baml` to validate `cache_control` usage in `role` function.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for e01ec44b8872cec8ed56398208748c4986a06f0e. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/engine/baml-lib/baml/tests/validation_files/functions_v2/prompt_errors/role.baml
+++ b/engine/baml-lib/baml/tests/validation_files/functions_v2/prompt_errors/role.baml
@@ -1,0 +1,10 @@
+function Foo() -> string {
+    client "openai/gpt-4o"
+    prompt #"
+        You are a helpful assistant.
+        {{ ctx.output_format }}
+
+        {{ _.role("assistant", cache_control={"type": "ephemeral"}) }}
+        Hi
+    "#
+}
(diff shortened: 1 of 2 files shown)


hellovai (commented):
Shouldn't we just allow any key here for functions since folks can add any parameter here? Very ok with the current solution to solve the issue for that user.

imalsogreg:
@hellovai Yah I agree. It wasn't immediately obvious to me how to allow any key without some refactors.

PR #2184, opened 2025-07-23
Add abort support [WIP]

tomaszkiewicz:
This is my first take on implementing support for AbortController for Typescript.

Details of feature implemented in the docs file in the commit. 

It's still work in progress as I was unable to run integration testing on my fork yet, however I'm curious of your opinion about this implementation.

----

> [!IMPORTANT]
> Adds abort functionality to BAML streams using AbortController, with examples and tests.
> 
>   - **Behavior**:
>     - Adds abort functionality to `BamlStream` in `stream.ts` using `AbortController`.
>     - Supports aborting via `stream.abort()` or external `AbortSignal`.
>     - Handles pre-emptive aborts and exposes `isAborted` status.
>   - **API Changes**:
>     - Exports `AbortError`, `AbortSignal`, and `AbortController` in `index.ts`.
>     - Adds `BamlCallOptions` with `signal` parameter in `stream_client.ts`.
>   - **Examples**:
>     - Adds `abort-controller-example.ts` for TypeScript usage.
>     - Adds `AbortControllerExample.tsx` for React usage.
>     - Adds `route.ts` for Next.js API route example.
>   - **Tests**:
>     - Adds `abort-controller.test.ts` to verify abort functionality in integration tests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for 224193d88193514a88fa6a13d28e651dee52b13c. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #2246, opened 2025-08-01
Cancel Playground Tests

cg-jl:
I'll be implementing WASM functionality for enabling cancellation in the playground on top of #2184. I'm not sure how to proceed afterwards wrt merging, since I'm not going to be doing the rest of the clients for now.

Remaining:

- [x] rebase against `canary`
- [x] I can run `pnpm dev:vscode`
- [x] I can run a test in the playground
- [ ] Cancel button on the playground cancels the tests
- [ ] Cancel button runs on playground extension
- [ ] Notify #2184 of video + available changes


aaronvg:
you can delete all the non-wasm stuff for now fyi. Like only push the minimum viable code that's all been tested in the playground.

cg-jl:
I don't have the energy to review the rest of the canceling logic. I'll have some rest & continue tomorrow. Last push is to keep the changes visible & backed up. Will commit --amend when reviewed.

cg-jl:
Canceling logic is in place, but the current system doesn't work: Calling `token.cancel()` from JS blocks until `run_tests` has finished. I'm not sure why this happens right now, I'll keep debugging.

cg-jl:
Apparently, cancelling the "joined" request (run_tests) *does* cancel the future.
Which is sad because I'd prefer to keep the tasks separately cancelable so that we can grab their results as-needed.
I'll add a comment explaining & referencing this, because then some work has to be done wrt canceling if we stop joining the tasks before proceeding.

cg-jl:
overriden by #2357 .

PR #2269, opened 2025-08-07
chore: claude code permissions should not be checked in

ActuallyTrent:
The permissions file allows all `bash` commands (`Bash(*)`), but even if it didn't it still allows many dangerous commands (`Bash(rm:*)`, `Bash(node:*)`). 

While I don't think this in and of itself is really all that exploitable, I do think it increases the exposure risk of community members.

People who want to contribute to the repository may become at risk to something like a compromised MCP server or a model going a bit wild, even if they would normally manually approve commands. 

My personal take is that a public repository should not prescribe permissions for something as powerful as Claude Code.

[Anthropic's documentation](https://docs.anthropic.com/en/docs/claude-code/settings#settings-files) suggest adding `.claude/settings.local.json` to the `.gitignore`, as well.

----

> [!IMPORTANT]
> Remove all allowed Bash commands from `.claude/settings.local.json` to reduce exposure risk.
> 
>   - **Permissions**:
>     - Remove all allowed Bash commands from `permissions` in `.claude/settings.local.json`, leaving `allow` list empty.
>     - No changes to `deny` list.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=BoundaryML%2Fbaml&utm_source=github&utm_medium=referral)<sup> for e6cc4665bdedc33a71e329bc6c80274ac209363c. You can [customize](https://app.ellipsis.dev/BoundaryML/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


sxlijin:
Thanks for this! We've definitely been a little aggressive with our own Claude Code adoption and I totally hear you on the security risks here. Sorry this lounged for so long, but I'll merge this now :)

PR #2331, opened 2025-08-17
Support for abort/cancel in rust w/ idiomatic support for python/go/typescript

dexhorthy:
Built in close collaboration with @hellovai who sat in my apartment and shouted at claude with me all day

*Note: plan/Research artifacts in thoughts/ can be removed once they get put in a good location, leaving them in for now*

## What problem(s) was I solving?

Previously, once a BAML function call was initiated, there was no way to cancel it. This meant:
- Long-running LLM operations couldn't be stopped if no longer needed
- Retry chains would continue even after the user navigated away
- Resources were wasted on unnecessary LLM calls
- Poor user experience when operations needed to be cancelled

## What user-facing changes did I ship?

Added abort/cancellation support for all BAML operations using native language patterns:

### TypeScript
```typescript
const controller = new AbortController();
const promise = b.ExtractName("John Doe", { 
  abortController: controller 
});

// Cancel after 100ms
setTimeout(() => controller.abort(), 100);

try {
  await promise;
} catch (e) {
  if (e instanceof BamlAbortError) {
    console.log("Operation cancelled");
  }
}
```

### Python
```python
from baml_py import AbortController

controller = AbortController()
task = b.ExtractName("John Doe", 
  baml_options={"abort_controller": controller})

# Cancel after 100ms
await asyncio.sleep(0.1)
controller.abort()

try:
  await task
except Exception as e:
  if "abort" in str(e).lower():
    print("Operation cancelled")
```

### Go
```go
ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
defer cancel()

_, err := b.ExtractName(ctx, "John Doe")
if err != nil && strings.Contains(err.Error(), "context canceled") {
  fmt.Println("Operation cancelled")
}
```

## How I implemented it

- Added cancellation support at the orchestrator level in Rust runtime using `stream-cancel` crate's `Tripwire` mechanism
- Created language-specific bindings:
  - **TypeScript**: Native `AbortController` support via NAPI bridge
  - **Python**: Custom `AbortController` class exposed via PyO3
  - **Go**: `context.Context` cancellation with early detection goroutines
- Implemented early cancellation detection that monitors signals immediately when functions are called
- Added new `LLMResponse::Cancelled` variant to handle cancelled operations
- All changes are backward compatible - existing code works unchanged

### Run the tests:
```bash
# Python tests
cd integ-tests/python && uv run pytest tests/test_abort_handlers.py -v

# TypeScript tests  
cd integ-tests/typescript && npm test -- tests/abort-handlers.test.ts

# Go tests (requires CFFI rebuild)
cd engine/language_client_cffi && cargo build
cd integ-tests/go && go test -v -run TestAbortHandler
```

### Test status:
- [x] Python abort handler tests (7 tests) - All passing
- [x] TypeScript abort handler tests (11 tests) - All passing
- [x] Go abort handler tests (7 tests) - All passing
- [x] I have ensured `make check test` passes

Also re-enabled 6 previously disabled test files (PDF, video, OpenAI responses, expression functions) that are now compatible.

Added comprehensive abort handler support for TypeScript, Python, and Go, allowing cancellation of in-flight LLM operations through native language cancellation patterns.


sxlijin:
thanks for PR! working on taking this over the finish line! https://github.com/BoundaryML/baml/pull/2357

PR #2357, opened 2025-08-21
feat: implement AbortController in py, ts, go, wasm (also cancel buttons)

sxlijin:
Implement AbortController in python, typescript, go, and also wasm. Also implement cancel functionality for run test in wasm and add docs for py/ts/go.

---

First 80% of this work done by @dexhorthy and @hellovai in #2331 :

Built in close collaboration with @hellovai who sat in my apartment and shouted at claude with me all day

*Note: plan/Research artifacts in thoughts/ can be removed once they get put in a good location, leaving them in for now*

## What problem(s) was I solving?

Previously, once a BAML function call was initiated, there was no way to cancel it. This meant:
- Long-running LLM operations couldn't be stopped if no longer needed
- Retry chains would continue even after the user navigated away
- Resources were wasted on unnecessary LLM calls
- Poor user experience when operations needed to be cancelled

## What user-facing changes did I ship?

Added abort/cancellation support for all BAML operations using native language patterns:

### TypeScript
```typescript
const controller = new AbortController();
const promise = b.ExtractName("John Doe", { 
  abortController: controller 
});

// Cancel after 100ms
setTimeout(() => controller.abort(), 100);

try {
  await promise;
} catch (e) {
  if (e instanceof BamlAbortError) {
    console.log("Operation cancelled");
  }
}
```

### Python
```python
from baml_py import AbortController

controller = AbortController()
task = b.ExtractName("John Doe", 
  baml_options={"abort_controller": controller})

# Cancel after 100ms
await asyncio.sleep(0.1)
controller.abort()

try:
  await task
except Exception as e:
  if "abort" in str(e).lower():
    print("Operation cancelled")
```

### Go
```go
ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
defer cancel()

_, err := b.ExtractName(ctx, "John Doe")
if err != nil && strings.Contains(err.Error(), "context canceled") {
  fmt.Println("Operation cancelled")
}
```

## How I implemented it

- Added cancellation support at the orchestrator level in Rust runtime using `stream-cancel` crate's `Tripwire` mechanism
- Created language-specific bindings:
  - **TypeScript**: Native `AbortController` support via NAPI bridge
  - **Python**: Custom `AbortController` class exposed via PyO3
  - **Go**: `context.Context` cancellation with early detection goroutines
- Implemented early cancellation detection that monitors signals immediately when functions are called
- Added new `LLMResponse::Cancelled` variant to handle cancelled operations
- All changes are backward compatible - existing code works unchanged

### Run the tests:
```bash
# Python tests
cd integ-tests/python && uv run pytest tests/test_abort_handlers.py -v

# TypeScript tests  
cd integ-tests/typescript && npm test -- tests/abort-handlers.test.ts

# Go tests (requires CFFI rebuild)
cd engine/language_client_cffi && cargo build
cd integ-tests/go && go test -v -run TestAbortHandler
```

### Test status:
- [x] Python abort handler tests (7 tests) - All passing
- [x] TypeScript abort handler tests (11 tests) - All passing
- [x] Go abort handler tests (7 tests) - All passing
- [x] I have ensured `make check test` passes

Also re-enabled 6 previously disabled test files (PDF, video, OpenAI responses, expression functions) that are now compatible.

Added comprehensive abort handler support for TypeScript, Python, and Go, allowing cancellation of in-flight LLM operations through native language cancellation patterns.


trojanowski:
@sxlijin @hellovai  I see one problem with the API added to the TypeScript SDK in this PR. The operations expect an `AbortController` instance as a parameter, but in my opinion, `AbortSignal` would be a much better choice. This would be consistent with what `fetch` [expects](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#signal), as well as multiple Node.js APIs, such as several functions in the [node:fs module](https://nodejs.org/api/fs.html) (just try searching for `signal` there) and even [`addEventListener` from the DOM API](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal). Additionally, this would allow us to use static methods that return `AbortSignal` instances, like [`AbortSignal.timeout`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static) or [`AbortSignal.any`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/any_static), which isn't possible with the current implementation in this PR.

After making that change, the TypeScript snippet from the PR description could be simplified as follows:

```typescript
const promise = b.ExtractName("John Doe", {
  signal: AbortSignal.timeout(100)
});

try {
  await promise;
} catch (e) {
  // For consistency, I think we should also use the built-in `AbortError`
  // class here, instead of a custom one.
  if (e instanceof AbortError) {
    console.log("Operation cancelled");
  }
}
```

In my opinion, it would be a good idea to make these changes before a new release; otherwise, it could be a breaking change.

hellovai:
@trojanowski thanks for this! we actually patched it to be just that!

#2373 shows that off!

this was a really really great catch :)

trojanowski:
@hellovai that's really great, but for consistency with other APIs, I would suggest two additional changes:

1. Renaming the parameter from `abortSignal` to `signal`, like in all APIs mentioned in my previous comment.
2. Using the built-in `AbortError` class instead of a custom one, also for consistency with the Node.js APIs and `fetch`.

hellovai:
thanks @trojanowski 

> 1. Renaming the parameter from `abortSignal` to `signal`, like in all APIs mentioned in my previous comment.

Oh! interesting, we explicitly left it as `abortSignal` as its a bit more verbose / clear, but i could be convinced. let me run this back with the team!

> 2. Using the built-in `AbortError` class instead of a custom one, also for consistency with the Node.js APIs and `fetch`.

We sadly can't do this as there is no such thing as `AbortError` that user land has:
web world: Note: When abort() is called, the fetch() promise rejects with an Error of type DOMException, with name AbortError.
NodeJS: https://github.com/nodejs/node/issues/38361

PR #2483, opened 2025-09-19
hellovai/tb: TypeBuilder -> TypeRegistry

hellovai:
- **Migrate TypeScript TypeBuilder to use baml_type_builder::TypeBuilder**
- **Keep original method names while using new type builder**
- **it works**
- **everything compiles!**
- **python works!**


Haoxincode:
Any update?

PR #2487, opened 2025-09-20
bugfix: add import as a type import for TS code

hellovai:


aaronvg:
see instead https://github.com/BoundaryML/baml/pull/2489

PR #2508, opened 2025-09-23
feat: baml cli lint command

### LINKED ISSUE #2479: [feat] baml-cli lint
### Product

BAML

### Problem Statement / Use Case

Provide linting via `baml-cli`

### Proposed Solution

1. Add a `lint` command to `baml-cli`.

`baml-cli lint --from path/to/baml_project` should print out the list of compiler errors and warnings.

This printout should be the default display format for Diagnostics used in the playground.

It should exit with exitcode `1` if there are any warnings or errors, `0` if clean.

2. Add a new Diagnostic level beyond `DiagnosticError` and `DiagnosticWarning` called `DiagnosticLint`. This level would not be visible in the playground, only to the linter.

3. Add a `--fix` flag.

If the issues raised by the linter can be fixed automatically, write the change.

Note: 1 is the most immediate task, and we would like to merge a PR for that before we tackle 2 and 3.

### Alternative Solutions

_No response_

### Additional Context

_No response_

AntennaeVY:
Closes https://github.com/BoundaryML/baml/issues/2479

Notes:
`--only-errors` flag implementation is missing.

### DIFF
--- a/engine/baml-runtime/src/cli/check.rs
+++ b/engine/baml-runtime/src/cli/check.rs
@@ -6,12 +6,12 @@ use internal_baml_core::configuration::GeneratorDefaultClientMode;
 use crate::{baml_src_files, BamlRuntime, InternalRuntimeInterface};
 
 #[derive(clap::Args, Debug)]
-pub struct LintArgs {
+pub struct CheckArgs {
     #[arg(long, help = "path/to/baml_src", default_value = "./baml_src")]
     pub from: PathBuf,
     #[arg(
         long,
-        help = "Lint without checking for version mismatch",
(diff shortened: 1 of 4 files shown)


imalsogreg:
Looks great! Just waiting for CI.

imalsogreg:
@AntennaeVY This is great! Would you mind changing the command name from `lint` to `check`? What we're doing is more like checking than linting, since lints are more often about little pedantic things rather than compiler warnings/errors.

AntennaeVY:
@imalsogreg done in 94a015be2eca86ebc13c0b7b29adbf5c4e015594

aaronvg:
congrats on your first PR :)

PR #2608, opened 2025-10-13
generate python code and integrate with runtime

imalsogreg:
Codegen for emit in python


engine/generators/languages/python/src/watchers.rs
```diff
@@ -0,0 +1,307 @@
...
+};
+
+use crate::{
+    ir_to_py::{stream_type_to_py, type_to_py},
+    package::CurrentRenderPackage,
+    r#type::SerializeType,
+};
+
+#[derive(Debug, Clone)]
+pub struct VarEventPy {
```
aaronvg, line 24: no need for the Py suffix IMO

integ-tests/python/tests/test_emit.py
```diff
@@ -0,0 +1,175 @@
...
+    handler2_calls = []
+
+    def handler1(ev):
+        handler1_calls.append(ev)
+
+    def handler2(ev):
+        handler2_calls.append(ev)
+
+    # Register multiple handlers for the same variable
+    listener.on_var("x", handler1)
```
aaronvg, line 168: how typesafe are the events in these handlers? do you get type completion from them?

If i use a python lambda as my even tlistener callback, does type completion work well?

imalsogreg: The goal is fully typesafe and completion that shows up as soon as you start typing the arguments to `listener.on_something(`.

The code as it's implemented now does neither, but is incoming.

I did some experiments a while back to see what sorts of autocomplete are possible under which spellings of `listener.on_____`, and the results weren't promising with the TS style, so we'll probably have to go with `listener.vars.VAR_NAME`. ... `listener.on_var("x", ...)` gave me good typechecking (so you do get type errors if you try to handle the wrong variable type), but for some reason type inference didn't work (hovering over the `ev` in `listener.on_var("x", lambda ev: handle_ev(ev))` didn't give any information about `ev` ).

PR #2618, opened 2025-10-16
Fix/instanceof operator on primitives

AntennaeVY:
Fixes typechecking of `instanceof` operator only working on classes. Now working on primitive types and enums

Before these gave an error
```
1 instanceof int
x instanceof MyEnum
```

Now the typechecker should properly check these are valid


imalsogreg:
Thanks for this! Typechecker side looks perfect! As discussed on Discord, we just need the interpreter side implemented, in order to merge this 🎉

sxlijin:
[automated] closing old PRs.

PR #2712, opened 2025-11-10
Add project URLs to Pypi

Toundra:
Hi!
The pypi distribution currently has no project URLs metadata https://pypi.org/project/baml-py/, so i wanted to add it - according to PEP-621.

https://peps.python.org/pep-0621/


sxlijin:
Thanks for this! Really appreciate it <3 (sorry for the slow response)

PR #2723, opened 2025-11-17
Baml `salsa` HIR

antoniosarosi:


baml_language/crates/baml_hir/src/item_tree.rs
```diff
@@ -0,0 +1,263 @@
+//! Position-independent item storage.
+//!
+//! The `ItemTree` contains minimal representations of all items in a container.
+//! It acts as an "invalidation barrier" - only changes to item signatures
+//! cause the `ItemTree` to change, not edits to whitespace, comments, or bodies.
```
imalsogreg, line 5: Curious - why don't changes to bodies (does this mean function bodies) cause the function to get invalidated?

imalsogreg: Oh - function bodies are stored elsewhere apparently https://github.com/BoundaryML/baml/pull/2723/files#diff-6cdcf2d13bbd1882d24da94627a7b16e3442133c617f96c7f9d9735701814abcR117

baml_language/crates/baml_hir/src/lib.rs
```diff
@@ -1,85 +1,469 @@
...
+#[salsa::db]
+pub trait Db: salsa::Database {}
+
+//
+// ───────────────────────────────────────────────────── TRACKED STRUCTS ─────
+//
+
+/// Tracked struct holding all items defined in a file.
+///
+/// This follows the Salsa 2022 pattern: instead of returning Vec<ItemId<'db>>
```
imalsogreg, line 66: (nit - we can remove some of the redundant explanations)

antoniosarosi: True, I would say we do intensive human review after we're done

baml_language/crates/baml_hir/src/path.rs
```diff
@@ -0,0 +1,75 @@
+//! Path representation for name resolution.
```
imalsogreg, line 1: Let's chat about this module before merging - want to make sure it's in line with future thoughts about our module system. The safest thing IMO is to delete it and revisit after we design modules.

antoniosarosi: There are a couple things we support right now:

- `baml.unstable.string(x)` or even `baml.deep_copy(x)`
- `baml.HttpMethod.Post`

Currently we parse them as `FieldAccess` and then later figure out they are actually function calls on a pseudo-module (`baml`) or enum access (`Post`).

Right way of doing this is parse it as a path and then resolve names & types to actually determine what they are. So I think `Path` should be kept in the HIR.

imalsogreg: Keeping would be Ok, we'd just want to be careful not to import Rust module system assumptions (like `Super` and `Absolute` PathKind variants, and the assumption that modules other than `baml` exist, or that files themselves represent modules), and we'd want to sanitize the doc comments to make them less Rust-y. I can push that change.

PR #2734, opened 2025-11-21
Fix metadata on tracing BamlValues

imalsogreg:
`BamlValue`s going into the tracing system previously attached type data with `.with_const_meta()`, which attaches a fresh copy of the metadata at every node of the BamlValue (including nested nodes), which was incorrectly typing those inner nodes, and also producing explosions in the BamlValue size for complex nested values with many leaves, because each leaf carries the type information of the big composite value.

This change replaces the use of `.with_const_meta()` with `.infer_value_with_type()`, which is similar to `.distribute_type_with_meta()` but doesn't require a type hint (it only does inference, not checkinging), and attaches the correct (small) types to leaves in deeply nested BamlValues.

The new integ test runs tracing on this Python value:

```python
{"a": 10, "b": { "c": 20, "d": "hi"}, "e": [1, 2, "hi", {"f": 30, "g": "hello"}]}
```

The output tracing file under the changes in this PR for the above value drops from 34k to 11k.

Bonus change: renamed `.with_const_meta()` to the something less confusing: `.with_same_meta_at_all_nodes()`.

Todo:
 - [x] cleanup
 - [x] integ tests


aaronvg:
continuing on https://github.com/BoundaryML/baml/pull/2729

PR #2749, opened 2025-11-25
THIR

imalsogreg:


baml_language/crates/baml_thir/src/lower.rs
```diff
@@ -0,0 +1,183 @@
...
+                "int" | "Int" => return Ty::Int,
+                "float" | "Float" => return Ty::Float,
+                "string" | "String" => return Ty::String,
+                "bool" | "Bool" | "boolean" | "Boolean" => return Ty::Bool,
+                "null" | "Null" | "None" => return Ty::Null,
+                "image" | "Image" => return Ty::Image,
+                "audio" | "Audio" => return Ty::Audio,
+                "video" | "Video" => return Ty::Video,
+                "pdf" | "Pdf" | "PDF" => return Ty::Pdf,
+                _ => {}
```
antoniosarosi: @imalsogreg I suppose `Int` shouldn't be inferred as a primitive `int`?

baml_language/crates/baml_hir/src/body.rs
```diff
@@ -592,38 +691,141 @@ impl LoweringContext {
...
+                .children_with_tokens()
+                .filter_map(baml_syntax::NodeOrToken::into_token)
+                .find(|t| t.kind() == SyntaxKind::WORD);
+
+            if let Some(token) = word_token {
+                let name = token.text().to_string();
+                self.exprs.alloc(Expr::Path(Name::new(&name)))
+            } else {
+                self.exprs.alloc(Expr::Missing)
+            }
```
antoniosarosi, line 646: This seems to be using the CST directly instead of AST accessors @imalsogreg

baml_language/crates/baml_thir/src/lower.rs
imalsogreg: Absolutely! Fixed.

baml_language/crates/baml_hir/src/body.rs
imalsogreg: 👍  Moved it to ast.rs

PR #2803, opened 2025-12-12
fix: Undocumented null return for reasoning content

PatrickE94:
OpenAI documents the content to always be a sequence but often returns null instead. Seen when using gpt-5-mini on OpenAI and gpt-oss-120b via OpenRouter.

Serde is strict around parsing and won't consider null to be an empty sequence. With a custom deserialize function, this is trivially solved.


aaronvg:
thanks for the contribution! It's a bummer the APIs are so inconsistent!

PatrickE94:
I can only agree! Thankfully it was quite a quick fix :)

PR #2812, opened 2025-12-15
Add source span tracking to HIR and type checker 

jmcardon:
Introduce span tracking infrastructure throughout the expression lowering and type checking pipeline to enable accurate source location reporting in type errors.

Changes:
- Add inline source maps to ExprBody (expr_spans, stmt_spans, pattern_spans) using FxHashMap for O(1) lookups
- Create alloc_expr/alloc_stmt/alloc_pattern helpers that record spans during HIR lowering
- Store type annotation spans in Stmt::Let for type mismatch diagnostics
- Thread FileId through TypeContext via FunctionLoc parameter
- Add build_span/build_span_default helpers to construct Span from TextRange
- Replace hardcoded placeholder spans (FileId(0), empty range) with actual source locations in type errors
- Add Default derives to FileId and Span for fallback when spans unavailable

This enables the type checker to report errors pointing to the actual source location rather than synthetic positions, improving developer experience when debugging type mismatches.


baml_language/crates/baml_hir/src/type_ref.rs
```diff
@@ -84,6 +84,10 @@ impl TypeRef {
     }
 
     /// Create a `TypeRef` from an AST `TypeExpr` node.
+    /// TODO: it seems that type parsing is occurring here!!
+    /// I believe that the parser should be what resolves types syntactically
+    /// this code is brittle (e.g, try the type int?? or int[][]). it will be resolved
+    /// as a type path when it should be either incorrect syntax, or something else.
```
antoniosarosi: Why would `int[][]` be incorrect syntax here? Isn't it just list of lists of `int`.

Then `int??` may seem like incorrect syntax but thinking about type aliases cases like:

```ts
type Example = int?;

type Valid = Example?;
```

I'd say it shouldn't fail.

PR #2825, opened 2025-12-17
Match

rossirpaulo:
This is the first pass on `match`s implementation. I've tried to make sure there are plenty of comments referencing the BEP shared yesterday on critical moments; expect a cleanup later on. 

There are somethings to keep in mind:

### Core Infrastructure
| Component | Status | Location |
|-----------|--------|----------|
| Lexer tokens | Done | TokenKind::Match, FatArrow, Pipe |
| Syntax kinds | Done | syntax_kind.rs |
| Parser | Done | parser.rs - parse_match_expr, parse_match_arm, etc. |
| HIR | Done | body.rs - Expr::Match, MatchArm, Pattern variants |
| AST helpers | Done | ast.rs |

### Type Checking
| Feature | Status | Location |
|---------|--------|----------|
| Pattern variable binding | Done | lib.rs:781-785 - ctx.define(name, narrowed_ty) |
| Type narrowing | Done | lib.rs:834-855 - extract_pattern_binding returns narrowed type |
| Exhaustiveness checking | Done | lib.rs:1011-1121 - check_match_exhaustiveness |
| Unreachable arm detection | Done | Same function, detects arms after catch-all |

### Phase 3: Code Generation
| Feature | Status | Location |
|---------|--------|----------|
| instanceof codegen | Done | compiler.rs:913-943 - compile_match_pattern emits CmpOp::InstanceOf |
| Pattern binding | Done | compiler.rs:1047-1061 - bind_pattern_variable |
| Guard handling | Done | compiler.rs:518-535 |
| If-else desugaring | Done | Full match-to-if-else chain implemented |

### Tests (PARTIAL)
| Test Type | Status | Location |
|-----------|--------|----------|
| VM integration tests | Done | match_expr.rs - 20+ tests covering bindings, literals, typed patterns, guards |
| Exhaustiveness tests | Done | match_exhaustiveness/ |

### Type System Improvements
**1. Type Alias Expansion**
Current: Type aliases like type `Result = Success | Failure` are treated as opaque `Ty::Named("Result")`.
Required: Expand type aliases to their definitions for proper **union** exhaustiveness checking.

Without this, exhaustiveness checking doesn't work for type alias unions.

**2. Enum Variant Exhaustiveness**
Current: Enum variant patterns (`Status.Active`) are treated as literals, not contributing to exhaustiveness.

Required: Track which enum variants are covered and error if variants are missing.

```baml
enum Status { Active, Inactive, Pending }
match (s) {
  Status.Active => "active"
  // ERROR: Inactive and Pending not handled
}
```

**3. Literal Type Exhaustiveness**
Current: Literal patterns don't contribute to type-level exhaustiveness.

Required: For literal union types (`200 | 201 | 204`), track which values are covered.

### Refinements
**1. Overlapping Pattern Verification**
The BEP says this should work (second arm only matches B, not unreachable):

```baml
match (x) {
  _: A => "a"
  _: A | B => "ab"  // Should NOT be flagged as unreachable
}
```
Status: Likely works, but needs explicit test coverage.

**2. Better Error Spans**
Current: Error spans point to the match expression, not the specific arm.

Required: Point to the specific arm that's unreachable or the missing cases.


baml_language/crates/baml_thir/src/lib.rs
```diff
@@ -751,13 +751,387 @@ fn infer_expr<'db>(ctx: &mut TypeContext<'db>, expr_id: ExprId, body: &ExprBody)
             }
         }
 
+        Expr::Match { scrutinee, arms } => {
+            let scrutinee_ty = infer_expr(ctx, *scrutinee, body);
+
+            if arms.is_empty() {
+                // Empty match is non-exhaustive (unless scrutinee is uninhabited)
+                if !scrutinee_ty.is_unknown() {
```
imalsogreg, line 759: Comment is right. I think there are more ways to be uninhabited than just being `is_unknown()`.

baml_language/crates/baml_thir/src/lib.rs
```diff
@@ -751,13 +751,387 @@ fn infer_expr<'db>(ctx: &mut TypeContext<'db>, expr_id: ExprId, body: &ExprBody)
...
+                        // Type-check the arm body
+                        let body_ty = infer_expr(ctx, arm.body, body);
+
+                        ctx.pop_scope();
+                        body_ty
+                    })
+                    .collect();
+
+                // If all arms have the same type, use that; otherwise union
+                if arm_types.iter().all(|t| t == &arm_types[0]) {
```
imalsogreg, line 809: Use `.unwrap_or` (uses default) to avoid `.unwrap()` (which panics).

The code as written can never panic, but proving it requires thought at reading time. (so this is a comment nit).

In other words, avoid `unwrap` when possible.

baml_language/crates/baml_thir/src/lib.rs
```diff
@@ -751,13 +751,387 @@ fn infer_expr<'db>(ctx: &mut TypeContext<'db>, expr_id: ExprId, body: &ExprBody)
             }
         }
 
+        Expr::Match { scrutinee, arms } => {
```
imalsogreg, line 754: Worth putting a comment at the top of this branch: `// Match expressions synthesize a type`.

At some point, we should also add a function that checks matches, because bidirectional typing switches from synthesis to checking in certain circumstances:

```
/// This should _check_ the match branches, rather than synthesizing.
let x: int = match (true) {
  true => 0,
  false => 1,
}

File: baml_language/crates/baml_thir/src/lib.rs [hunk 4]
```diff
@@ -751,13 +751,387 @@ fn infer_expr<'db>(ctx: &mut TypeContext<'db>, expr_id: ExprId, body: &ExprBody)
...
+            let mut cases = expand_type_to_cases(inner);
+            cases.push(Ty::Null);
+            cases
+        }
+        _ => vec![ty.clone()],
+    }
+}
+
+/// Check if a type is covered by a pattern coverage.
+#[allow(dead_code)]
```
imalsogreg, line 966: Where's the dead code?

baml_language/crates/baml_thir/src/lib.rs
rossirpaulo: Makes sense. Changed to `arm_types.into_iter().next().unwrap_or(Ty::Unknown)`

baml_language/crates/baml_thir/src/lib.rs
rossirpaulo: Yep. I'll have to implement something similar to finish the remaining work, but I'll remove this for now.

PR #2826, opened 2025-12-17
Loop desugaring and builtins type checking (`baml.Array.length`, etc)

antoniosarosi:


baml_language/crates/baml_hir/src/pretty.rs
```diff
@@ -0,0 +1,383 @@
+//! Pretty printing HIR as code.
```
imalsogreg, line 1: Maybe as a TODO, can we implement this in terms of `pretty`? (uniformity with the other pretty printers in the codebase)

antoniosarosi: Yeah, I just wanted to verify desugaring, but sure we should use some crate that does this better.

PR #2828, opened 2025-12-17
Match (updated)

rossirpaulo:
> [!IMPORTANT]
> Supersedes #2825 
> I renamed the `paulo/map` branch to `paulo/match` and had to recreate this PR. The original is referenced above to preserve the original comments.

This is the first pass on `match`s implementation. I've tried to make sure there are plenty of comments referencing the BEP shared yesterday on critical moments; expect a cleanup later on. 

There are somethings to keep in mind:

### Core Infrastructure
| Component | Status | Location |
|-----------|--------|----------|
| Lexer tokens | Done | TokenKind::Match, FatArrow, Pipe |
| Syntax kinds | Done | syntax_kind.rs |
| Parser | Done | parser.rs - parse_match_expr, parse_match_arm, etc. |
| HIR | Done | body.rs - Expr::Match, MatchArm, Pattern variants |
| AST helpers | Done | ast.rs |

### Type Checking
| Feature | Status | Location |
|---------|--------|----------|
| Pattern variable binding | Done | lib.rs:781-785 - ctx.define(name, narrowed_ty) |
| Type narrowing | Done | lib.rs:834-855 - extract_pattern_binding returns narrowed type |
| Exhaustiveness checking | Done | lib.rs:1011-1121 - check_match_exhaustiveness |
| Unreachable arm detection | Done | Same function, detects arms after catch-all |

### Phase 3: Code Generation
| Feature | Status | Location |
|---------|--------|----------|
| instanceof codegen | Done | compiler.rs:913-943 - compile_match_pattern emits CmpOp::InstanceOf |
| Pattern binding | Done | compiler.rs:1047-1061 - bind_pattern_variable |
| Guard handling | Done | compiler.rs:518-535 |
| If-else desugaring | Done | Full match-to-if-else chain implemented |

### Tests (PARTIAL)
| Test Type | Status | Location |
|-----------|--------|----------|
| VM integration tests | Done | match_expr.rs - 20+ tests covering bindings, literals, typed patterns, guards |
| Exhaustiveness tests | Done | match_exhaustiveness/ |

### Type System Improvements
**1. Type Alias Expansion**
Current: Type aliases like type `Result = Success | Failure` are treated as opaque `Ty::Named("Result")`.
Required: Expand type aliases to their definitions for proper **union** exhaustiveness checking.

Without this, exhaustiveness checking doesn't work for type alias unions.

**2. Enum Variant Exhaustiveness**
[...]

PR #2838, opened 2025-12-18
Match Exhaustiveness Checking

rossirpaulo:
> [!IMPORTANT]  
> Based on #2828 

## Summary

This PR adds **exhaustiveness checking** and **unreachable arm detection** for match expressions in BAML's type system. The implementation uses a value-based model (`ValueSet`) inspired by the treatment of pattern matching in TAPL.

**Key capabilities:**
- Detect non-exhaustive matches (missing cases) with helpful error messages listing uncovered patterns
- Detect unreachable arms (arms that can never match because previous arms already cover them)
- Handle type aliases by expanding them to their underlying types
- Treat enums as finite types, expanding them to their individual variants
- Support literal patterns (integers, strings, booleans) and union patterns (`200 | 201`)

### 1. Value-Based Model (ValueSet)

The core insight is that **pattern matching operates on values, not types**. A pattern like `Status.Active` matches one specific value, while `s: Status` matches all values of type Status.

The `ValueSet` enum captures this:

```rust
pub enum ValueSet {
    All,                                    // Catch-all: `_` or `other`
    OfType(Name),                           // Type pattern: `s: Status`
    EnumVariant { enum_name, variant_name }, // Specific variant: `Status.Active`
    Literal(Literal),                       // Literal value: `42`, `"hello"`
    Union(Vec<ValueSet>),                   // Union pattern: `200 | 201`
    Empty,                                  // Guarded pattern (doesn't contribute)
}
```

This separation allows clean reasoning about what values each arm covers.

### 2. Guards Do NOT Contribute to Exhaustiveness

Per BEP-002, guarded patterns are treated as `Empty` value sets. A pattern like `x: int if x > 0` might not match negative numbers, so it can't guarantee coverage.

```baml
match (n) {
  x: int if x > 0 => "positive"
  // Still need a catch-all because guard doesn't cover all ints
}
```

### 3. Type Alias Expansion

Type aliases are expanded during exhaustiveness checking. For `type Result = Success | Failure`, matching on `r: Result` requires covering both `Success` and `Failure`.

### 4. Finite vs. Infinite Types

- **Finite types** (enums, booleans): Expanded into individual values for precise tracking
- **Infinite types** (int, string, classes): Represented as abstract `OfType` sets, only exhaustive via catch-all

### 5. Singleton Types for Literal Unions

Added singleton types (`Ty::Singleton(SingletonValue)`) to support exhaustiveness checking of literal unions like `200 | 201 | 204`. This follows the type-theoretic notion of singleton types—types inhabited by exactly one value.

## Files Changed

| File | Purpose |
|------|---------|
| `baml_thir/src/exhaustiveness.rs` | New module: `ValueSet` abstraction, `ExhaustivenessChecker`, coverage algorithms |
| `baml_thir/src/lib.rs` | Integration: Calls exhaustiveness checker during match type inference |
| `baml_thir/src/types.rs` | Added `Ty::Singleton`, `SingletonValue`, and `is_uninhabited()` method |
| `baml_hir/src/type_ref.rs` | Added `TypeRef::StringLiteral`, `TypeRef::IntLiteral`, `TypeRef::BoolLiteral` for literal type support |
| `baml_hir/src/lib.rs` | Type text parsing for literal types in union syntax |

## Algorithm Overview

The exhaustiveness checker maintains a list of covered `ValueSet`s as it processes arms:

1. **Expand scrutinee type** into required value sets
   - Enums → individual variants
   - Optionals → inner type + null
   - Type aliases → expanded recursively

2. **For each arm**, convert pattern to `ValueSet`:
   - Check if arm is unreachable (already fully covered by previous arms)
   - Add to coverage list (unless guarded)

3. **After all arms**, find uncovered value sets:
   - Report as `TypeError::NonExhaustiveMatch` with helpful error message

## Test Coverage

The implementation includes unit tests for:

- `ValueSet` display formatting
- Coverage logic (type patterns, enum variants, literals)
- Type alias expansion (union members covered separately)
- Enum exhaustiveness (type pattern expands to all variants)
- Literal-to-type coverage (`42` is covered by `int` pattern)
- Catch-all coverage (covers everything)

### Error Span Improvements

Currently, unreachable arm errors point to the entire match expression rather than the specific unreachable arm. This requires tracking arm spans in HIR.

```rust
// TODO(exhaustiveness): Improve error spans to point to the specific unreachable
// arm rather than the entire match expression. Requires tracking arm spans in HIR.
```

### Uninhabited Types (Zero-Variant Enums)

Empty matches on uninhabited types should be allowed (there are no cases to handle). Currently only empty unions are detected; zero-variant enums require database access.

```rust
// TODO(exhaustiveness): Check for zero-variant enums. This requires database
// access to look up enum definitions. Currently only empty unions are detected.
```

### Cyclic Type Alias Detection

Recursive type aliases like `type A = A | B` could cause infinite recursion during expansion. A cycle detection mechanism should be added.

```rust
// TODO(exhaustiveness): Add cycle detection for recursive type aliases like
// `type A = A | B`. Currently this would cause infinite recursion.
```

## Example Usage

```baml
enum Status { Active, Inactive, Pending }

function ProcessStatus(s: Status) -> string {
  match (s) {
    Status.Active => "active"
    Status.Inactive => "inactive"
    // Error: Non-exhaustive match, missing: Status.Pending
  }
}

function WithCatchAll(s: Status) -> string {
  match (s) {
    Status.Active => "active"
    _ => "other"  // OK: catch-all covers remaining cases
  }
}

type Result = Success | Failure

function HandleResult(r: Result) -> string {
  match (r) {
    s: Success => "ok"
    f: Failure => "error"
    // OK: both union members covered
  }
}
```


baml_language/crates/baml_hir/src/lib.rs
```diff
@@ -667,14 +667,68 @@ fn lower_test(node: &SyntaxNode) -> Option<Test> {
...
+    if has_pipe && parts.len() > 1 {
+        let members: Vec<TypeRef> = parts.iter().map(|p| lower_type_text(p)).collect();
+        return TypeRef::Union(members);
+    }
+
+    // Otherwise, lower as a single type
+    let text = syntax.text().to_string();
     let text = text.trim();
+    lower_type_text(text)
+}
```
antoniosarosi: One suggestion I have here is to make this code that works on top of CST syntax nodes an AST accessor in `baml_language/crates/baml_syntax/src/ast.rs` so that HIR lowering is not as manual as checking for `PIPE` tokens.

You can see examples of AST accessor usage right above this `lower_type_ref` function:

```rust
/// Extract enum definition from CST.
fn lower_enum(node: &SyntaxNode) -> Option<Enum> {
    use baml_syntax::ast::EnumDef;

    let enum_def = EnumDef::cast(node.clone())?;

    // Check if the enum has proper structure (braces)
    // Malformed enums from error recovery (e.g., "enum" without name/braces) should be skipped
    if !enum_def.has_body() {
        return None;
    }

    // Extract name using AST accessor
    let name = enum_def
        .name()
        .map(|t| Name::new(t.text()))
        .unwrap_or_else(|| Name::new("UnnamedEnum"));

    // Extract variants using AST accessor
    let variants = enum_def
        .variants()
        .filter_map(|variant| {
            variant.name().map(|name_token| crate::EnumVariant {
                name: Name::new(name_token.text()),
            })
        })
        .collect();

    Some(Enum { name, variants })
}
```

Corresponding AST node in `ast.rs`:

```rust
impl EnumDef {
    /// Get the enum name.
    pub fn name(&self) -> Option<SyntaxToken> {
        self.syntax
            .children_with_tokens()
            .filter_map(rowan::NodeOrToken::into_token)
            .filter(|token| {
                token.kind() == SyntaxKind::WORD && token.parent() == Some(self.syntax.clone())
            })
            .nth(0) // Get the first WORD (enum keyword is KW_ENUM, not WORD)
    }

    /// Check if this enum has a body (braces).
    /// Malformed enums from error recovery may not have braces.
    pub fn has_body(&self) -> bool {
        self.syntax
            .children_with_tokens()
            .filter_map(rowan::NodeOrToken::into_token)
            .any(|t| t.kind() == SyntaxKind::L_BRACE)
    }

    /// Get all variants.
    pub fn variants(&self) -> impl Iterator<Item = EnumVariant> {
        self.syntax.children().filter_map(EnumVariant::cast)
    }

    /// Get block attributes.
    pub fn block_attributes(&self) -> impl Iterator<Item = BlockAttribute> {
        self.syntax.children().filter_map(BlockAttribute::cast)
    }
}
```

baml_language/crates/baml_hir/src/lib.rs
```diff
@@ -687,16 +741,35 @@ pub fn lower_type_ref(node: &baml_syntax::ast::TypeExpr) -> TypeRef {
         "video" => TypeRef::Video,
         "pdf" => TypeRef::Pdf,
         _ => {
+            // Check for string literal types
+            if (text.starts_with('"') && text.ends_with('"'))
+                || (text.starts_with('\'') && text.ends_with('\''))
```
antoniosarosi: In our current syntax single quote strings don't work: `'literal'` is not valid.

baml_language/crates/baml_hir/src/type_ref.rs
```diff
@@ -84,27 +88,105 @@ impl TypeRef {
...
+            return TypeRef::BoolLiteral(false);
+        }
+
+        // Check for integer literal types (for exhaustiveness like 200 | 201)
+        if let Ok(int_val) = text.parse::<i64>() {
+            return TypeRef::IntLiteral(int_val);
+        }
+
+        Self::from_type_name(text)
     }
```
antoniosarosi, line 173: All this logic here seems to be duplicated from `fn lower_type_ref` and `fn lower_type_text` from `baml_hir/src/lib.rs`

baml_language/crates/baml_hir/src/lib.rs
```diff
@@ -687,16 +741,35 @@ pub fn lower_type_ref(node: &baml_syntax::ast::TypeExpr) -> TypeRef {
...
+            {
+                // String literal type; extract the value
+                let inner = &text[1..text.len() - 1];
+                return TypeRef::StringLiteral(inner.to_string());
+            }
```
antoniosarosi: This seems to be an early return that cuts recursion for string literals, I think it will not handle the case where string literals are optional:
`"literal"?`

baml_language/crates/baml_hir/src/lib.rs
```diff
@@ -687,16 +741,35 @@ pub fn lower_type_ref(node: &baml_syntax::ast::TypeExpr) -> TypeRef {
...
+            // Check for boolean literal types
+            else if text == "true" {
+                TypeRef::BoolLiteral(true)
+            } else if text == "false" {
+                TypeRef::BoolLiteral(false)
+            }
+            // Check for integer literal types (for exhaustiveness)
+            else if let Ok(int_val) = text.parse::<i64>() {
+                TypeRef::IntLiteral(int_val)
+            }
```
antoniosarosi: We need to check for numbers that are greater than `i64::MAX` and issue diagnostics.

baml_language/crates/baml_hir/src/lib.rs
```diff
@@ -687,16 +741,35 @@ pub fn lower_type_ref(node: &baml_syntax::ast::TypeExpr) -> TypeRef {
...
             // Otherwise treat as named type
             else {
                 TypeRef::named(text.into())
```
antoniosarosi: If the `text.parse::<i64>()` call above fails you run the `else` branch returning a named type, but `i64` parsing can fail even with numbers. An example is overflow:

```rust
let text = "9999999999999999999";
let n = text.parse::<i64>();
```

You get:

```
Err(ParseIntError { kind: PosOverflow })
```

So you'd return `TypeRef::named("9999999999999999999")` if that happens.

antoniosarosi: I think float literals are not handled here and they are treated as `TypeRef::Named`?

baml_language/crates/baml_thir/src/types.rs
```diff
@@ -5,6 +5,41 @@ use std::fmt;
...
+/// In type theory, singleton types are types inhabited by exactly one value.
+/// For example, `SingletonValue::Int(42)` represents the value `42` which
+/// defines the singleton type `{42}` — the type whose only inhabitant is `42`.
+///
+/// This enum unifies all "specific value" types:
+/// - Primitive literals: `42`, `"hello"`, `true`
+/// - Enum variants: `Status.Active` (also singleton types with exactly one value)
+///
+/// Note: Float values are intentionally excluded because floating-point
+/// equality is problematic (NaN != NaN, -0.0 == 0.0, etc.).
```
antoniosarosi: In your PR description you mentioned you will remove comments before merging, but comments like this one explaining terms / concepts / reasons why something is excluded should stay.

Remove only useless comments introduced by AI like:

```rust
// Check if is optional
if let TypeRef::Optional(_) = t {
```

(haven't seen this in the PR just an example)

baml_language/crates/baml_thir/src/types.rs
```diff
@@ -21,6 +56,11 @@ pub enum Ty<'db> {
     Video,
     Pdf,
 
+    /// Singleton type: a type inhabited by exactly one value.
+    /// Used for exhaustiveness checking of literal unions like `200 | 201 | 204`.
+    /// The singleton type `{42}` contains only the value `42`.
+    Singleton(SingletonValue),
```
antoniosarosi: "Singleton" seems to be a term to refer to _any_ type that can have only one value from a set of values. `Ty::Null` is also a "singleton" by this definition.

Wouldn't it be clearer if this was named `Ty::Literal` instead? And then you use the singleton term in comments in the code that checks exhaustiveness.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+    EnumVariant { enum_name: Name, variant_name: Name },
+
+    /// Matches exactly one literal value.
+    ///
+    /// # Examples
+    /// - `42` -> `Literal(Int(42))`
+    /// - `"hello"` -> `Literal(String("hello"))`
+    /// - `true` -> `Literal(Bool(true))`
+    /// - `null` -> `Literal(Null)`
+    Literal(Literal),
```
antoniosarosi, line 96: Here I think it would make more sense to use the term "Singleton" for these two as opposed to using it in the general `Ty` struct. The meaning of singleton is important in the code that checks exhaustiveness with value sets.

Though this can stay as is, looks pretty clear to me without having to read term definitions.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+/// constituent values and check coverage.
+pub struct ExhaustivenessChecker<'a, 'db> {
+    /// Database for type resolution
+    db: &'db dyn Db,
+
+    /// Enum definitions: enum_name -> [variant_names]
+    enum_variants: &'a HashMap<Name, Vec<Name>>,
+
+    /// Type alias definitions: alias_name -> underlying_type
+    type_aliases: &'a HashMap<Name, Ty<'db>>,
```
antoniosarosi, line 163: Ah yes thank you for using lifetimes and not `.clone()`ing everything everywhere (bad tendency of AI models 🙃)

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+#[derive(Debug)]
+pub struct ExhaustivenessResult {
+    /// Whether all cases are covered
+    pub is_exhaustive: bool,
+
+    /// Value sets that are not covered (empty if exhaustive)
+    pub uncovered: Vec<ValueSet>,
+
+    /// Indices of unreachable arms (arms that can never match)
+    pub unreachable_arms: Vec<usize>,
```
antoniosarosi, line 176: What is this `usize`? Position of the unreachable arm? Brief comment would help.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+            db,
+            enum_variants,
+            type_aliases,
+        }
+    }
+
+    /// Check exhaustiveness of a match expression.
+    ///
+    /// # Arguments
+    /// - `scrutinee_ty`: The type of the value being matched
```
antoniosarosi, line 196: Keep as much of these as possible, these comments are the useful ones. If someone unfamiliar with the terms in this domain sees the code it makes it much easier to understand what's going on without Google searches / AI.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+                ValueSet::Literal(Literal::Bool(false)),
+            ],
+
+            // Null is a singleton
+            Ty::Null => vec![ValueSet::Literal(Literal::Null)],
+
+            // Singleton types (literal types like `200` in `200 | 201`)
+            Ty::Singleton(sv) => {
+                use crate::SingletonValue;
+                match sv {
```
antoniosarosi: This is what I was mentioning before about the naming of `Ty::Singleton`, it's a bit weird that `Null` is a singleton but is outside of `Singleton` enum. I'd do:

```rust
// Singleton types (types that contain a single value)

Ty::Null => { ... }

Ty::Literal(literal) => match literal {
    Literal::String(s) => { ... }
    Literal::Int(i) => { ... }
    // ...
}
```

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+                    }
+                    SingletonValue::Bool(v) => vec![ValueSet::Literal(Literal::Bool(*v))],
+                    SingletonValue::EnumVariant(enum_name, variant_name) => {
+                        vec![ValueSet::EnumVariant {
+                            enum_name: enum_name.clone(),
+                            variant_name: variant_name.clone(),
+                        }]
+                    }
+                }
+            }
```
antoniosarosi: Here you can remove one level of nesting if you put the `use::crate::SingletonValue` on top of the file with the rest of imports or at the beginning of this function, and code ends up flatter:

```rust
Ty::Singleton(sv) => match sv {
    SingletonValue::Int(v) => { ... }
    SingletonValue::String(v) => { ... }
    // ...
}
```

baml_language/crates/baml_thir/src/exhaustiveness.rs
antoniosarosi: I assume `sv` means "single value" but since it's not a standard abbreviation I would straight up use `single_value` or `Singleton(value)` already tells me enough 🙂

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+                }
+                values
+            }
+
+            // Named type: could be enum, class, or type alias
+            Ty::Named(name) => {
+                // Check if it's a type alias
+                if let Some(alias_ty) = self.type_aliases.get(name) {
+                    return self.expand_type_to_values(alias_ty);
+                }
```
antoniosarosi, line 329: Couple things:

1. Type alias resolution needs its own phase that runs only once, probably after name resolution, because resolved type aliases are used in multiple places:
    - Bytecode generation
    - Target language codegen (TS, Python, Go, ...)
    - Prompt rendering
And we can't resolve them every single time.

2. This needs a giant `// TODO` comment for recursive type aliases. Recursive type aliases are their own beast and we've solved this problem in the legacy compiler last year. I'll link the PRs for reference:
    - PR [#1163](https://github.com/BoundaryML/baml/pull/1163)
    - PR [#1207](https://github.com/BoundaryML/baml/pull/1207)
    - PR [#1416](https://github.com/BoundaryML/baml/pull/1416)

This is it's own different task though, porting this to new compiler isn't easy, but we'll have to if we want feature parity. For now adding `// TODO` deals with it :)

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+            Ty::Int => vec![ValueSet::OfType(Name::new("int"))],
+            Ty::Float => vec![ValueSet::OfType(Name::new("float"))],
+            Ty::String => vec![ValueSet::OfType(Name::new("string"))],
+            Ty::Image => vec![ValueSet::OfType(Name::new("image"))],
+            Ty::Audio => vec![ValueSet::OfType(Name::new("audio"))],
+            Ty::Video => vec![ValueSet::OfType(Name::new("video"))],
+            Ty::Pdf => vec![ValueSet::OfType(Name::new("pdf"))],
+            Ty::Class(_) => vec![ValueSet::OfType(Name::new("<class>"))],
+            Ty::Enum(_) => vec![ValueSet::OfType(Name::new("<enum>"))],
+            Ty::List(_) => vec![ValueSet::OfType(Name::new("<list>"))],
```
antoniosarosi: Shouldn't these be the names of the classes and enums?

```rust
Ty::Class(cls) => vec![ValueSet::OfType(Name::new(cls.name()))],
Ty::Enum(enm) => vec![ValueSet::OfType(Name::new(enm.name()))],
```

For the list case I suppose you can generate a unique name adding brackets? `ClassName[]`

If using `"<class>"` and `"<enum>"` is fine because this is treated as some generic class and not particular `ClassName` then add a comment explaining why.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+            Ty::Map { .. } => vec![ValueSet::OfType(Name::new("<map>"))],
```
antoniosarosi, line 398: Does this not distinguish between maps of different types?
- `map<string, int>` 
- `map<string, string>`
- `map<Status, string>`

In our legacy code map keys can be strings or enums. Any others were not supported. That would be it's own PR but doesn't the code here need to check the types of key and value?

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+                ValueSet::Union(vec![inner_set, ValueSet::Literal(Literal::Null)])
+            }
+            Ty::Named(name) => {
+                // For type aliases, keep the alias name (don't expand)
+                // The coverage check will handle expansion
+                ValueSet::OfType(name.clone())
+            }
+            Ty::Singleton(sv) => {
+                use crate::SingletonValue;
+                match sv {
```
antoniosarosi: Same comment here as before:

1. You can reduce nesting by importing `SingletonValue` above and doing 

```rust
Ty::Singleton(sv) => match sv {
    // ...
}
```

This is just styling, not as important.

2. Naming of `sv` isn't very clear, I'd call this `single_value` or `value` given it's inside `Ty::Singleton(value)`, that's readable enough.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+                    }
+                }
+                _ => {
+                    if !covered.contains(value_set) {
+                        covered.push(value_set.clone());
+                    }
+                }
+            }
+        }
+    }
```
antoniosarosi, line 706: > `// Replicate the logic from ExhaustivenessChecker::is_fully_covered`

This is pretty bad, Claude does this sometimes but duplicating two functions and testing the duplication doesn't really make sense 😭. Implementations can get out of sync and you have to write them twice.

Try to build a better testing framework for this, tell Claude how you'd want the test functions to look like and it'll come up with the API.

baml_language/crates/baml_thir/src/lib.rs
```diff
@@ -199,6 +201,62 @@ pub fn lower_project_class_fields(
...
+    let items = baml_hir::project_items(db, root);
+    let mut type_aliases = HashMap::new();
+
+    for item in items.items(db) {
+        if let baml_hir::ItemId::TypeAlias(alias_loc) = item {
+            let file = alias_loc.file(db);
+            let item_tree = baml_hir::file_item_tree(db, file);
+            let alias_data = &item_tree[alias_loc.id(db)];
+
+            let lowered_ty = lower_type_ref(db, &alias_data.type_ref);
```
antoniosarosi, line 253: `// TODO` for recursive aliases.

baml_language/crates/baml_tests/projects/match_exhaustiveness/match_exhaustiveness.baml
antoniosarosi, line 1: Could you add a couple examples here with nested unions to make sure they are covered?

```ts
function NestedTypeBindings(r: Result) -> int {
    match (r) {
        primitive: int | float | (string | bool) => 0,
        named: Result | (Failure | Success) => 1
    }
}
```

```ts
function NestedLiteralUnions(code: int) -> string {
    match (code) {
        200 | 201 | (202 | 204) => "success",
        400 | 404 | (405 | (406 | 409)) => "client error",
        _ => "other"
    }
}
```

baml_language/crates/baml_tests/snapshots/match_exhaustiveness/baml_tests__match_exhaustiveness__06_codegen.snap
antoniosarosi, line 1: Snapshot tests are good to verify changes between PRs, but there are a couple more interesting tests that you should add to verify codegen:

- [`baml_language/crates/baml_codegen/tests/`](https://github.com/BoundaryML/baml/tree/canary/baml_language/crates/baml_codegen/tests): Unit tests where you `assert` bytecode output given a Baml source against an expected list of bytecode instructions. Here you check whether the compiler outputs the exact instructions you want.

- [`baml_language/crates/baml_vm/tests/`](https://github.com/BoundaryML/baml/tree/canary/baml_language/crates/baml_vm/tests):  Unit tests where you write some Baml source code, the VM runs it and you assert the output against an expected result. You can use this to check the VM runs the bytecode you expect correctly. Can help identify bugs in the VM.

antoniosarosi (changes requested):
Nice work! All my review is in comments referencing source code. Overall great PR, needs some AI cleaning, additional tests, and other less important nits.

baml_language/crates/baml_hir/src/lib.rs
imalsogreg: +1 on Antonio's comment here. Could either do it now or mark it with a `// TODO` in the code so we don't lose track.

baml_language/crates/baml_codegen/src/lib.rs
```diff
@@ -115,12 +115,18 @@ pub fn compile_files(db: &dyn baml_thir::Db, files: &[SourceFile]) -> Program {
                 let body = function_body(db, *func_loc);
 
                 // Run type inference
+                // Note: type_aliases and enum_variants are not passed here,
+                // so exhaustiveness checking for type aliases and enums won't work.
+                // This is acceptable since codegen is for runtime execution,
+                // and type errors should be caught in the THIR.
```
imalsogreg: Could you add a `// TODO` here to note that we'll have to circle back to this? I think that if this function needs to run inference for any reason, and aliases & enums aren't available, then the inference done here won't be valid. Shouldn't block merge b/c this appears to be some kind of testing utility function.

(Unless I'm wrong and this instance of inference for some reason actually doesn't need that data).

baml_language/crates/baml_hir/src/type_ref.rs
```diff
@@ -36,11 +36,15 @@ pub enum TypeRef {
     },
     Union(Vec<TypeRef>),
 
-    /// Literal types in unions.
+    /// Literal types for exhaustiveness checking.
```
imalsogreg: (nit) this comment feels out of place because it discusses exhaustiveness checking in a context where we're just defining types.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+    /// - `"hello"` -> `Literal(String("hello"))`
+    /// - `true` -> `Literal(Bool(true))`
+    /// - `null` -> `Literal(Null)`
+    Literal(Literal),
+
+    /// Matches values in ANY of the sub-sets (union/disjunction).
+    ///
+    /// # Example
+    /// `200 | 201 | 204` -> `Union([Literal(200), Literal(201), Literal(204)])`
+    Union(Vec<ValueSet>),
```
imalsogreg, line 112: The fact that `OfType` exists _under_ `ValueSet` implies that a `ValueSet` may contain multiple different `OfType` variants with different types.

This might be fine, since the grammar rules out `1: int | false: bool =>`, so we'd just never construct such a mixed-type multi-pattern discriminant.

baml_language/crates/baml_thir/src/exhaustiveness.rs
```diff
@@ -0,0 +1,833 @@
...
+
+    /// Enum definitions: enum_name -> [variant_names]
+    enum_variants: &'a HashMap<Name, Vec<Name>>,
+
+    /// Type alias definitions: alias_name -> underlying_type
+    type_aliases: &'a HashMap<Name, Ty<'db>>,
+}
+
+/// Result of exhaustiveness checking.
+#[derive(Debug)]
```
imalsogreg, line 167: Ooh I like this. Clean :)

baml_language/crates/baml_thir/src/types.rs
rossirpaulo: yessir 🫡. any theory stuff or function docs, is non-ai and am planning on keeping it.

baml_language/crates/baml_thir/src/types.rs
rossirpaulo: The story is that while I was researching literal types, I mostly found books using this nomenclature and liked it enough to be consistent with the theory. That said, I'm happy to change or go in either direction.

baml_language/crates/baml_thir/src/exhaustiveness.rs
rossirpaulo: Yeah. The moment I read `clone` in rust's docs, i asked it to change 🤣

baml_language/crates/baml_codegen/src/lib.rs
rossirpaulo: Your concern is valid and well-taken. I see how if codegen needs inference results that depend on `type_aliases` or `enum_variants`, the inference done here would be incomplete. But if you trace through the code, the current assumption appears to hold:

1. Exhaustiveness checking (which needs these maps) produces errors like `NonExhaustiveMatch` and `UnreachableArm`. These are caught in the THIR/Diagnostics phase (`baml_onionskin/src/compiler.rs:661-669`), which *does* pass full context including `type_aliases` and `enum_variants`.
2. Core type inference (`infer_expr`, `infer_binary_op`, etc.) doesn't directly depend on these maps—they're stored in `TypeContext` primarily for the exhaustiveness checker.
3. Codegen only uses `inference.expr_types` and `inference.path_segment_types`, neither of which require resolving type aliases or enumerating enum variants.

That said, if someone later adds codegen logic that needs:
- Type alias resolution (e.g., `Result` → `Success | Failure`)
- Enum variant enumeration
- Exhaustiveness results for codegen decisions

...then this would silently break. A more robust solution would be to either pass full context or cache the THIR-phase inference result instead of re-running.

I'm writing a detailed reply since I'll be linking this comment in the codebase for someone to have full context of what's going on. This should give future maintainers enough context to make the right call.

baml_language/crates/baml_hir/src/lib.rs
rossirpaulo: Done! Added `TypeExpr::is_union()` and `TypeExpr::parts()` accessors to `ast.rs`, following the existing pattern. Consolidated the 3 duplicate implementations into a single canonical `TypeRef::from_ast()` that uses these accessors. The HIR lowering code now calls `TypeRef::from_ast(&t)` directly instead of manual CST walking.

Ah, this also fixed a bug where the old `signature.rs` implementation wasn't properly recognizing primitives inside type constructors (e.g., `int?` was becoming `Optional(Path("int"))` instead of `Optional(Int)`).

Great suggestion :)

baml_language/crates/baml_hir/src/lib.rs
rossirpaulo: oh! great catch! I've removed the dead  code.

baml_language/crates/baml_hir/src/lib.rs
rossirpaulo: I traced through this and confirmed the code is actually correct: the condition requires both starts_with('"') and ends_with('"'). For "literal"?, the second check fails (ends with ?), so we fall through to the optional check which strips ? and recurses.

That said, there were no tests for this case, so I've added unit tests for optional/array string literals plus a clarifying comment. Thanks for flagging it!

baml_language/crates/baml_hir/src/lib.rs
rossirpaulo: I've added a comment and TODO here so we can address this at a later time.
 
For anyone else in the team reading, find more context here: https://magnaplay.slack.com/archives/C0A3P816B0W/p1766183421615089

baml_language/crates/baml_hir/src/lib.rs
rossirpaulo: Same as above.

rossirpaulo: The TODO comment is still relevant here; we should still introduce some sort of helpful diagnostic in the future (or do some magic stuff under the hood). but, in the meantime, ive added an explicit `return TypeRef:Error` if we cannt parse as an `i64` or starts with text (float literal). at least we are not silently failing.

rossirpaulo: fyi: i'll not resolve these comments.

baml_language/crates/baml_hir/src/type_ref.rs
rossirpaulo: Fair. So updated.

baml_language/crates/baml_hir/src/type_ref.rs
rossirpaulo: Yes, my bad. These are relics from two different implementation attempts. So fixed.

baml_language/crates/baml_tests/projects/match_exhaustiveness/match_exhaustiveness.baml
rossirpaulo: Hmm. Per the BEP as written, we didn't cover the case of union type literals, and therefore that is not implemented, and I would expect these examples to fail. 

That said, I've just extended the parser to support these cases and added tests :)

antoniosarosi (approved):

PR #2868, opened 2025-12-23
v1: set up wasm bridge for new runtime

sxlijin:


baml_language/crates/baml_runtime/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+tokio = { workspace = true, optional = true }
+
+# HTTP client with WASM support
+reqwest = { workspace = true, optional = true }
+
+# Cancellation token
+tokio-util = { workspace = true, optional = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2"
```
aaronvg, line 37: it seems these all should be workspace deps as well no?

typescript2/app-vscode-webview/src/App.tsx
```diff
@@ -0,0 +1,19 @@
...
+  void import('jotai-devtools/styles.css');
+}
+
+const App: React.FC = () => (
+  <main className="app">
+    <header className="app__header">
+      <h1>Standalone Playground</h1>
+      <p>Shared UI components and state management come from the common package.</p>
+    </header>
+    <SplitPreview />
```
aaronvg, line 14: just a heads-up, in the new playground (post-refactor) we have an sdk.ts object that does all the initialization. It has an underlying wasm atom storage but it is semi-abstracted out. You'll want to use that pattern, so that we could in the future literally replace WASM with perhaps something else (e.g. the runtime API lives in a cloud service).

PR #2871, opened 2025-12-25
docs: clarify set_primary behavior has no fallback

### LINKED ISSUE #2827: clarify set_primary function
Context: 
- docs: https://docs.boundaryml.com/guide/baml-advanced/llm-client-registry
- related issue: https://github.com/BoundaryML/baml/issues/766

Problem: the "set_primary" terminology leaves me uncertain about what to expect in terms of behavior: "primary" suggests that there could be a silent fallback "secondary" of some kind; in my current use case I'd like to effectively do `set_only` or simply `set_client`.

Ayush2k02:
## Summary

Fixes #2827

This PR clarifies the documentation for the `set_primary` / `setPrimary` function to address user confusion about whether there's fallback behavior to a "secondary" client.

## Problem

The issue reporter noted that the term "primary" in `set_primary` was confusing because:
- "Primary" suggests there could be a silent fallback "secondary" client
- Users wanted clarity on whether it effectively acts as `set_only` or simply `set_client`
- The documentation didn't explicitly state whether there's any fallback behavior

After reviewing the implementation in `engine/baml-runtime/src/client_registry/mod.rs`, I confirmed that `set_primary` simply stores the client name with **no fallback logic**.

## Changes

Updated documentation in two files:

### 1. `fern/01-guide/05-baml-advanced/client-registry.mdx`
- Added a prominent `<Note>` block explaining there is **no fallback** to a secondary client
- Changed function description from "sets the client for the function to use" to "sets the **only** client that the function will use"
- Updated all code example comments (Python, TypeScript, Ruby, Go) to clarify "no fallback to other clients"
- Changed comments from "will now use MyAmazingClient as the calling client" to "will now use **only** MyAmazingClient"

### 2. `fern/03-reference/baml_client/client-option.mdx`
- Added clarification that the specified client will be "the **only** client used (no fallback to other clients)"

## Impact

This makes it crystal clear to users that:
- `set_primary` sets the **only** client that will be used
- If that client fails, the function call fails
- There is **no automatic retry** with a different client
- No silent fallback behavior exists

Users can now confidently use `set_primary` knowing exactly what to expect in terms of behavior.

### DIFF
--- a/fern/01-guide/05-baml-advanced/client-registry.mdx
+++ b/fern/01-guide/05-baml-advanced/client-registry.mdx
@@ -35,10 +35,10 @@ async def run():
         "api_key": os.environ.get('OPENAI_API_KEY')
     })
     
-    # Sets MyAmazingClient as the primary client
+    # Sets MyAmazingClient as the client to use (no fallback to other clients)
     cr.set_primary('MyAmazingClient')
 
-    # ExtractResume will now use MyAmazingClient as the calling client
+    # ExtractResume will now use only MyAmazingClient
     res = await b.ExtractResume("...", { "client_registry": cr })
 ```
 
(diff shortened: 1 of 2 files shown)


antoniosarosi:
Hi, I fixed this by adding a new section that explains how `set_primary` works in PR #2875. I would have merged this one but the way it's written is misleading because it implies you cannot use fallback clients with this option, which is not true.

And since you can add clients that use different strategies like `round-robin` to alternate between providers saying "this is the **only** client" also requires some nuance.

PR #2936, opened 2026-01-07
Better bidirectional typing

imalsogreg:
- Fix issue with typechecker where it wasn't conforming to the normal bidirectional typing recipe. Now we do more checking and less inference
- Add info_span to type errors, indicating where a type constraint came from under type mismatch
- Fix spans of function call arguments, to improve type errors with badly typed args
- Infer if exprs to plain types (not literal types)
- Respect literal types in type-annotated let expressions (check, don't infer, for annotated `let`s).
- Remove more panics from codegen (replace with error logging and default values)


baml_language/crates/baml_codegen/src/emit.rs
```diff
@@ -625,7 +628,10 @@ impl<'ctx, 'obj> StackifyCodegen<'ctx, 'obj> {
                             .class_object_indices
                             .get(class_name)
                             .copied()
-                            .unwrap_or_else(|| panic!("undefined class: {class_name}"));
+                            .unwrap_or_else(|| {
+                                log::error!("undefined class: {class_name}");
+                                return 0;
```
aaronvg: why wouldn't we panic here?

baml_language/crates/baml_codegen/src/emit.rs
```diff
@@ -645,15 +651,21 @@ impl<'ctx, 'obj> StackifyCodegen<'ctx, 'obj> {
                             .enum_object_indices
                             .get(enum_name)
                             .copied()
-                            .unwrap_or_else(|| panic!("undefined enum: {enum_name}"));
+                            .unwrap_or_else(|| {
+                                log::error!("undefined enum: {enum_name}");
+                                return Default::default();
+                            });
```
aaronvg: same, seems strange to not panic right? At this point the IR should be valid

baml_language/crates/baml_lsp_tests/test_files/syntax/expr/instanceof_narrowing.baml
```diff
@@ -43,4 +43,39 @@ function test_multiple_classes(x: Foo | Bar | Baz) -> string {
 
 //----
 //- diagnostics
-// <no-diagnostics-expected>
+// Error: Type `Foo | Bar` has no field `field`
+//     ╭─[ expr_instanceof_narrowing.baml:13:15 ]
+//     │
+//  13 │         return x.field;
+//     │               ────┬───  
+//     │                   ╰───── Type `Foo | Bar` has no field `field`
```
aaronvg: there should be no diagnostics here

baml_language/crates/baml_lsp_tests/test_files/syntax/expr/let_annotations.baml
```diff
@@ -25,20 +25,38 @@ function Ok() -> int {
 //----
 //- diagnostics
 // Error: Expected `int`, found `float`
+//    ╭─[ expr_let_annotations.baml:5:17 ]
```
aaronvg, line 28: seems like a redundant diagnostic got added here

baml_language/crates/baml_lsp_tests/test_files/syntax/invalid_assignment/arrays.baml
```diff
@@ -60,39 +60,39 @@ function ArrayAssignments() -> int {
...
 //     │
 //  34 │     h = "array";
 //     │        ────┬───  
-//     │            ╰───── Expected `int[]`, found `string`
+//     │            ╰───── Expected `int[]`, found `"array"`
 //     │ 
 //     │ Note: Error code: E0001
 // ────╯
-// Error: Expected `string`, found `int[]`
+// Error: Expected `"list"`, found `int[]`
```
aaronvg: this is also confusing -- it should expect a string still.

baml_language/crates/baml_lsp_tests/test_files/syntax/invalid_assignment/compound_ops.baml
```diff
@@ -35,48 +35,57 @@ function CompoundAssignments() -> int {
...
 //     │
 //  15 │     c += "five";
 //     │         ───┬───  
-//     │            ╰───── Expected `int`, found `string`
+//     │            ╰───── Expected `10`, found `"five"`
 //     │ 
 //     │ Note: Error code: E0001
 // ────╯
-// Error: Expected `string`, found `int`
+// Error: Expected `"start"`, found `100`
```
aaronvg: same comment as above

baml_language/crates/baml_codegen/src/emit.rs
imalsogreg: Onionskin tries to run all the stages, so if I'm trying to see something about an earlier stage (like how type errors look when a function is missing), I can't, because the later stage panics and kills the process.

baml_language/crates/baml_lsp_tests/test_files/syntax/expr/instanceof_narrowing.baml
imalsogreg: ty! taking a look.

imalsogreg: Fixed 👍

baml_language/crates/baml_lsp_tests/test_files/syntax/expr/let_annotations.baml
imalsogreg: Fixed

baml_language/crates/baml_lsp_tests/test_files/syntax/invalid_assignment/arrays.baml
imalsogreg: +1 fixed

baml_language/crates/baml_codegen/src/emit.rs
aaronvg: but wouldnt this mask errors in 'production'?

baml_language/crates/baml_mir/src/lower.rs
```diff
@@ -1678,21 +1678,22 @@ impl<'a, 'ctx> LoweringContext<'a, 'ctx> {
                 if let Some(&idx) = fields.get(&field.to_string()) {
                     return idx;
                 }
-                panic!(
+                log::error!(
```
aaronvg, line 1681: same it seems that with onionskin we would want a cfg that doesnt panic, but otherwise we would no?

imalsogreg: 👍  Discussed in person: let's restore the panic here, and change onionskin to skip possibly-panicking compiler phases when any diagnostic errors are present from earlier phases.

baml_language/crates/baml_codegen/src/emit.rs
antoniosarosi: If it hits that path it's a bug in the compiler, codegen should not panic. It means we parsed all class names and somehow we're getting a name that is not stored anywhere.

antoniosarosi: (`panic!` is ok if it happens we know there's a bug)

baml_language/crates/baml_lsp_tests/test_files/on_hover/unknown_variable.baml
```diff
@@ -13,12 +13,3 @@ function UseUnknown() -> int {
...
 // ───╯
-// Error: Cannot apply operator 'Add' to types `unknown` and `int`
-//    ╭─[ unknown_variable.baml:2:3 ]
-//    │
-//  2 │   unknown_var + 1
-//    │   ───────┬───────  
-//    │          ╰───────── Cannot apply operator 'Add' to types `unknown` and `int`
-//    │ 
-//    │ Note: Error code: E0004
-// ───╯
```
antoniosarosi, line 24: Isn't this a bug? The error should come from `unknown_var` not from the binary operator, but there should still be an error?

baml_language/crates/baml_mir/src/lower.rs
antoniosarosi: I don't think we should optimize our code for onionskin, it's a debugging tool for us mostly, if you open onionskin and codegen panics then you already have a repro and we can fix it.

baml_language/crates/baml_tests/snapshots/parser_stress/baml_tests__parser_stress__05_diagnostics.snap
```diff
@@ -13795,18 +13795,6 @@ expression: output
...
-    │
-  8 │ ╭─▶   while \ {
-    ┆ ┆   
- 10 │ ├─▶   }
-    │ │         
-    │ ╰───────── Expected `bool`, found `void`
-    │     
-    │     Note: Error code: E0001
-────╯
-
```
antoniosarosi, line 13809: Not really related to this PR but for this code:

```ts
function BodyTorture() -> int {
  let x = ¿;
  
  while \ {
      x += 1;
  }
}
```

The error we get is:

```
  [parse] Error: Expected expression, found error
   ╭─[ very_invalid.baml:6:11 ]
   │
 6 │   let x = ¿;
   │           ─┬  
   │            ╰── Expected expression, found error
   │ 
   │ Note: Error code: E0010
───╯

  [parse] Error: Expected expression, found error
   ╭─[ very_invalid.baml:8:10 ]
   │
 8 │   while \ {
   │          ┬  
   │          ╰── Expected expression, found error
   │ 
   │ Note: Error code: E0010
───╯
```

Maybe we should carry more data in the error variant so we can say something like:

```
 8 │   while \ {
   │          ┬  
   │          ╰── Expected expression, found unexpected token "\"
```

baml_language/crates/baml_lsp_tests/test_files/on_hover/unknown_variable.baml
antoniosarosi: Nevermind, just realized there were two errors, "unknown variable error" is still there 👍

PR #2993, opened 2026-01-17
fix(ts): add BamlError base class and improve error hierarchy

antoniosarosi:
## Summary

This PR improves the TypeScript error handling by:

1. **Adding `BamlError` as base class** for all BAML errors
2. **Adding `BamlClientError`** as intermediate class for client-related errors  
3. **Fixing `isBamlError()` and `toBamlError()`** type signatures
4. **Adding `raw_response` property** to `BamlClientHttpError` for structured access to the raw LLM API response
5. **Fixing error classification bug** where HTTP errors were incorrectly thrown as `BamlValidationError`

**Error hierarchy:**
```
Error
└── BamlError
    ├── BamlClientFinishReasonError
    ├── BamlValidationError
    ├── BamlAbortError
    └── BamlClientError
        └── BamlClientHttpError (now includes raw_response field)
            └── BamlTimeoutError
```

### `raw_response` on `BamlClientHttpError`
Users can now access the raw LLM API response separately from the error message:
```typescript
catch (error) {
  if (error instanceof BamlClientHttpError) {
    console.log(error.status_code);     // e.g., 401
    console.log(error.raw_response);    // e.g., '{"error":{"message":"Invalid API key",...}}'
    const parsed = JSON.parse(error.raw_response);
    // Access structured error info from the LLM provider
  }
}
```

### Error Classification Fix
Previously, HTTP errors (400, 401, 404, etc.) from LLM providers could be incorrectly thrown as `BamlValidationError`. Now:
- `BamlValidationError` → Only for parsing failures on **successful** LLM responses
- `BamlClientHttpError` → For all HTTP/client errors (with `status_code` and `raw_response`)
- `TimeoutError` → For timeout errors
- `AbortError` → For cancelled operations

## Context

Bug report from Discord: https://discord.com/channels/1119368998161752075/1461282398493741234

User issues addressed:
- Error hierarchy not matching documentation (all errors extended `Error` instead of `BamlError`)
- No way to get raw LLM response separately from error message
- HTTP errors incorrectly classified as `BamlValidationError`

## Test plan

- [x] Verify TypeScript compilation succeeds
- [x] Verify `instanceof BamlError` catches all BAML error types
- [x] Verify existing `instanceof` checks for specific error types still work
- [x] Verify `raw_response` is populated with actual LLM API response
- [x] Verify HTTP errors are correctly classified as `BamlClientHttpError`
- [x] All error-handling tests pass


engine/language_client_typescript/typescript_src/errors.ts
```diff
@@ -295,7 +317,13 @@ function createBamlErrorUnsafe(error: unknown): BamlErrors | Error {
   return error;
 }
 
-export function isBamlError(error: unknown): error is BamlErrors {
+export function isBamlError(error: unknown): error is BamlError {
```
aboqasem, line 349: This asserts `error is BamlError`, but it might not be, because it's checking on properties as well.

Is there a possibility that BAML gives the user access to this kind of error that is NOT a `BamlError` ? Shouldn't it be fixed to convert it into a `BamlError`?

aboqasem:
Also, I think there's another bug other than the `BamlError` issue: https://discord.com/channels/1119368998161752075/1461282398493741234/1461587885164855518

antoniosarosi:
@aboqasem ok I added more changes, there's `.raw_response` now on `BamlClientHttpError` to access response separately from message, and there's also the `BamlValidation` change to throw `BamlClientHttpError` if it's an HTTP error.

engine/language_client_typescript/typescript_src/errors.ts
```diff
@@ -295,45 +327,47 @@ function createBamlErrorUnsafe(error: unknown): BamlErrors | Error {
...
+      // Only return if we successfully converted to a BamlError
+      if (converted instanceof BamlError) {
+        return converted;
+      }
+    }
+
+    // Return original error if not convertible
+    return error;
+  } catch {
+    return error;
```
aboqasem: I think this would make more sense if it was `toBamlError(error: unknown): BamlError | null` and only return `BamlError` if it can:

```suggestion
    return null;
  } catch {
    return null;
```

But at least the issue is fixed, thanks!

aboqasem (approved):

antoniosarosi:
@aboqasem Done, made it `BamlError | null`

PR #3004, opened 2026-01-20
feat(bex): add Union variant to BexExternalValue for declared type info

antoniosarosi:
## Summary

- **Add `BexExternalValue::Union` variant** with union metadata for FFI consumers
- **Add type info to `Array` and `Map` variants** - `element_type` for arrays, `key_type`/`value_type` for maps
- **Fix heterogeneous array compilation** - arrays like `[1, "two", 3]` were losing integer elements
- **Fix union type display** in HIR snapshots - `(int | string)[]` was incorrectly displayed as `int | string[]`
- **Merge canary** - includes token-based type parsing from PR #3006

### BexExternalValue Changes

**Union variant** - When the declared return type is a union or optional, the value is wrapped with metadata:

```rust
BexExternalValue::Union {
    value: Box<BexExternalValue>,  // The actual value
    metadata: UnionMetadata,        // Type info
}
```

**Array variant** - Now includes element type for proper FFI serialization:

```rust
BexExternalValue::Array {
    element_type: Ty,              // e.g., Ty::Union([Int, String]) for (int | string)[]
    items: Vec<BexExternalValue>,
}
```

**Map variant** - Now includes key and value types:

```rust
BexExternalValue::Map {
    key_type: Ty,                  // Usually Ty::String
    value_type: Ty,                // e.g., Ty::Union([Int, String]) for map<string, int | string>
    entries: IndexMap<String, BexExternalValue>,
}
```

This mirrors the CFFI protocol (`CFFIValueList.item_type`, `CFFIValueMap.key_type/value_type`, `CFFIValueUnionVariant`).

### UnionMetadata

```rust
pub struct UnionMetadata {
    pub name: Option<String>,      // For named type aliases
    pub is_optional: bool,         // Whether this is T?
    pub is_single_pattern: bool,   // Only one non-null option
    pub union_type: Ty,            // Full declared union type
    pub selected_option: Ty,       // Which member was selected
}
```

### Bug Fix: Heterogeneous Arrays

Arrays with mixed types like `[1, "two", 3]` with type `(int | string)[]` were only capturing string elements.

**Root cause**: In `lower_array_literal`, STRING_LITERAL is a node (has quote children) while INTEGER_LITERAL is a bare token. The code processed nodes first, then only processed tokens if the list was empty.

**Fix**: Process both nodes and tokens in a single iteration using `children_with_tokens()`.

### PR Review Fixes

- **O(n²) → O(n) field lookup**: Build HashMap for field types instead of iterating for each field
- **Return `EngineError` instead of silent fallback**: `find_matching_member` now returns `Result<Ty, EngineError>` instead of silently falling back to first member
- **File renames**: `snapshot.rs` → `bex_external_value.rs`, `external_value.rs` → `bex_value.rs`
- **MIR display**: LLM functions now show "(LLM function - no MIR)" instead of misleading "no MIR due to errors"

---

## Test plan

- [x] All 14 `typed_values.rs` tests pass (union wrapping, array/map type info)
- [x] All bex_engine tests pass (gc.rs, concurrent.rs)
- [x] Full test suite passes (`cargo test -p baml_tests`)
- [x] Snapshot tests updated


baml_language/crates/bex_engine/src/lib.rs
```diff
@@ -286,120 +292,266 @@ impl BexEngine {
...
+                                .map(|f| &f.field_type)
+                                .ok_or_else(|| EngineError::SchemaInconsistency {
+                                    message: format!(
+                                        "Field '{name}' not found in class '{class_name}'"
+                                    ),
+                                })?;
+
+                            Ok((name.clone(), self.vm_value_to_external(value, field_type)?))
+                        })
+                        .collect();
```
rossirpaulo, line 565: For a class with n fields, this is O(n^2) total. For most classes it's fine, but if you had a class with 50+ fields called in a hot loop, you could build a lookup map once:

```rs
let field_types: HashMap<&str, &Ty> = class_def
      .fields
      .iter()
      .map(|f| (f.name.as_str(), &f.field_type))
      .collect();
```

which is O(1). 

That said, this is probably premature optimization, but I'd advocate thinking through that lens if we can!

baml_language/crates/bex_engine/src/lib.rs
```diff
@@ -286,120 +292,266 @@ impl BexEngine {
...
+
+    /// Find which union member matches a value.
+    fn find_matching_member(value: &BexExternalValue, members: &[Ty]) -> Ty {
+        for member in members {
+            if Self::value_matches_type(value, member) {
+                return member.clone();
+            }
+        }
+        // Fallback: shouldn't happen if type checking is correct
+        members.first().cloned().unwrap_or(Ty::Null)
```
rossirpaulo: Should we `Return<Ty, EngineError>` and propagate the error or just simply `panic` in this case?

rossirpaulo: tldr: this silently fails. might be good to defend against that

baml_language/crates/bex_external_types/src/bex_external_value.rs
```diff
@@ -1,42 +1,99 @@
...
+//!     BexExternalValue::Union { value, metadata } => {
+//!         println!("Selected: {}", metadata.selected_option);
+//!         println!("Could have been: {:?}", metadata.union_type);
+//!     }
+//!     _ => {}
+//! }
+//! ```
+
+// Re-export Ty from baml_snapshot for convenience
+pub use baml_snapshot::Ty;
```
rossirpaulo, line 26: We are re-exporting `baml_snapshot::Ty` from `bex_external_types`. This creates a dependency direction (external types -> snapshot) that might be backwards. Should `Ty` live in `bex_external_types instead`?

I was trying to understand how all of these files relate to each other and the dependency graph made me question this. Unsure though.

baml_language/crates/bex_engine/src/lib.rs
antoniosarosi: `EngineError`

baml_language/crates/bex_engine/src/lib.rs
antoniosarosi: Nah it's not premature, good catch.

baml_language/crates/bex_external_types/src/bex_external_value.rs
antoniosarosi: I think it should, also `snapshot.rs` needs a different name.

baml_language/crates/baml_compiler_hir/src/type_ref.rs
```diff
@@ -140,6 +142,22 @@ impl TypeRef {
             return TypeRef::Optional(Box::new(inner));
         }
 
+        // Check for parenthesized types (e.g., "(int | string)")
+        if text.starts_with('(') && text.ends_with(')') {
```
hellovai: lel i love this hack

baml_language/crates/bex_external_types/src/bex_value.rs
```diff
@@ -33,127 +33,127 @@ use crate::{Handle, Snapshot};
...
+impl Default for BexValue {
     fn default() -> Self {
-        ExternalValue::Snapshot(Snapshot::Null)
+        BexValue::External(BexExternalValue::Null)
     }
 }
 
-impl ExternalValue {
+impl BexValue {
     /// Check if this value is null.
```
hellovai, line 54: I think we can kill all these `.as_***` methods

baml_language/crates/bex_external_types/src/bex_external_value.rs
hellovai: i think i agree with @rossirpaulo we should just directly define the `Ty` in `bex_external_types`

and then `baml_snapshot` can import from it. but i'm ok to do this in a later world

baml_language/crates/bex_external_types/src/bex_external_value.rs
```diff
@@ -1,42 +1,99 @@
...
+
+    /// Which option of the union was selected (e.g., `Ty::Int`, `Ty::String`, `Ty::Class("Success")`).
+    pub selected_option: Ty,
+}
+
+impl UnionMetadata {
+    /// Create metadata for a union type.
+    pub fn new(union_type: Ty, selected_option: Ty) -> Self {
+        let (is_optional, is_single_pattern) = match &union_type {
+            Ty::Union(members) => {
```
hellovai, line 57: future work, but i think our `Ty` should not have both unions + Optional, just one or the other.

baml_language/crates/bex_external_types/src/bex_external_value.rs
```diff
@@ -53,167 +110,50 @@ pub enum Snapshot {
     /// Owned string.
     String(String),
 
-    /// Owned array of snapshots.
-    Array(Vec<Snapshot>),
+    /// Owned array of values.
+    Array(Vec<BexExternalValue>),
```
hellovai: container types need `Ty` as well to construct them in the FFI layer.

baml_language/crates/baml_compiler_hir/src/type_ref.rs
antoniosarosi: Properly solved in #3006

PR #3115, opened 2026-02-13
fixed let annotations

rossirpaulo:
**Summary**
This PR fixes diagnostics for annotated `let` assignments so type mismatches include both:
- the existing primary span on the mismatched RHS expression, and
- a secondary info span pointing to the type annotation that introduced the requirement.

It keeps the existing diagnostic code/message/phase behavior (`E0001`, `Expected ..., found ...`, type phase), and only improves span context.

**Problem**
For `let` annotations like `let y: int = 10.0;`, the typechecker emitted only one span (on `10.0`).  
It should also show where the expected type came from (the `int` annotation).

**What Changed**
- **Type-checking flow (`baml_compiler_tir`)**
  - Added a scoped `check_expr_with_info_location(...)` path for passing optional constraint-source location.
  - Annotated `let` initializers now pass the annotation span as `info_location`.
  - Mismatch emission now carries that `info_location` through to diagnostics.
- **Diagnostic conversion (`baml_compiler_diagnostics`)**
  - `TypeMismatch.info_location` now maps to a **secondary annotation** (`with_secondary`) instead of related-info metadata.
  - Secondary label text: `Type required here`.
- **Tests**
  - Added unit coverage for `TypeMismatch` with `info_location` in `to_diagnostic`.
  - Updated `let_annotations.baml` expected diagnostics output to include secondary spans.
- **Lint/compliance cleanup**
  - Adjusted `info_location` passing to satisfy strict clippy (`Option<&ErrorLocation>` + `.cloned()`), no behavior change.

**Before / After**
- **Before**: one span per mismatch (RHS only).
- **After**: same primary RHS span + secondary span on annotation token.
  - `let y : int = 10.0;` -> secondary on `int` (line 5, cols 11–13)
  - `let c: int = "hello";` -> secondary on `int` (line 6, cols 10–12)

## Release Notes

* **Bug Fixes**
  * Enhanced error messages for type mismatches to display secondary annotations indicating where the expected type constraint originates, providing clearer guidance for resolving type-related issues in your code.

### DIFF
--- a/baml_language/crates/baml_compiler_diagnostics/src/to_diagnostic.rs
+++ b/baml_language/crates/baml_compiler_diagnostics/src/to_diagnostic.rs
@@ -71,7 +71,7 @@ impl<C: ErrorContext> TypeError<C> {
                 )
                 .with_primary_span(loc_fn(location));
                 if let Some(info_location) = info_location {
-                    diag.with_related(loc_fn(info_location), "Type defined here")
+                    diag.with_secondary(loc_fn(info_location), "Type required here")
                 } else {
                     diag
                 }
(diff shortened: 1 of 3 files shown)


baml_language/crates/baml_compiler_diagnostics/src/to_diagnostic.rs
```diff
@@ -71,7 +71,7 @@ impl<C: ErrorContext> TypeError<C> {
                 )
                 .with_primary_span(loc_fn(location));
                 if let Some(info_location) = info_location {
-                    diag.with_related(loc_fn(info_location), "Type defined here")
```
rossirpaulo, line 74: @imalsogreg does this seem sound to you?

PR #3137, opened 2026-02-19
Fix off-by-one in streaming parser should_close_unescaped_string for non-array contexts

davidsyoung:
## Summary

Fixes an off-by-one error in `should_close_unescaped_string` that causes field name/value corruption when streaming chunks split mid-token inside JSON objects. The `InArray` branch correctly increments `counter` before returning on iterator exhaustion, but the `InNothing`, `InObjectKey`, and `InObjectValue` branches do not — causing the outer parsing loop to re-process the last consumed character.

## Background

We've had a recurring but intermittent failure in production for several months where BAML's streaming parser corrupts field names in parsed output. For example, `therapeutic_approach` becomes `therapeutic_apeutic_approach`. The corruption was rare enough (~1-2% of calls) and non-deterministic enough that it took significant effort to isolate — it only occurs when a streaming chunk boundary falls inside an unquoted string within an object context, and the specific split produces visible duplication rather than a harmless repeated whitespace character.

## Root cause

In `json_parse_state.rs`, `should_close_unescaped_string` is called when the parser encounters an unquoted string. It consumes characters from a peekable iterator looking for a structural delimiter (`:`, `,`, `}`, `]`). When the iterator is exhausted (stream incomplete), it returns `CloseStringResult::Close(counter, Incomplete)` where `counter` tells the outer loop how many characters to skip.

The `InArray` branch (line 370) correctly does:
```rust
counter += 1; // Indicate that we called next() one time after the final `Some`.
CloseStringResult::Close(counter, CompletionState::Incomplete)
```

But `InNothing` (line 225), `InObjectKey` (line 240), and `InObjectValue` (line 355) omit the `+= 1`:
```rust
// missing counter += 1
CloseStringResult::Close(counter, CompletionState::Incomplete)
```

The caller in `fixing_parser.rs` uses this value to advance the outer iterator:
```rust
Ok(increments) => {
    for _ in 0..increments {
        chars.next(); // skip already-consumed characters
    }
}
```

Without the `+= 1`, the last character consumed inside the function gets re-processed by the outer loop, producing character duplication at chunk boundaries.

## When it triggers

The bug requires all of these conditions simultaneously:

1. **Streaming mode** — only affects incremental/partial parses, not complete JSON
2. **Unquoted string context** — the parser is processing text that isn't inside quotes
3. **Object context** — the unquoted string is inside an object key, object value, or at top level (not an array — that branch was already correct)
4. **Chunk boundary splits mid-token** — the streaming chunk ends before a structural delimiter is found, causing the iterator to exhaust

This makes it a narrow edge case that depends on how the LLM provider chunks its streaming response — which varies per request, making it non-deterministic and difficult to reproduce.

## Example BAML that triggers it

Any BAML class with multiple string fields in an object can trigger it when streamed. In our case:

```baml
class ResponseParagraphWithCitations {
    paragraph string @description("...")
}

class ResponseWithInlineCitations {
    situation_analysis string @description("...")
    relevant_context string @description("...")
    therapeutic_approach string @description("...")
    initial_response_plan string @description("...")
    plan_critique string @description("...")
    reflection string @description("...")
    final_response_plan string @description("...")
    response_in_paragraphs ResponseParagraphWithCitations[] @description("...")
}

function GenerateResponse(context: string) -> ResponseWithInlineCitations {
    client KimiK25Direct {
        provider openai-generic
        options {
            base_url "https://api.moonshot.ai/v1"
            model "kimi-k2.5"
        }
    }
    prompt #"
        Analyze the situation and respond.
        {{ ctx.output_format }}
    "#
}
```

When the LLM streams back the JSON and a chunk boundary falls inside the value of `therapeutic_approach` (or any other field), the parser re-processes the last character of the chunk, producing corrupted output like `therapeutic_apeutic_approach`.

## Fix

Three lines — add `counter += 1` to the `InNothing`, `InObjectKey`, and `InObjectValue` branches to match the existing `InArray` behavior.

## Summary by CodeRabbit

* **New Features**
  * Added incremental streaming parser state to assemble partial JSON-ish values across input boundaries.

* **Bug Fixes**
  * Fixed off-by-one and duplication issues when streams end mid-token; improved handling of quoted/unquoted values, object keys, arrays and lookahead at boundaries.

* **Tests**
  * Added unit tests covering partial-token and end-of-stream boundary cases to prevent regressions.

### DIFF
--- a/engine/baml-lib/jsonish/src/jsonish/parser/fixing_parser.rs
+++ b/engine/baml-lib/jsonish/src/jsonish/parser/fixing_parser.rs
@@ -171,43 +171,24 @@ mod tests {
         }
     }
 
-    // Tests for the off-by-one fix in should_close_unescaped_string.
+    // Regression tests for the off-by-one fix in should_close_unescaped_string.
     // When the iterator exhausts without finding a structural delimiter, the
     // counter must account for the last consumed character to prevent the outer
-    // loop from re-processing it (which causes character duplication).
-
-    #[test]
-    fn test_partial_unquoted_value_no_char_duplication() {
-        // InObjectValue: unquoted string "hello" as a value, stream ends mid-token.
-        // Without the fix, the last char would be duplicated (e.g. "helloo").
-        let opts = ParseOptions::default();
-        let vals = parse(r#"{"key": hello"#, &opts).unwrap();
-        match &vals[0].0 {
-            Value::Object(fields, _) => {
-                assert_eq!(fields.len(), 1);
-                let (key, val) = &fields[0];
-                assert_eq!(key.as_str(), "key");
-                match val {
-                    Value::String(s, cmplt) => {
-                        assert_eq!(s.as_str(), "hello", "Value should be 'hello' without duplicated chars");
-                        assert_eq!(cmplt, &CompletionState::Incomplete);
-                    }
-                    _ => panic!("Expected string value, got: {val:?}"),
-                }
-            }
-            _ => panic!("Expected object"),
-        }
-    }
+    // loop from re-processing it. These tests fail without the fix.
+    //
+    // Note: The InObjectValue branch has the same bug but it only manifests
+    // during streaming (multiple parse() calls on successive chunks), not in
+    // single-pass parsing, so it cannot be tested at this level.
 
     #[test]
     fn test_partial_unquoted_key_no_char_duplication() {
-        // InObjectKey: unquoted key, stream ends before ':' is found.
-        // Without the fix, the last char of the key would be duplicated.
+        // InObjectKey: unquoted key "mykey", stream ends before ':' is found.
+        // Without the fix, the off-by-one causes the last char to be
+        // re-processed, creating a spurious key-value pair.
         let opts = ParseOptions::default();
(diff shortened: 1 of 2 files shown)


davidsyoung:
Added docstrings and 2 regression tests. Both tests pass with the fix and fail without it:

- `test_partial_unquoted_key_no_char_duplication` — exercises the `InObjectKey` branch. Without the fix, the off-by-one re-processes the last character, creating a spurious key-value pair.
- `test_partial_unquoted_toplevel_no_char_duplication` — exercises the `InNothing` branch. Without the fix, parsing fails entirely ("No JSON objects found").

Note: the `InObjectValue` branch has the same off-by-one bug but it only manifests during **streaming** (multiple `parse()` calls on successive chunks). In a single-pass `parse()` call, the re-processed character gets absorbed harmlessly. This matches our production experience — the corruption only appeared when LLM providers chunked responses at specific boundaries. The `InObjectValue` branch would need a streaming-level integration test to cover, which is beyond the scope of this fix.

aaronvg:
nice deep dive! We'll get this merged for the release next week likely

2kai2kai2:
Merged here with formatting fixes https://github.com/BoundaryML/baml/pull/3202

PR #3173, opened 2026-02-24
baml_language: route vm function execution to `BexEngine`

miguelcsx:
### Why this change

`BexVm::from_program` let callers run the VM without the engine, but the engine is where execution correctness lives (event loop, notify/span-notify handling, GC, and consistent result/error propagation). In practice, bypassing it meant re-implementing that logic, and getting edge cases wrong.

`tools_onionskin` was the only real non-test user, and its custom loop had a user-visible bug: `SpanNotify` could be dropped, leaving the final result unset (blank Runner panel). It also treated `Notify` as an error instead of a normal mid-execution event.

### What this PR does

* Removes `from_program` to prevent bypassing the engine.
* Switches `tools_onionskin` to run via `BexEngine` using `SysOps::all_unsupported()` (safe no-op SysOps), so all intermediate events are handled correctly and results are always set.
* Adds `make_vm()` in `baml_tests` for VM-level tests/benchmarks that need direct VM control; everywhere else should use `BexEngine`.

## Summary by CodeRabbit

* **Refactor**
  * Switched the tool’s execution backend from a VM-based runner to an engine-driven runner for more reliable execution and UI sync.
  * Updated value formatting to reflect engine results, improving display for arrays, variants and union cases.
  * Improved runtime error and arity reporting when running functions.

* **Chores**
  * Updated crate dependencies to support the new execution path (including async runtime).

* **Tests**
  * Added formatter tests for scalar, string, array and variant representations.

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -682,6 +682,7 @@ dependencies = [
  "baml_project",
  "baml_type",
  "baml_workspace",
+ "bex_heap",
  "bex_vm",
  "bex_vm_types",
  "codspeed-divan-compat",
@@ -4865,6 +4866,7 @@ dependencies = [
  "baml_project",
  "baml_type",
  "baml_workspace",
+ "bex_engine",
  "bex_vm",
  "bex_vm_types",
  "clap",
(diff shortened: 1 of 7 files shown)


sxlijin:
Hi Miguel! Thanks for this PR, sorry we weren't able to get this merged back when you were contributing!

This PR is pretty out-of-date now, so I'm going to close this out. Let us know if you'd like to get involved again!

PR #3176, opened 2026-02-25
Add inlay hints for let statements and function call parameters to LSP

codeshaunted:
Adds inlay hints for let statements (including those in both C-style and enhanced for loops) and function call parameters to the LSP. Type hints have goto-def support, including for individuals in compound types. Type hints for let statements not generated by the compiler (ex: those in a for-in loop) can also be inserted with text edits.

## Summary by CodeRabbit

* **New Features**
  * Comprehensive inlay hints: parameter-name and inferred-type hints with navigation targets and optional quick-edit edits; LSP now advertises and serves inlay-hint requests.

* **Bug Fixes**
  * goto-definition works for symbols outside functions.
  * Improved type inference caching for more accurate hints and displays.
  * Compiler-generated bindings are marked to suppress inappropriate type hints.

* **Tests**
  * Added end-to-end inlay-hints fixtures, parsing, runner, and updater support.


baml_language/crates/baml_compiler_hir/src/body.rs
```diff
@@ -521,6 +522,18 @@ pub enum LoopOrigin {
     ForLoop,
 }
 
+/// Indicates where a let statement originated from.
+///
+/// This is used primarily for inlay hints to enable the LSP to determine if
+/// the let statement came from the source or was generated by the compiler.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LetOrigin {
```
imalsogreg, line 530: This could be generalized to just `Origin`, still used in `Stmt::Let` and simultaneously used by @sxlijin for the compiler generated streaming types?

baml_language/crates/baml_lsp_actions/src/inlay_hints.rs
```diff
@@ -0,0 +1,385 @@
...
+    }]
+}
+
+/// Convert a [`Ty`] into label parts, wrapping in parentheses if it's a
+/// compound type (union or function) that would be ambiguous without them.
+fn wrap_if_compound(db: &ProjectDatabase, ty: &Ty) -> Vec<InlayHintLabelPart> {
+    if matches!(ty, Ty::Union(_) | Ty::Function { .. }) {
+        let mut parts = vec![InlayHintLabelPart {
+            value: "(".into(),
+            target: None,
```
imalsogreg, line 103: Maybe a TODO: it looks like clicks on the parts of a compound type won't goto-def because of this default `target: None`?

baml_language/crates/baml_lsp_actions/src/inlay_hints.rs
```diff
@@ -0,0 +1,385 @@
...
+            });
+        }
+    }
+}
+
+/// Compute all inlay hints for the given file.
+///
+/// To add new hint categories, implement [`HintCollector`] and add to `collectors`.
+pub fn inlay_hints(db: &ProjectDatabase, file: SourceFile, _project: Project) -> Vec<InlayHint> {
+    let collectors: &[&dyn HintCollector] = &[&CallArgNames, &LetTypeAnnotations];
```
imalsogreg: This feels more complicated than it needs to be. (Personally I'd drop the HintCollector trait and just write two functions with different names). But it's a nit. Clearly the approach here works out!

imalsogreg (approved):
(we don't really do code review with approve/reject, but) just hitting Approve to communicate that I looked through the PR. looks awesome! Testing it out right now. One thing I couldn't tell from the source is whether it continues to work when the code is broken.

PR #3185, opened 2026-02-26
fix(go-cffi): fix deadlock and panics under high concurrency

invakid404:
## Summary

Fixes three concurrency bugs in the Go CFFI callback layer that cause deadlocks and process crashes under sustained concurrent load (~100 streaming/call requests in parallel). Discovered during load testing.

All changes are in `engine/language_client_go/pkg/` (Go only — no Rust changes).

### 1. Replace random callback IDs with atomic counter (`callbacks.go`)

`create_unique_id` uses `rand.Intn(1000000)` in a retry loop while holding an exclusive lock. When callback entries leak (see bug #2 below), the ID space fills up, the loop spins longer, and all goroutines needing `callbackMutex.RLock()` starve — full deadlock.

Fix: `sync/atomic.Uint32` counter. IDs are unique by construction, no retry loop, lock acquired only after ID is chosen.

### 2. Prevent panics from concurrent callback channel operations (`callbacks.go`, `runtime.go`)

`error_callback` and `trigger_callback` can race on the same callback ID — one closes the channel while the other sends. `send on closed channel` panics crash the entire plugin process and all in-flight requests.

Go has no atomic check-and-send for channels (`select { case ch <- v: default: }` still panics on closed channels). Fix: `safeSend()`/`safeClose()` helpers that wrap operations with `defer`/`recover()`. Zero overhead in the non-panic path.

Also makes callback channels buffered (64 for stream, 1 for object-handle) so that post-cancellation sends don't block forever, which was causing the callback entry leaks that feed bug #1.

### 3. Recover from panics in `on_tick_callback` (`callbacks.go`)

`on_tick_callback` invokes user-provided handlers that send on application-level output channels. When the stream consumer closes its channel (context cancelled, iteration done), a concurrent tick panics. This is a different code path from #2 — it's the user's channel, not the internal callback channel.

Fix: `defer func() { recover() }()` at the top of `on_tick_callback`.

## Testing

Each fix was validated independently using a "leave one out" methodology against a 100k-request stress test (40k call + 20k stream + 40k parse, concurrency 100, mock LLM):

| Fix removed | Result |
|---|---|
| Atomic IDs | Deadlock at ~23k requests |
| safeSend/safeClose | 77% success, repeated panics |
| on_tick recover | 99.9% success, 22 failures |
| All fixes applied | **100% success, 1500+ req/s** |

## Summary by CodeRabbit

* **Bug Fixes**
  * Replaced non-deterministic callback IDs with deterministic sequencing to improve reliability.
  * Introduced buffered callback channels and safer send/close handling to prevent panics, races, and deadlocks.
  * Added panic recovery and safer cleanup for error paths to ensure callbacks are closed reliably and reduce crashes.


invakid404:
Addressed in 6542e63:

**Logging on_tick panics** — done; panics are now logged via `callbackLog` so they're visible when `BAML_FFI_CLIENT_LOG` is set.

**Missing `safeClose` in `CallFunctionStream` and `CallFunctionParse`** — good catches, fixed. These were pre-existing bugs with the same pattern we already fixed in `CallFunction` and `BuildRequest`.

**`atomic.Uint32` wrap-around** — the C FFI boundary uses `uint32_t`, so switching to Uint64 would require changes on both sides. At 1500 req/s continuous throughput, wrap-around takes ~33 days, and a collision would require the wrapped ID to land on one of the handful of currently in-flight callbacks. Not worth the churn.

invakid404:
Addressed in 605b936:

**Log recovered panics in safeSend/safeClose** — done, now consistent with on_tick_callback.

**Callback entry leak on synchronous FFI failure** — good catch. Added `cleanupCallback(id, ch)` that does `safeClose` + `deleteCallback` under the mutex. All four error paths in runtime.go now use it instead of bare `safeClose`.

aaronvg:
wow really nice work, we'll work on merging this in tomorrow

PR #3195, opened 2026-03-01
bridge(feat): track created handles and enable batch cleanup

miguelcsx:
This PR makes opaque-handle cleanup predictable across runtimes (C/Python/WASM).

When we serialize an outbound value, we now also return `handles_created`: the list of handle keys allocated during encoding. Hosts can reliably free any handles they don’t retain, preventing slow leaks in long-running processes.

Cleanup is centralized with `HandleTable::release_many` plus shared binding helpers, replacing duplicated per-runtime logic and reducing lock contention and extra cloning.

We also harden the FFI boundary by validating handle types returned by hosts (while staying forward-compatible with unknown enum values), and we avoid unexpectedly huge payloads by only inlining large media under a configurable size limit.

Migration: `BamlOutboundValue` gains `handles_created`, so consumers need to re-generate protobuf bindings to access the new field; host code should call the batch-release helper (client libs include helpers to compute/release unused handles).

## Release Notes

* **New Features**
  * Added bulk handle release capability across all language bindings.
  * Implemented automatic cleanup of unused handles after operations.
  * Added handle type validation to ensure type correctness.
  * Introduced configurable size limits for inlined media content.


sxlijin:
Hi Miguel! Thanks for this PR, sorry we weren't able to get this merged back when you were contributing!

This PR is pretty out-of-date now, so I'm going to close this out. Let us know if you'd like to get involved again!

PR #3366, opened 2026-04-14
fix(lsp): set will_save to false in advertised sync options

MFS-code:
## Problem

The Bex language server advertised `textDocumentSync.willSave: true`, which tells LSP clients to send `textDocument/willSave` before every save even though Bex performs no pre-save work.

Bex accepts `willSave` as a no-op, so advertising the notification remains unnecessary.

## Fix

- Advertise `willSave: false` from Bex.
- Keep Bex tolerant of unsolicited `willSave` notifications, but document that they are not advertised and perform no work.
- Add a regression test for the Bex capability builder.

`didSave`, full-document synchronization, and the other advertised capabilities are unchanged. The legacy `engine/` language server is unchanged.

## Validation

- Reproduced the incorrect advertisement by running the new regression test with `willSave: true`; it failed with `left: Some(true), right: Some(false)`.
- `cargo fmt --manifest-path baml_language/Cargo.toml -p bex_project -- --check`
- `cargo test --manifest-path baml_language/Cargo.toml -p bex_project capabilities_do_not_request_will_save_notifications`
- `git diff --check origin/canary...HEAD`
- Merged current `canary` at `593a513632` without conflicts.


sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
Sorry for the long wait here, this should get merged today and released soon!

PR #3378, opened 2026-04-17
Implement bound and unbound method closure types

hellovai:
[implement-bound-and-unbound-method-closure-types](https://cloud.codelayer.cloud/deep/tasks/019d99be-aa06-7956-8ad5-f8318877c63d) | [Artifacts](https://cloud.codelayer.cloud/tasks/019d99be-aa06-7956-8ad5-f8318877c63d/artifacts)

## What problems was I solving

BAML methods like `p.get_name` (bound to an instance) and `Person.get_name` (unbound, expects `self` as first arg) were both typed identically with `self` in the parameter list. Calling a bound method reference like `let cb = p.get_name; cb()` failed with arity mismatch (TIR counted `self`) and stack underflow at runtime (receiver never captured). This made first-class method references unusable as callbacks, `.map()` arguments, or strategy-pattern values.

Now:
- `let cb = p.get_name; cb()` works end-to-end with `cb` typed as `() -> string`
- `let cb2 = Person.get_name; cb2(p)` works with `cb2` typed as `(self: Person) -> string`
- Both forms work as higher-order arguments to `.map()` and custom functions
- Bound methods capture the receiver by reference (mutations visible after bind)
- Bound methods work inside lambdas (capture propagation)

## What user-facing changes did I ship

- Bound method references (`instance.method`) now strip `self` from the callable type and capture the receiver
- Unbound method references (`ClassName.method`) keep `self` as the first parameter
- Arity/type mismatch errors correctly reflect the bound vs unbound distinction
- Function subtype checking now enforces exact parameter count matching (previously allowed extra params silently)

## How I implemented it

Follows Go's model: bound method values become closure-like objects that capture the receiver. Unbound method values are plain function references with `self` as the first positional parameter.

### TIR Layer — Type Inference
- `crates/baml_compiler2_tir/src/inference.rs` — Split `MemberResolution::Method` into `BoundMethod` / `UnboundMethod` variants. Added lambda capture re-seeding: ancestor scope types are propagated to lambda captures (previously all `Ty::Unknown`), enabling bound method references inside lambda bodies.
- `crates/baml_compiler2_tir/src/builder.rs` — `resolve_member` takes `bound: bool` parameter. When bound, strips `self` via `skip_self_param`. `MemberAccess` determines bound vs unbound by checking if base is a runtime value or type name. `infer_local_rooted_path` propagates `root_is_value` through path segments.
- `crates/baml_compiler2_tir/src/normalize.rs` — Fixed function subtype arity check: changed `params2.len() > params1.len()` to `params1.len() != params2.len()` so that assigning a 1-param function to a 0-param annotation correctly errors.

### MIR Layer — Intermediate Representation
- `crates/baml_compiler2_mir/src/ir.rs` — Added `Rvalue::MakeBoundMethod { item_ref, receiver }`.
- `crates/baml_compiler2_mir/src/lower.rs` — `lower_member_access` emits `MakeBoundMethod` for bound refs, `Constant::Function` for unbound. `lower_path_expr` handles multi-segment receiver paths. `lower_call` immediate-call paths resolve callee directly via `resolution_to_item_ref` to avoid double-wrapping. Added `capture_indices` fallback for receivers that are captured variables.
- `crates/baml_compiler2_mir/src/pretty.rs` — Pretty-prints `make_bound_method <item>(<receiver>)`.
- `crates/baml_compiler2_mir/src/optimize.rs` — All five MIR optimization passes handle `MakeBoundMethod`.

### Bytecode Layer — Emission
- `crates/bex_vm_types/src/bytecode.rs` — Added `Instruction::MakeBoundMethod(GlobalIndex)`.
- `crates/bex_vm_types/src/types.rs` — Added `Object::BoundMethod(BoundMethod)` with `function: HeapPtr` and `receiver: Value`.
- `crates/baml_compiler2_emit/src/emit.rs` — Emits receiver, resolves global index, emits `MakeBoundMethod` instruction.
- `crates/baml_compiler2_emit/src/stack_carry.rs` — Stack effect simulation for `MakeBoundMethod` (pop receiver, push bound method).

### VM Layer — Runtime Execution
- `crates/bex_vm/src/vm.rs` — `MakeBoundMethod` handler creates `Object::BoundMethod`. `CallIndirect` detects bound methods, inserts receiver at `args_offset` via `stack.insert`, dispatches with full arity. New `resolve_bound_method_callee` helper handles `YieldToCall` paths (needed for `Array.map(bound_method)`). Updated all exhaustive `Object` matches.
- `crates/bex_heap/src/gc.rs` — GC traces both `function` (HeapPtr) and `receiver` (Value) of `BoundMethod`.

### Tests
- `crates/baml_tests/projects/closures/closures.baml` — 10 snapshot test functions covering bound mapper, unbound map, bound validator, field chain bind, strategy pattern, unbound higher-order, mutation-after-bind, generic bind, bound-in-lambda, repeated calls.
- `crates/baml_tests/projects/closure_errors/closure_errors.baml` — 5 error test functions for arity/type mismatches.
- `crates/baml_tests/tests/closures.rs` — 9 execution tests with inline BAML, all passing.

### Implemented as planned
- All 5 phases (Tests, TIR split, MIR MakeBoundMethod, Bytecode emission, VM execution) implemented as specified
- `MemberResolution::Method` split, `resolve_member` bound flag, `Rvalue::MakeBoundMethod`, `Instruction::MakeBoundMethod`, `Object::BoundMethod`, `CallIndirect` receiver insertion — all match the plan

### Deviations/surprises
- **Immediate call paths bypass `lower_to_operand`**: The plan only said to update match predicates, but `lower_to_operand` now calls `lower_member_access` which emits `MakeBoundMethod` — for immediate calls the receiver must be passed explicitly as `self`, not wrapped. Fixed by resolving callee directly via `resolution_to_item_ref`.
- **`MemberAccess` bound detection for type-name bases**: The plan assumed `MemberAccess` always has a value base (`bound = true`). But `Factory<int>.create(42)` parses as `MemberAccess` with a type-name base. Fixed by checking if the base is a local variable vs a type name.

### Additions not in plan
- **Lambda capture re-seeding** (`inference.rs`, ~148 lines): Lambda scope inference seeded all captures as `Ty::Unknown`, breaking bound method refs inside lambdas. Added ancestor-scope type propagation for captures.
- **`capture_indices` fallback in MIR lowering**: When receiver is a captured variable (not a local), emit `Place::Capture(cap_idx)` instead of `Constant::Null`.
- **`resolve_bound_method_callee` VM helper**: Handles `YieldToCall` paths so `Array.map(bound_method)` works (native builtins dispatch via `YieldToCall`, not `CallIndirect`).
- **`normalize.rs` subtype fix**: Strict equality on function param counts — pre-existing bug exposed by the new bound/unbound type distinctions.
- **`bound_in_lambda` execution test**: 9th test (plan specified 8) exercising the capture propagation fix.

### Items planned but not implemented
- None — all planned items were implemented.

### Setup
```bash
git fetch
git checkout vbv/implement-bound-and-unbound-method-closure-types
cd baml_language
```

### Automated Tests
```bash
# All 9 execution tests pass
cargo test -p baml_tests --test closures

# All snapshot tests pass
cargo test -p baml_tests -- closures
cargo test -p baml_tests -- closure_errors

# Full regression suite — 0 failures
cargo test -p baml_tests
```

### Manual Testing
- [ ] Review TIR snapshot: bound refs show self stripped, unbound refs keep self
- [ ] Review MIR snapshot: bound refs show `make_bound_method`, unbound show `const fn`
- [ ] Review bytecode snapshot: `MAKE_BOUND_METHOD` instructions for bound refs
- [ ] Review diagnostics snapshot: all 6 expected errors fire

Implement bound and unbound method closure types: `p.method` captures receiver and strips `self` from type; `Class.method` keeps `self` as first parameter. Both work as first-class values in `.map()`, higher-order functions, and lambdas.

## Summary by CodeRabbit

* **New Features**
  * First-class bound methods: create, format and call method values bound to instances; VM and bytecode support included.
  * Type-system distinction between bound vs unbound method resolutions.

* **Bug Fixes**
  * Prevent bound methods from being mis-converted or matched as other runtime types.

* **Tests**
  * Added extensive tests exercising bound/unbound method and closure behaviors.


hellovai (commented):
**Re: `matches!(rvalue, ...)` ownership concern on `emit.rs:1974`**

This is a false positive. In Rust, `matches!(rvalue, ...)` expands to `match rvalue { ... => true, _ => false }`, and matching a local variable doesn't move it — the compiler auto-refs in match expressions. The code compiles cleanly and passes `cargo clippy -D warnings`. `rvalue` is successfully used on both subsequent lines (1975: `self.emit_rvalue_pull(&rvalue)` and 1978: `Box::new(rvalue)`) without issue. No change needed.

PR #3450, opened 2026-05-01
Fix TypeBuilder recursive class cycles

### LINKED ISSUE #3390: [bug] Imperatively-built recursive classes crash ctx.output_format renderer
### Product

BAML

### Describe the bug

A class added imperatively via TypeBuilder (`tb.add_class` + `cb.add_property`) that references itself crashes the runtime during `ctx.output_format` rendering with a native stack overflow. The same class declared statically in BAML source renders correctly.

The failure is in `OutputFormatContent` building / Jinja printing of the class fields. Parse-only paths against the same schema succeed (no prompt rendering involved), which narrows it to the output-format renderer rather than schema construction or validation.

### Reproduction Steps

`baml_src/main.baml`:

```baml
class DynamicOutput {
  @@dynamic
}

function ExtractItems(input: string) -> DynamicOutput {
  client "openai/gpt-4o-mini"
  prompt #"
    {{ ctx.output_format }}
    {{ input }}
  "#
}
```

`repro.py`:

```python
import asyncio
from baml_client import b
from baml_client.type_builder import TypeBuilder

async def main():
    tb = TypeBuilder()
    item = tb.add_class("Item")
    item.add_property("name", tb.string())
    item.add_property("children", item.type().list())   # self-reference
    tb.DynamicOutput.add_property("items", item.type().list())

    # Fails inside ctx.output_format render, before any HTTP egress.
    await b.ExtractItems(input="anything", baml_options={"tb": tb})

asyncio.run(main())
```

Control (renders fine, equivalent shape declared statically):

```baml
class Item {
  name string
  children Item[]
}

class DynamicOutput {
  items Item[]
  @@dynamic
}
```

### Root cause

`OutputFormatContent.recursive_classes` is populated from two places in `engine/baml-runtime/src/internal/prompt_renderer/render_output_format.rs` (canary L459, L465):

```rust
for cycle in ir.finite_recursive_cycles() { ... }
for cycle in &ctx.recursive_class_overrides { ... }
```

The first is the static IR. The second comes from `TypeBuilder::recursive_classes`, which is only written inside `TypeBuilder::add_baml` after the validation pipeline runs — `engine/baml-runtime/src/type_builder/mod.rs` canary L717-L720:

```rust
self.recursive_classes()
    .lock()
    .unwrap()
    .extend(recursive_classes);
```

`ClassBuilder::add_property` never touches `recursive_classes`. `TypeBuilder::to_overrides` (canary L725, L822) just clones the existing field:

```rust
let recursive_classes = self.recursive_classes.lock().unwrap().clone();
```

So an imperatively-built self-ref reaches the renderer with empty cycle overrides, the renderer never registers it as recursive, and the printer descends into the field type forever. Native stack overflow follows.

### Expected behavior

Same render behavior for an imperatively-built recursive class as for the equivalent statically-declared recursive class.

### Proposed fix direction

Run cycle detection inside `TypeBuilder::to_overrides()` over the imperatively assembled class graph and write the resulting cycles into the returned `recursive_classes`. The validation pipeline already does this for `add_baml`; lifting that step into a shared helper that operates on the assembled `IndexMap<String, RuntimeClassOverride>` would converge the two paths.

### BAML Version

0.221.0

### Language/Framework

Go

### Operating System

Linux

suyua9:
## What

Fixes #3390 by detecting recursive class cycles that are assembled imperatively through `TypeBuilder` before returning runtime overrides.

`TypeBuilder::add_baml` already receives recursive class cycle metadata from the validation pipeline, but `tb.add_class(...)` + `class.add_property(...)` did not populate `recursive_classes`. When such a class referenced itself, `ctx.output_format` could render it without recursive-class overrides and recurse until a native stack overflow.

This change:

- walks the classes currently stored in `TypeBuilder`
- builds a class-reference graph from property types, including lists, maps, tuples, arrows, and unions
- adds any detected class cycles to the recursive class overrides returned by `to_overrides()`
- extends the existing recursive property test to cover a `Node[]` self-reference and assert that it is reported as a recursive class cycle

## Tests

- `cargo fmt --manifest-path engine/Cargo.toml --package baml-runtime --check`
- `git diff --check`

I also attempted:

```bash
cargo test --manifest-path engine/Cargo.toml -p baml-runtime type_builder::tests::test_recursive_property
```

but this local machine cannot link Rust build scripts because its Xcode Command Line Tools install has an architecture mismatch in `libxcrun.dylib` (`x86_64`, while the active toolchain is building for `arm64`). The failure happens before `baml-runtime` is compiled.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable detection of recursive class cycles during conversion, merged with any existing cycle info to avoid duplicates.
  * Recursive properties now display as collection types (e.g., child → Node[]) and recursive cycle metadata correctly includes the Node cycle.

### DIFF
--- a/engine/baml-runtime/src/type_builder/mod.rs
+++ b/engine/baml-runtime/src/type_builder/mod.rs
@@ -19,6 +19,7 @@ use crate::{
 };
 
 type MetaData = Arc<Mutex<IndexMap<String, BamlValue>>>;
+type ClassPropertySnapshot = Vec<(String, Option<TypeIR>, PropertyAttributes)>;
 
 trait Meta {
     fn meta(&self) -> MetaData;
@@ -735,12 +736,37 @@ impl TypeBuilder {
         Vec<IndexMap<String, TypeIR>>,
     ) {
         log::debug!("Converting types to overrides");
-        let cls = self
+        let class_entries = self
             .classes
             .lock()
             .unwrap()
             .iter()
             .map(|(name, cls)| {
+                let properties = {
+                    let cls = cls.lock().unwrap();
+                    cls.properties.lock().unwrap().clone()
+                };
+                let property_snapshot = properties
+                    .iter()
+                    .map(|(property_name, f)| {
+                        let attrs = PropertyAttributes::from(f);
+                        let t = {
+                            let property = f.lock().unwrap();
+                            let t = property.r#type.lock().unwrap();
+                            t.clone()
+                        };
+                        (property_name.clone(), t, attrs)
+                    })
+                    .collect::<ClassPropertySnapshot>();
+                (name.clone(), property_snapshot)
+            })
+            .collect::<Vec<_>>();
+
+        let detected_recursive_class_cycles = detect_recursive_class_cycles(&class_entries);
+
+        let cls = class_entries
+            .into_iter()
+            .map(|(name, properties)| {
                 log::debug!("Converting class: {name}");
                 let mut overrides = RuntimeClassOverride {
                     alias: None,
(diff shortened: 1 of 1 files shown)


suyua9:
Addressed the CodeRabbit follow-up in 4477f42:

- `to_overrides()` now snapshots the dynamic class property/type data once and uses that same immutable snapshot for both `RuntimeClassOverride` construction and recursive class detection, so the two outputs cannot observe different `self.classes` states.
- Replaced the per-start-node DFS/simple-path cycle enumeration with a one-time SCC pass over the class reference graph. The SCC output includes multi-class components and self-loops, which is the shape needed for `recursive_classes` without enumerating exponentially many paths.
- Removed the old cycle path enumeration helpers from the code path.

Validation:
- `cargo fmt --manifest-path engine/baml-runtime/Cargo.toml -- engine/baml-runtime/src/type_builder/mod.rs`
- `git diff --check`
- conflict-marker scan on `engine/baml-runtime/src/type_builder/mod.rs`

I also attempted the focused Rust test command:

`cargo test --manifest-path engine/baml-runtime/Cargo.toml type_builder --lib`

but this machine's CommandLineTools linker is currently unusable for arm64 builds (`libxcrun.dylib` is x86_64-only while rustc needs arm64e), so local linking fails before this crate is compiled. The PR branch has been updated and GitHub checks / CodeRabbit are now the authoritative validation surface.

sxlijin:
Closed and rejected: this is an ai-only drive-by pull request.

PR #3520, opened 2026-05-13
`spawn / await` foundation

antoniosarosi:
## Summary

Implements [BEP-034 v1 — concurrency via `spawn` / `await`](https://github.com/BoundaryML/beps/tree/main/BEP-034-concurrency-via-spawn-await): a new BAML concurrency model where any function can fan out work in parallel without function coloring. `spawn { body }` starts a green thread and returns a `Future<T, E>`; `await f` blocks the current thread until the future settles. Cancellation is ambient and cascades via `tokio_util::sync::CancellationToken` child-tokens — no parent/child tracking, no ctx-plumbing.

```baml
function main() -> int {
    let a = spawn { 10 };
    let b = spawn { 20 };
    let c = spawn { 30 };
    (await a) + (await b) + (await c)
}
```

End-to-end demos validate real wall-clock parallelism. An HTTP server written on this branch with `let listener = baml.net.listen(addr); while true { let sock = listener.accept(); spawn { handle(sock) } }` handles **200 concurrent curls in 0.69s** (vs ~100s sequential) — exactly the spawn-heavy workload that exposed the GC-park deadlock fix in this PR.

## Scope landed

All of v1 (phases A through G), plus the deviations from the original plan and the cleanup pass.

| Phase | What it does | Status |
|---|---|---|
| **A** — foundation | rename `dispatch_future` → `schedule_future`; extend `Ty::Future` to `(value, error, attr)`; add `BexThread` wrapper around `BexVm` | ✅ |
| **B** — runtime | `BexEngine::spawn_thread` (one tokio task per spawn); `run_thread_event_loop` with `ThreadOutcome::{RootValue, SettledChild}`; child cancel-token cascade via `parent.child_token()` | ✅ |
| **C** — compiler | lexer/parser/AST/HIR/TIR/MIR/bytecode for `spawn` / `await` / `Future<T, E>`. Spawn body lowers to a 0-arg synthetic lambda; reuses the existing closure pipeline | ✅ |
| **D′** — sys-op optimization | collapse the old sys-op `ScheduleFuture + Await` pair into a single `Instruction::SysOp(g)`. One round trip, no `Object::Future` allocation, no `FutureManager` entry. Net: faster IO for *every* program, not just spawn users | ✅ |
| **E** — `baml.future.Future` surface | `enum FutureState`, `class Future<T, E>` with `cancel` / `is_settled` / `is_cancelled` / `is_result` / `is_error` / `state` declared in `ns_future/future.baml` | ✅ |
| **F** — Future methods native | `$rust_function` impls in `bex_vm/src/package_baml/future.rs`. Each method operates on the heap `Object::Future` directly — no engine round-trip — and races safely with the producer via CAS on the discriminant | ✅ |
| **G** — tests | 18 integration tests covering spawn / await / cancel / methods / nested spawn / parent-throw cascade / fire-and-forget propagation. Plus a real parallel-sleep test that proves wall-clock concurrency | ✅ |

### Architecture: how parallelism actually happens

Three independent existing mechanisms compose to give real wall-clock parallelism:

1. **One `tokio::spawn` per `spawn { … }`** — tokio's multi-threaded runtime picks up tasks across worker threads.
2. **`HeapPermitManager` semaphore has `MAX_PERMITS = u32::MAX` tokens** — many mutators hold permits simultaneously without blocking each other; only GC parking is exclusive.
3. **TLAB exclusivity per `BexThread`** — each VM allocates into its own bump-pointer region; no allocator contention.

The runtime concept added by BEP-034 is `BexThread` — a `BexVm` plus scheduling metadata (`name`, `cancel`, `settles_future`, `pending_child_errors`). Cross-thread coordination lives on the heap `Future` object: `AtomicU8` state + `UnsafeCell<MaybeUninit<Value>>` + `Arc<tokio::sync::SetOnce>` + `CancellationToken`. CAS on the discriminant arbitrates producer vs `f.cancel()`.

Cancellation cascade is one of the more elegant parts: every spawn captures `parent_cancel.child_token()`. When the parent cancels (or unhandled-throws), tokio-util's child-token machinery fires every descendant token automatically. No parent/child tree to maintain.

### Phase D′ deviation from the plan

The original Phase D had every existing `baml.io / fs / http / net / env / shell` call site migrate to explicit `await sys_call(...)`. That would have:
- Surfaced `Future<T, E>` in stdlib signatures (a leaky pattern users would copy)
- Created two ways to produce a `Future<T, E>` (sys-op call vs `spawn`)
- Required updating hundreds of call sites and snapshots

**Dropped.** Sys-ops still look blocking from BAML source. Phase D′ replaces it with a much smaller optimization: collapse the schedule + await pair into a single `Instruction::SysOp(g)`. The user-visible surface is unchanged; the cost of every sys-op call is half what it was before this PR.

`Future<T, E>` in v1 is produced **only** by `spawn { … }`.

### Out of scope (v2+)

- `with` clause: `spawn "name" with withRetry(3), baml.spawn.options(group=g) { … }`
- `TaskGroup` (value-based rate limiting)
- `CancelToken` as a BAML-surface class (the runtime token exists and cascades; BAML code just can't construct one)
- `detach = true`
- `Future.then(on_ok, on_err?)`
- Combinators: `baml.future.{race, any, all, all_complete, all_settled}`
- `await` distribution over union types

## What else is in this PR

Beyond the v1 implementation, this branch also lands:

### Cleanup pass (post-implementation review)

- **Pruned test suite from 26 → 18 tests** with no coverage loss. The original suite had drifted into one-assertion-per-test territory; e.g. `future_methods.rs` had 10 tests calling one method each after the same `spawn + await` setup. Consolidated to 4 tests grouped by terminal state.
- **Rewrote `spawn_parallel.rs`** which previously admitted in its own comments that it didn't actually test parallelism. Now uses three concurrent `baml.sys.sleep(200)` calls; asserts elapsed < 1000ms (parallel ≈ 200ms + bootstrap vs sequential > 1s).
- **De-duped 6 settle-as-cancel / settle-as-err code paths** in `run_thread_event_loop` into two helpers (`settle_child_cancelled`, `settle_child_errored`). The `ThrownUnhandled` and `Thrown` arms each went from ~40 to ~10 lines. Behavior identical.
- **Switched `is_cancelled_panic` from class-name string comparison to direct `HeapPtr` equality** against the cached `resolved_class_names["baml.panics.Cancelled"]`. Drops one heap walk per cancel detection.
- **Deleted dead code in `thread.rs`**: `ThreadId` (and its `next`/`raw`/`Display` impls), the `BexThread::id` and `BexThread::parent` fields, `_assert_send`/`_assert_sync` stubs. Cascade and propagation are token-driven, not id-driven. Removed the `#![allow(dead_code, unreachable_pub)]` attribute as a result.

### Stdlib additions used by the spawn demos

- `baml.io.print/println/eprint/eprintln` — stdout/stderr writes. `println` coalesces `s + "\n"` into one `write_all` so concurrent calls from spawned threads don't interleave on a line boundary.
- `baml.net.listen(addr)` + `TcpListener.{accept, close}` + `Socket.{write, close}` — a per-connection accept loop is the canonical spawn shape and the basis of the HTTP-server demo.

Backends: tokio-backed in `sys_native` (CLI); `Unsupported` stubs in `sys_ops::DefaultIoOps`, `playground_io`, and `bridge_wasm`.

### Bug fixes discovered during BEP-034 work

- **`bex_heap`: HeapPermitManager GC-park AB-BA deadlock**. `request_park` took the `holders` mutex before awaiting `acquire_many(MAX_PERMITS)`; `new_permit` needed the same mutex. Latent on canary since #3386 ("New garbage collector") but invisible to single-call workloads. BEP-034's spawn-heavy programs make it reproducible in seconds — the HTTP-server demo with concurrent curls hangs deterministically without this fix. Reorder: drain the semaphore first (that's the actual stop-the-world barrier), take the mutex only for the brief snapshot work. Reproducer test + writeup also landed.
- **`baml_fmt`: empty raw-string interior**. `RawString::print` assumed the trimmed interior split on lines always yielded ≥1 element; `#"\n"#` trimmed to empty caused `unwrap_or_else(|| unreachable!())` to panic.
- **`fix A` and `fix B`** during the BEP work itself: race await against thread cancel + idempotent settle; fire-and-forget child errors propagate to parent at the parent's next await checkpoint.

### Addressed CodeRabbit review

Triaged 20 inline comments. Addressed 7 real issues; 1 false positive; 12 nice-to-haves deferred to follow-up:

| # | Issue | Resolution |
|---|---|---|
| CR-1 | Parser misparses `spawn nm { y: 1 }` when body looks like fields | Added `suppress_object_literal_depth` counter; regression tests in `spawn_name_object.rs` |
| CR-2 | (false positive) call-arg classification missing SPAWN/AWAIT | Verified `add(await spawn { 10 }, await spawn { 20 })` works; no `is_expr_node_kind` helper exists |
| CR-3 | MIR `lower_await` doesn't normalize projection dests to local | Mirror `lower_call`'s temp-local pattern |
| CR-4 | `throws_analysis.rs` walked spawn_body, leaking body throws to spawning function | Removed the walk |
| CR-5 | Fire-and-forget carve-out at await checkpoint was keyed on `HeapPtr`, broke on producer-settled-first races | Refactored `ChildErrorQueue` to store `(FutureId, HeapPtr)`; carve-out keys on stable `FutureId` |
| CR-6 | `await` on non-Future silently returned `Unknown` | Emit `TypeMismatch` with `Future<unknown, unknown>` as expected |
| CR-7 | `await f` didn't add `f`'s E type to surrounding throws set | `flatten_ty_to_facts(E)` in TIR Await arm |
| CR-8 | `settle_internal_error` ignored CAS result, could silently lose engine errors to concurrent cancel | Log at WARN; cancel-wins is intentional |

## Test plan

- [x] `cargo test --workspace` green (only pre-existing flake: `heap_guard_deadlock` test is the reproducer for the AB-BA deadlock fix in this PR — un-`#[ignore]`d on this branch, passing).
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean.
- [x] Spawn / cancel / futures test suite (18 tests): green.
- [x] Wall-clock parallel test: 3 × 200ms sleep, total < 1000ms.
- [x] End-to-end stress: HTTP-server demo on this branch handles 200 concurrent curls in 0.69s.
- [ ] Reviewer: pull this branch, build `baml-cli`, point it at any sufficiently-spawn-heavy BAML program to confirm.

## Commit organization for review

23 commits, organized so each phase is a single commit:

```
beb2d61d2 BEP-034 phase A: rename dispatch_future → schedule_future, extend Ty::Future, add BexThread
8a9157f6d BEP-034 phase B: multi-threaded scheduling runtime
b50bf1d0b BEP-034 phase C: parser → MIR for spawn / await / Future<T,E>
027414afe BEP-034 phase E: declare baml.future.Future + FutureState surface
b7fcfaf4e BEP-034 phase G + bug fixes: end-to-end spawn / await works
d0948771c BEP-034 phase D′: collapse sys-op schedule+await into single SysOp instruction
ef17d31b1 BEP-034 phase F: native methods on baml.future.Future via $rust_function
a5c5be6b6 BEP-034: spawn body's actual throws populates Future<T, E>'s E
204f9454d BEP-034 fix A: race await against thread cancel + idempotent settle
1b8d744e4 BEP-034 fix B: fire-and-forget child errors propagate to parent
34956d7a2 BEP-034 phase G: fill in missing v1 test coverage
c0ea84f6e BEP-034: resolve baml.future.Future<T,E> as Ty::Future; fix catch+await pre-emption
e007fbca3 BEP-034 review: prune duplicate spawn tests; add real parallel test
6f145df9f BEP-034 review: dedupe settle paths, drop dead thread metadata
75bc38a3d stdlib: add baml.io.print/println/eprint/eprintln and baml.net.listen
1aea78f3a BEP-034 doc: clarify Future<T,E> is only produced by spawn
cc67bfbd2 baml_fmt: handle empty raw-string interior without panicking
219b3d3b9 bex_heap: add heap_guard AB-BA deadlock reproducer
921ea7d49 bex_heap: fix HeapPermitManager GC-park AB-BA deadlock
f6b43ded7 Merge remote-tracking branch 'origin/canary'
b66c207f4 BEP-034 review: address CodeRabbit feedback (7 fixes, 1 false positive)
```

Easiest review is commit-by-commit. The merge commit is mechanical (canary's GC Fixes #3514 added write-barriers to the future-resolution path; merged with the CAS-based settle methods that BEP-034 needs for cross-thread cancel coordination).


antoniosarosi:
Responding to the CodeRabbit `CHANGES_REQUESTED` review from May 18.

I triaged all 20 inline comments in commit b66c207f4. Summary:

**Addressed (7 real issues):**
- CR-1: parser misparses `spawn nm { y: 1 }` — added `suppress_object_literal_depth` flag in `parse_spawn_expr`, with two regression tests in `spawn_name_object.rs`.
- CR-3: `lower_await` now normalizes projection dests to a temp local, mirroring `lower_call`.
- CR-4: removed the spawn-body walk in `throws_analysis.rs` that leaked body throws into the spawning function's escaping set.
- CR-5: refactored `ChildErrorQueue` to store `(FutureId, HeapPtr)` and key the await-checkpoint carve-out on `FutureId`, which is stable across producer-settle and GC moves. Fixed a latent timing bug in `methods_on_errored_future` that previously passed only when scheduling happened to put the parent before the child.
- CR-6: `await` on a non-Future operand now emits a `TypeMismatch` with `Future<unknown, unknown>` as the expected type instead of silently returning `Unknown`.
- CR-7: TIR `collect_throw_facts_from_expr` Await arm now adds the future's E-channel to the surrounding throws set via `flatten_ty_to_facts`.
- CR-8: `settle_internal_error` now checks the CAS result and logs at WARN if it loses to a concurrent terminal transition (cancel-wins is intentional).

**False positive (1):**
- CR-2: CodeRabbit referenced an `is_expr_node_kind` helper that doesn't exist. The actual dispatcher (`lower_expr_inner`) already routes SPAWN_EXPR / AWAIT_EXPR correctly. Verified by integration test.

**Deferred to follow-up (12 nice-to-haves):**
- Display lowercase `future<...>` in TIR, SettledChild diagnostic context, spawn opcode arg validation, shallow Spawn equality, onionskin compact rendering, etc. None of these are correctness-affecting; they're style/diagnostic refinements that fit better in a polish PR.

Branch is now rebased on canary HEAD and CI is fresh. Requesting re-review.

PR #3583, opened 2026-05-27
Harden spawn-raced arithmetic and field access

antoniosarosi:
Follow up on #3581.

## Summary
- skip specialized int/float arithmetic emission when either operand reads a captured local or capture-derived value
- keep ordinary non-captured arithmetic on the specialized opcode path
- convert stale VM field-slot access into catchable `baml.panics.InvalidFieldAccess` instead of letting an out-of-bounds field index become a Rust panic
- add bytecode/execution coverage for captured scalar arithmetic, captured `float[]` element arithmetic, and a deterministic stale field-slot access

## Why not checked field instructions here?
We considered adding checked field opcodes that encode the expected class along with the field slot. That would catch more stale-narrowing cases than bounds checking, including same-width class layouts where the stale field index is still in range.

We are intentionally not doing that in this PR. To emit checked ops only where needed, the compiler would need to prove whether the receiver can be accessed in parallel or aliases shared mutable state. That gets conservative very quickly: captured/cell values, parameters, container elements like `(A | B)[]` and `A[] | B[]`, map entries, field-derived receivers, nested combinations, virtual/inlined locals, and `InitSpread` would all need careful treatment. Something as simple as a function parameter prevents the compiler from proving the value is not shared with spawned code.

The practical result is that checked field access would become the default for much of real code, either widening hot field instructions or adding side-table lookups plus an extra branch. Bounds checking is narrower: it preserves the current bytecode shape and prevents the host/runtime crash class we care about now. It may still allow user-visible logic races, but stale out-of-range field slots become catchable BAML panics rather than Rust panics.

## Tests
Passed locally:
- `cargo test -p baml_tests --test spawn_basic`
- `cargo test -p baml_tests --test operators assign_add`
- `cargo test -p baml_tests --test exceptions catch_stale_field_slot_invalid_field_access`
- `cargo test -p baml_tests --test exceptions catch_index_out_of_bounds`
- `cargo test -p bex_vm_types --lib`
- `cargo test -p bex_vm --lib`
- `cargo test -p baml_cli describe_command_tests::render_builtin_package_listing`

Attempted locally:
- `cargo test --lib` fails before test execution while linking `bridge_nodejs` test binaries: unresolved Node N-API symbols such as `_napi_call_threadsafe_function`.
- `cargo test --workspace --all-targets --all-features` fails before test execution while linking bridge test binaries: unresolved Node N-API symbols and Python/PyO3 symbols such as `_PyLong_AsLong`.

Hook also ran `cargo fmt` and `cargo clippy --workspace --all-targets --all-features -- -D warnings` during commit.

## Summary by CodeRabbit

* **New Features**
  * New structured InvalidFieldAccess panic for clearer error reporting.
  * Public safe field access helpers: field_len, try_load_field, try_store_field.

* **Bug Fixes**
  * Prevented invalid field-index reads/writes across several VM ops; field-access now validates and throws InvalidFieldAccess.

* **Tests**
  * Added regression tests for field-access error handling and spawn bytecode specialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/BoundaryML/baml/pull/3583?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


antoniosarosi:
Addressed the CodeRabbit follow-ups in `c342d9711`:

- Added unit coverage for `Instance::{field_len, try_load_field, load_field, try_store_field}` in `bex_vm_types`, including the out-of-bounds panic case.
- Ran and passed `cargo test -p bex_vm_types --lib`.
- Ran and passed `cargo test -p bex_vm --lib`.
- Attempted `cargo test --lib`, but this local shell fails before test execution while linking `bridge_nodejs` test binaries due unresolved Node N-API symbols such as `_napi_call_threadsafe_function`.
- Attempted `cargo test --workspace --all-targets --all-features`, but this local shell fails before test execution while linking bridge test binaries due unresolved Node N-API symbols and Python/PyO3 symbols such as `_PyLong_AsLong`.

The PR description has been updated with the same validation details. The commit hook also passed `cargo fmt` and `cargo clippy --workspace --all-targets --all-features -- -D warnings`.

antoniosarosi:
Follow-up CI fix pushed in `8ec60280`: updated `baml_cli__describe_command_tests__render_builtin_package_listing.snap` for the new `panics.InvalidFieldAccess` builtin listing.

Validation: `cargo test -p baml_cli describe_command_tests::render_builtin_package_listing` passes locally.

PR #3596, opened 2026-05-28
feat(spawn): spawn options — `with baml.spawn.options(...)`, CancelToken, detach, TaskGroup (BEP-034)

antoniosarosi:
Implements the **spawn-options core** of BEP-034 — the `with` clause on `spawn` and the primitives it configures. Builds on the merged `spawn`/`await`/`Future<T,E>` foundation.

```baml
let tok = baml.spawn.CancelToken.new();
let group = baml.spawn.TaskGroup.new(limit = 5);
let f = spawn "fetch" with baml.spawn.options(group = group, cancel = tok, detach = false) {
    baml.http.fetch(url)
};
await f
```

## What's included (4 commits)

1. **`with baml.spawn.options(cancel = ...)` + `CancelToken`** — the `with` clause end-to-end (parser → AST → TIR → MIR → emit → VM → engine), the `baml.spawn` namespace, and `CancelToken` (`new`/`any`/`cancel`/`is_cancelled`). A user token links into a spawn's effective cancel token via a self-terminating watcher; firing it cancels the child and `await` re-throws `Cancelled` (catchable).
2. **`detach = true` + `OmittedArg` codegen fix** — a detached spawn gets a fresh effective token (parent cascade/throw doesn't reach it) and routes unhandled fire-and-forget errors to the root task. Also fixes the codegen so optional builtin params (incl. primitives like `bool?`) accept an omitted argument (general improvement).
3. **`TaskGroup` rate limiting** — `TaskGroup.new(limit, name?)`, `cancel(pending?, active?)`, `set_limit`, `limit`, `name`, `active_count`, `queued_count`, and `options(group = ...)`. Members register synchronously at spawn (so `cancel()`/counts are exact immediately) and park asynchronously for a slot (FIFO, without holding the heap permit), releasing on settle. Admission is cancel-biased so a `group.cancel()`'d member never runs.
4. **Strict `with` validation + `spawn_parallel` fix** — TIR rejects `with` expressions that aren't a `baml.spawn.options(...)` config (type-based, so `let c = options(...); spawn with c` works too). Also fixes `spawn_parallel.rs`, which was a false-positive (it used `sys_ops`, whose `sleep` is `Unsupported`, so its "parallel sleeps" were no-ops) — it now uses `sys_native` real sleeps and genuinely verifies concurrency.

## Notable decisions / findings

- `baml.spawn` collided with the `spawn` **keyword** → `spawn`/`await` are now accepted as member-path segments after `.` (mirrors the existing `KW_CLIENT` precedent), so `baml.spawn.options(...)` parses exactly as the BEP writes it.
- **Spawns do run concurrently for real IO** — verified with execution-only timing: `limit=3` runs three 200ms sleeps in ~200ms; `limit=1` serializes them (~600ms).
- Static methods (`TaskGroup.new`, `CancelToken.new`/`any`) use the existing non-`self` builtin-fn mechanism (like `int.parse`).

## Testing

17 engine cancellation tests, 7 TaskGroup tests (+2 admission unit tests), `ns_cancel_token`/`ns_task_group` BAML fixtures, `baml_src` (1500+ BAML tests + bytecode/CLI targets), `baml_tests --lib`, plus parser/AST/MIR/emit/codegen suites — all green; clippy clean; snapshots regenerated.

## Deferred (out of this scope)

Combinators (`race`/`any`/`all`/`all_complete`/`all_settled`) + `Future.then()`, the full user-middleware system (`SpawnParams` transformers), `await`-over-unions, shielding, `Duration`/`CancelToken.after`, and `TaskGroup.on_event`.

> Branch is currently a few commits behind `canary`; will rebase before merge.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added `spawn ... with baml.spawn.options(...)` syntax to configure spawned tasks
  * Introduced `CancelToken` for canceling spawned operations
  * Introduced `TaskGroup` for rate-limiting concurrent spawns with configurable concurrency caps
  * Added `detach` flag to allow spawned tasks to complete independently without blocking

* **Tests**
  * Added spawn cancellation behavior validation
  * Added task group rate-limiting verification tests

[Review Change Stack](https://app.coderabbit.ai/change-stack/BoundaryML/baml/pull/3596?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -1737,6 +1737,8 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-util",
+ "wasm-bindgen-futures",
  "web-time",
 ]
 
--- a/baml_language/crates/baml_builtins2/baml_std/baml/ns_spawn/spawn.baml
+++ b/baml_language/crates/baml_builtins2/baml_std/baml/ns_spawn/spawn.baml
@@ -54,10 +54,12 @@ class CancelToken {
   }
 }
 
-/// Assemble a `SpawnConfig` for a `spawn ... with` clause. In v1 only `cancel`
-/// is honored at runtime; `group` is accepted for forward compatibility and
-/// wired in a follow-up. (`detach` lands alongside its runtime support.)
+/// Assemble a `SpawnConfig` for a `spawn ... with` clause. `cancel` links a
+/// cancel token into the spawn's effective token; `detach = true` opts the
+/// spawn out of the parent→child cancel cascade and routes its unhandled
+/// errors to the root task instead of the spawner. (`group` is accepted for
+/// forward compatibility and wired in a follow-up.)
 //baml:mut_vm
-function options(group: TaskGroup? = null, cancel: CancelToken? = null) -> SpawnConfig {
+function options(group: TaskGroup? = null, cancel: CancelToken? = null, detach: bool? = null) -> SpawnConfig {
   $rust_function
 }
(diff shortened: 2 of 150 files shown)


antoniosarosi:
Superseded by #3625, which folds the spawn-options work and the BEP-034 future combinators into a single PR.

PR #3619, opened 2026-05-31
fix(beps): generate API tokens with crypto randomness

xuyua9:
## Summary
- Replace BEP API token generation based on `Math.random()` with `crypto.getRandomValues()`.
- Keep the existing `bep_` prefix and 32-character alphanumeric token format.
- Use rejection sampling so random bytes map evenly to the token alphabet.

## Testing
- `git diff --check`
- `node` smoke test for token format/uniqueness using the same generator logic
- `pnpm -C typescript/apps/beps exec tsc --noEmit --project convex/tsconfig.json` *(blocked locally: dependencies are not installed in this worktree, so Convex/Node modules cannot be resolved)*

## Summary by CodeRabbit

* **Bug Fixes**
  * API token generation has been upgraded with enhanced randomization and improved entropy handling mechanisms. The update reduces selection bias, strengthens overall token security, and ensures more unique and reliable tokens for API authentication and secure communication.

### DIFF
--- a/typescript/apps/beps/convex/users.ts
+++ b/typescript/apps/beps/convex/users.ts
@@ -304,13 +304,32 @@ export const linkSlackUserId = internalMutation({
 // API TOKEN MANAGEMENT
 // ─────────────────────────────────────────────────────────────────────────────
 
+const API_TOKEN_ALPHABET =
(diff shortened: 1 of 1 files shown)


sxlijin:
ai spam

PR #3621, opened 2026-05-31
fix(promptfiddle): compare log decorations by byte length

xuyua9:
## Summary
- Match the WASM playground log-decoration heuristic to the WebSocket path by comparing UTF-8 byte length.
- Rust source spans are byte offsets, so this avoids hiding inline log decorations for non-ASCII values.

## Testing
- `git diff --check`
- targeted static check that the worker no longer compares `message.length > sourceSpanLength`
- `node` smoke test showing a non-ASCII value (`你好世界`) is detected by UTF-8 byte length where UTF-16 code-unit length failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of inline log display in the editor by fixing byte-length calculation logic.


sxlijin:
ai spam

PR #3623, opened 2026-05-31
fix: report line-local columns for VM events

wuyua9:
## Summary

- track the 0-indexed source column alongside each bytecode line-table entry
- preserve that column when lowering bytecode to compact form
- use the recorded line-local column for `SendEvent` source locations while keeping start/end offsets unchanged

## Tests

- `rustfmt --edition 2024 baml_language/crates/baml_compiler2_emit/src/emit.rs baml_language/crates/bex_vm/src/vm.rs baml_language/crates/bex_vm_types/src/bytecode.rs`
- `git diff --check`
- Not run: `cargo test -p bex_vm_types line_table_translated --manifest-path baml_language/Cargo.toml` cannot link in this local macOS environment because `xcrun` fails to load the installed CommandLineTools `libxcrun.dylib` as arm64/arm64e while compiling build scripts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced source location tracking to include precise column offsets alongside line numbers in debug information, enabling more accurate error reporting and debugging capabilities.

### DIFF
--- a/baml_language/crates/baml_compiler2_emit/src/emit.rs
+++ b/baml_language/crates/baml_compiler2_emit/src/emit.rs
@@ -1050,11 +1050,18 @@ impl<'ctx, 'obj> StackifyCodegen<'ctx, 'obj> {
         self.bytecode.instructions.len()
     }
 
-    /// Convert a byte offset to a 1-indexed line number.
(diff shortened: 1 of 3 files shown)


sxlijin:
ai spam

PR #3625, opened 2026-06-01
Spawn options + future combinators

antoniosarosi:
BEP-034 — concurrency via `spawn` / `await`. Single PR covering the spawn-options core and the future combinators (supersedes #3596).

## Public BAML API (BEP-034 concurrency)

This is the complete user-facing surface this PR adds. Everything below is callable from `.baml` source.

### Syntax

```baml
spawn { body }                       // -> Future<T, E>; runs body concurrently
await f                              // -> T; blocks until f settles, re-throws its error
spawn with t1 with t2 { body }       // middleware: applies transformers left-to-right
```

- `spawn { e }` evaluates to `Future<T, E>` where `T` is the body's value type and `E` its throws type. The body starts running immediately; the future is the handle.
- `await f` suspends until `f` settles, then yields its value or re-throws its error. `await` **distributes over a union of futures**: `await (Future<A,E1> | Future<B,E2>)` yields `A | B` (the error side is contributed to the awaiter's throws surface).
- `spawn ... with <transformer> ...` runs each transformer over an implicit `SpawnParams` before the runtime creates the future (see middleware below).

### `baml.future` — futures & combinators

```baml
class Future<T, E> {
  fn cancel(self) -> bool throws never        // request cancel; false if already settled
  fn is_settled(self) -> bool throws never     // reached Ready | Error | Cancelled
  fn is_cancelled(self) -> bool throws never
  fn is_result(self) -> bool throws never      // settled with a value
  fn is_error(self) -> bool throws never        // settled with an error
  fn state(self) -> FutureState throws never
}

enum FutureState { Pending  Ready  Error  Cancelled }

// Combinators over a homogeneous Future<T,E>[] — each returns a new Future, so they compose.
fn all_complete<T, E>(futures: Future<T, E>[]) -> Future<T[], E>
//   all values in input order; losers KEEP RUNNING on failure; throws first error (input order)
fn all<T, E>(futures: Future<T, E>[]) -> Future<T[], E>
//   all values in input order; on first failure the rest are CANCELLED, then error re-thrown (JS Promise.all)
fn race<T, E>(futures: Future<T, E>[]) -> Future<T, E>
//   first to settle (success OR failure) wins; losers cancelled (JS Promise.race); empty array never settles
fn any<T, Err>(futures: Future<T, Err>[]) -> Future<T, AllFailed<Err>>
//   first to SUCCEED wins; losers cancelled; if all fail, throws AllFailed<Err> (JS Promise.any)

class AllFailed<E> { errors E[] }   // thrown by `any` when every input failed; errors in input order
```

### `baml.spawn` — spawn options & `with` middleware

```baml
// Every `spawn` implicitly builds a SpawnParams; the `with` pipeline transforms it.
// A transformer is any function (SpawnParams<T,E>) -> SpawnParams<U,F>.
class SpawnParams<T, E> {
  body   () -> T throws E
  name   string?
  group  TaskGroup?
  cancel CancelToken?
  detach bool
}

// The built-in config transformer. `cancel` links a token; `detach = true` opts out of the
// parent→child cancel cascade and routes unhandled errors to the root task; `group` enrolls
// the spawn in a TaskGroup (parks FIFO until admitted). Leaves `body`/types unchanged.
fn options<T, E>(group: TaskGroup? = null, cancel: CancelToken? = null, detach: bool? = null)
  -> (SpawnParams<T, E>) -> SpawnParams<T, E> throws never

class TaskGroup {                                  // caps concurrency; excess spawns queue FIFO
  fn new(limit: int, name: string? = null) -> TaskGroup throws never
  fn cancel(self, pending: bool? = true, active: bool? = true) -> int throws never  // returns # signalled
  fn set_limit(self, limit: int) -> void throws never   // raise admits queued; lower never preempts; 0 pauses
  fn limit(self) -> int throws never
  fn name(self) -> string? throws never
  fn active_count(self) -> int throws never
  fn queued_count(self) -> int throws never
}

class CancelToken {                                // cooperative, one-shot; fires at next await
  fn new() -> CancelToken throws never
  fn any(tokens: CancelToken[]) -> CancelToken throws never   // fires when ANY input fires (one-directional)
  fn cancel(self) -> int throws never              // 1 if this call transitioned it, else 0
  fn is_cancelled(self) -> bool throws never
}
```

A cancelled task throws `baml.panics.Cancelled` at its next `await`.

### Writing your own middleware

A transformer is just a function returning `(SpawnParams<T,E>) -> SpawnParams<U,F>`. It can set config
fields, wrap `body` (retry/timing), or replace it entirely (a type-changing transformer — e.g. a
fallback that catches and erases the error type to `null`):

```baml
function withFallback<T, E>(default_value: T) -> (SpawnParams<T, E>) -> SpawnParams<T, null> throws never {
  (params) -> {
    let original = params.body;
    SpawnParams {
      body: () -> { (original()) catch (e) { let e => default_value } },
      name: params.name, group: params.group, cancel: params.cancel, detach: params.detach,
    }
  }
}

// usage — `f : Future<int, null>`, so awaiting needs no catch:
let f = spawn with withFallback(99) { flaky() };
```

Transformers run **eagerly at the spawn site**, so a throwing transformer throws there (and is part of
the caller's throws surface). Wrong-shaped transformers are rejected at compile time with a concrete
"must return a SpawnParams" diagnostic.

## Spawn options
- `spawn ... with baml.spawn.options(group?, cancel?, detach?)` — config-only `with` clause
- `CancelToken` (`new` / `any` / `cancel` / `is_cancelled`) + parent→child cancel cascade
- `TaskGroup` rate limiting (FIFO admission, exact counts)
- `detach` (fresh cancel token; unhandled errors routed to the root task)
- strict `with` validation (must be a `baml.spawn.SpawnConfig`)

## Future combinators (`baml.future`)
- `race` / `any` / `all` / `all_complete` — pure-BAML stdlib over a new 1-byte `AwaitAny` engine primitive (`__await_any: Future<T,E>[] -> int throws never`, suspends until the first input settles)
- `AllFailed<E>` aggregate error class for `any`
- losers cancelled (`race`/`any`/`all`) or left running (`all_complete`) per the BEP table

## Engine/runtime fixes found along the way
- **fire-and-forget errors (bug 2):** a child's unhandled throw is now *deferred* (GC-rooted stash; future stays Pending with its wake signal fired) and consumed at the single observation point (`future_ready`, now `&mut`). The spawner's post-await drain surfaces only never-observed errors — a sibling handling a child's error no longer races the spawner into a spurious `UnhandledThrow`. (`tests/fire_and_forget.rs`)
- **AwaitAny index desync (bug 3):** global-index bookkeeping fix that broke `any`
- **bare-bind catch arms (bug 4):** `let e =>` in a generic fn emitted a constant-false `IsType` (TypeVar→Void erasure) and silently rethrew; bare binds are irrefutable — no runtime test
- **copy propagation** substitutes `Terminator::Await` operands (`await <param>` inside `.map` closures)
- unannotated lambda params infer from generic array elements mentioning only the enclosing fn's rigid generics

## Generics: inferred type args at runtime
- TIR records call-site inferred bindings (`call_type_instantiations`); MIR threads them to `frame.type_args` (`LoadType`/`TypeArgRef`) — instances built inside generic fns carry real `class_type_args`, and typed arms like `let e: AllFailed<string>` match (`tests/generics_runtime.rs`)
- typed arms naming enclosing TypeVars (`let e: AllFailed<E>`) lower via type-arg templates instead of erasing

## Canary merge (f5df38abf)
Absorbs the spawn/GC permit deadlock fix (full `baml_tests` now completes on this branch), BEP-044/BEP-57, instantiation expressions, bounds, and interface-dispatch type-arg threading. Canary's strict rigid-var checking exposed pre-existing inference gaps, fixed here:
- `await f` contributes the future's `E` to escaping throws
- callee generics matched only against rigid caller TypeVars now bind (`map`'s `U:=T`)
- lambda *effective* throws feed callee throws-generics — lifts the documented "current limitation" (covered callback throws no longer false-positive) and violations name concrete types
- **miscompile fix:** `ShortCircuit` emit assumed a stack-carried (PhiLike) destination; with `LoadType` before call args the carry is validly rejected, so `a || b` as an inferred-generic call argument read an uninitialized local. Emit now materializes the destination on the short-circuit edge (no-op for PhiLike).
- std `any<T, E>` → `any<T, Err>` (works around generic-param name capture vs `map<U, E>`; proper fix tracked separately)

## `with` middleware (c37027aa7)
Full BEP middleware: `spawn ... with a, b { body }` builds a `SpawnParams<T,E>` and applies each transformer left-to-right — any plain BAML function `(SpawnParams<T,E>) -> SpawnParams<U,F>` (wrap/replace body, set config, change types). `baml.spawn.options` is now a pure-BAML transformer (native SpawnConfig deleted). Flushed out + fixed three language gaps: keyword segments in type paths, lambdas as block tail expressions, and unannotated lambdas now INFER their throws surface from the body (two tests asserting the old throws-never default updated). Tests: `bex_engine/tests/middleware.rs` incl. the BEP's `withRetry`.

## Complete bug ledger (every bug found & fixed on this branch)

**Engine / runtime**
1. **Fire-and-forget errors bypassed consumers** — a spawned child's unhandled throw settled its Future eagerly AND queued on the spawner; a sibling that awaited + handled the error was invisible to the engine, so the spawner re-surfaced an already-handled error as `UnhandledThrow`. Fixed with the deferred-error model: the error parks in a GC-rooted stash, the future stays Pending (wake signal fired), `future_ready` is the single observe-and-consume point, and the spawner's post-await drain surfaces only never-observed errors.
2. **`AwaitAny` index desync** — global-index bookkeeping desync broke `any` (wrong winner index reported).
3. **Copy propagation didn't substitute `Terminator::Await` operands** — `await <param>` inside a `.map` closure read an uninitialized slot.
4. **`ShortCircuit` emit miscompile** — the emitter assumed the destination is always stack-carried (PhiLike); when inferred-generic calls push `LoadType` before args the carry is validly rejected, so `a || b` as a call argument left the value on the stack and read an uninitialized local (`true || false` evaluated false). Emit now materializes the destination on the short-circuit edge; MIR normalizes projection destinations. *Latent before the canary merge — the suite deadlocked before reaching the `||` fixtures.*

**Type system / generics**
5. **Bare-bind catch arms in generic fns always rethrew** — `let e =>` emitted an `IsType` from its inferred type; a rigid TypeVar erases to `Ty::Void` in MIR making the test constant-false. Bare binds are irrefutable — no runtime test emitted (also drops tautological field tests in destructures).
6. **Inferred call-site type args were discarded** — TIR ran inference, substituted the return type, and threw the bindings away; MIR emitted frame type args only for explicit `<...>` call sites. Instances built inside generic fns carried empty `class_type_args`; typed arms like `let e: AllFailed<string>` never matched. Bindings now recorded (`call_type_instantiations`) and threaded to `frame.type_args`.
7. **Typed arms naming enclosing TypeVars compiled to constant-false** — `let e: AllFailed<E>` inside `any<T, E>` erased E→Void; now lowers via `TypeArgRef` templates resolved against the frame.
8. **Recorded instantiations poisoned by fresh literal types** — `mk("hi")` recorded `T = Literal("hi")` so `is Box<string>` compared unequal; recordings now `widen_fresh()`. Plus generic→generic chains recorded `Unknown` (TypeVar actuals refused); fixed with an allow-typevars backfill.
9. **Unannotated lambda params didn't infer from generic array elements** whose type only mentions the enclosing fn's rigid generics.

**Throws tracking** *(pre-existing gaps exposed by canary's strict rigid-var checking)*
10. **`await f` contributed nothing to escaping throws** — the future's `E` is now part of the awaiting body's throw set.
11. **Callee generics matched only against rigid caller TypeVars never bound** — `map`'s `U:=T` in `futures.map((f) -> { await f })` was skipped as a TypeVar actual, leaving `U[]` in result types.
12. **Rigid TypeVar throw-facts collapsed lambda surfaces to `Never`** — only genuinely open facts (fresh effect slots, Unknown) collapse now.
13. **Callee throws-generics fed from lambda SURFACE instead of effective throws** — an omitted-throws lambda's open surface left `map`'s `E` unbound (leaking a raw TypeVar) or symbolic where the body demonstrably throws. Now fed from the side-table effective set: a non-throwing callback binds `E:=never`, a throwing one binds its concrete type. **Lifts a documented "current limitation"**: a callback throw covered by the caller's declared throws no longer false-positives, and violations name concrete types instead of symbolic `E`.

**Parser / lowering** *(flushed out by the middleware work)*
14. **`spawn`/`await` keywords rejected as path segments** in type annotations (`baml.spawn.SpawnParams`) — parser type-path loop + CST dotted-name extractor.
15. **A lambda as a block's tail expression was silently dropped** from block elements — the block typed as void ("missing return value"); any function returning a transformer hit it (the std's `json` code had been dodging it with explicit `return`).
16. **Unannotated lambdas defaulted to `throws never` with a local violation** — they now infer their throws surface from the body (required for body wraps like `() -> { original() }` whose throws is a generic `E`). Omitted throws on callback *params* remains effect-polymorphic; explicit `throws` contracts remain enforced (verified: E0096 fires through the forwarding chain — an enforcement that itself depends on fixes 10–13).

**Known issues (documented, not fixed here)**
- Generic-param name capture: a callee's generic (`map<U, E>`) sharing a name with a caller's generic false-binds through the receiver-substituted signature. Worked around (`any<T, E>` → `any<T, Err>`); proper fix is alpha-renaming on receiver substitution.
- `await` does not yet distribute over unions of futures (BEP §await-distributes).

## `await` over unions (94d2b463a)
Awaiting `Future<A, E1> | Future<B, E2>` types as `A | B` with both error types tracked (BEP §await-distributes). Tests: `bex_engine/tests/await_union.rs`.

## Review hardening (94d2b463a)
CodeRabbit batch: `any` errors in INPUT order (+test) · explicit lambda throws contracts preserved · `with`-transformer throws contributed at spawn · copy-prop protects `await xs[_i]` · `TaskGroup.set_limit(+N)` admits up to the cap (+test) · deferred-error hygiene on cancel/fulfill · VM Spawn config validation · unified/completed generics walkers. Plus the CI fix: `Object` enum exceeded its 64-byte budget under `heap_debug` (`UnscheduledFuture` boxed) — the cause of the all-legs CI failure.

## Canary merge round 4 (8b2c40380)
Canary's `c957417e7` ([codex] #3702) was a third independent implementation of inferred-generic runtime type args; this merge converges on it and fixes what the convergence broke:

1. **Adopted canary's CallPlan pipeline for inferred type args** — removed our parallel `call_type_instantiations` TIR→MIR fallback (dead code after the switch). Canary's ShortCircuit emit fix (`store_on_taken_path`) is the same insight as ours; took theirs.
2. **`runtime_call_type_args` lost literal widening** — canary's recording threads `T = literal "hi"` for `mk("hi")`, so an escaped instance never matches `is Box<string>` (runtime class-arg comparison is invariant). Restored `widen_fresh` on inferred bindings.
3. **Canary's stricter call-result recheck false-positived on `unknown` holes** — `spawn with` middleware checks transformers against `(SpawnParams<T,E>) -> SpawnParams<unknown, unknown>`; the normalizer implements class-arg invariance as structural equality, so the holes (and `BuiltinUnknown` from erased unresolved typevars, e.g. a variable-bound transformer; and equivalent spellings like `T = int | 99` from fallback-arg joins) all failed E0001. `is_subtype` now compares same-class args pairwise: holes match anything, concrete args must be **mutual** subtypes (invariance as equivalence). Concrete mismatches — including canary's new iterator-stdlib diagnostics — still report.
4. **`heap_debug` feature gap** — `FutureRead::ErrorPending` arm added to the heap debugger's validity walk.
5. **`task_group_cancel_cancels_members` timing** — the promptness bound times `run_main` *including compilation*; the stdlib grew (ns_iter). 2s → 5s, still far under the 10s sleep it guards.

Closes the deferred interface-dispatch inferred-type-args item (canary's #3702 ships it with runtime regressions).

Verified: `cargo test --workspace --all-features` — 185 suites / 0 failures (sdk_test_* excluded as in CI; they require cargo-nextest setup scripts) + prek clean, rustc 1.93.0.

## Deferred (follow-up branch)
- `Result<T, E>` type design + `all_settled`, `.then()`
- cancellation-test panic-hook hygiene

Tests: `bex_engine` 227/0 (incl. `combinators`, `fire_and_forget`, `generics_runtime`, `cancellation`, `task_group`); `baml_tests` diagnostics 418/418, runtime fixtures 1650/1650 (one timing-sensitive fixture flakes only under full parallel suite load); snapshots regenerated.

## Summary by CodeRabbit

* **New Features**
  * spawn ... with middleware and options(...) for spawn configuration; CancelToken and TaskGroup APIs; future combinators all_complete, all, race, any; await-any (returns winning input index)
* **Tests**
  * Comprehensive BEP-034 tests for spawn options, TaskGroup, CancelToken, combinators, middleware, fire-and-forget, and runtime behavior
* **Bug Fixes**
  * Improved child-task error propagation and deferred fire-and-forget error semantics across await/catch boundaries


antoniosarosi:
@coderabbitai Review disposition for the latest push (94d2b463a):

**Fixed (9):** `any` input-order errors (+ regression test) · explicit lambda throws surfaces preserved in `callee_throws_arg_ty` · spawn `with` transformer throws contributed at the spawn site · forward-ref check no longer descends into the deferred spawn body · `spawn`/`await` segments in generic-arg lookahead · `contains_non_rigid_typevar` unified over `contains_typevar_where` · `erase_typevars_where` completed (Interface/projection/Evolving/Function-shadowing) · deferred fire-and-forget errors cleared on cancel/fulfill transitions · VM validates the Spawn config operand is an Instance · copy-prop protects `await xs[_i]` index locals · `TaskGroup::promote` admits up to the new cap (+ regression test).

**False positive (1):** AwaitAny empty-waiters park — `future_ready` returns immediately-ready waiters for settled inputs, so empty waiters ⟺ empty input array, and BEP-034 specifies `race([])` never settles (matching JS `Promise.race([])`). Comment expanded to document this.

**Deferred (2, tracked):** interface-dispatch inferred type args (`emit_method_candidate_switch`) and the cancellation-test panic-hook hygiene.

PR #3635, opened 2026-06-02
Fix baml_tests hang: O(files²) compile blow-up + spawn/GC permit deadlock

hellovai:
## TL;DR

`cargo test -p baml_tests` (and `baml-cli test --from baml_src`) **hung**. It turned out to be **two unrelated bugs stacked on top of each other**, and the first one masked the second:

1. **Compile-time `O(files²)` blow-up** — the deterministic hang. Compiling the 55-file `baml_src` corpus as one project ran two whole-project passes once *per file* / *per function*. Fixed with salsa memoization. Corpus went from *never finishing* → ~30s, **byte-identical bytecode**.
2. **A runtime GC/permit deadlock in `spawn`** — once compilation was fast enough to actually reach execution, the test run started hanging ~50% of the time in the engine. Root cause was a **nested heap-permit acquire that deadlocks against stop-the-world GC**. Fixed by allocating the spawned future under the parent's existing permit.

Both were diagnosed by measurement (CPU sampling, scaling curves, forced-GC stress repro), not guesswork. Details + code references below so this is auditable.

**Commits**
1. `perf(compiler2): …` — memoize the two whole-project compile passes (Problem 1).
2. `fix(engine): …` — allocate the spawned future under the parent's permit (Problem 2).
3. `perf(mir): …` — follow-up: borrow `resolved_aliases` instead of cloning it per context (see *Follow-up* section), addressing review feedback.

---

### Symptom

`baml-cli test --from baml_src` pinned one core at ~95% CPU and never finished on the full corpus. Not a cargo-lock deadlock, not network — pure CPU-bound compilation. Scaling was clearly super-linear:

| files | time (before) |
|------:|--------------:|
| 1  | 4s |
| 15 | 13s |
| 27 | 31s |
| 35 | 44s |
| 55 | **never finished** (~100s+ extrapolated) |

`sample`-ing the process showed all time under `collect_diagnostics` → type inference → `ppir_expansion_items` → `collect_alias_bodies` → `lower_file`, and later under `generate_project_bytecode` → `lower_function` → `LoweringContext::new` → `populate_from_package`.

### Two quadratics

**(a) `baml_compiler2_ppir`** — `ppir_expansion_items` is a **per-file** `#[salsa::tracked]` query (`lib.rs:205`), but each invocation called `collect_block_attrs` / `collect_alias_bodies` — plain functions that iterate **every** file in the project and call `ast::lower_file` on each. So N files × re-lowering N files = **O(N²)** lowerings.

**(b) `baml_compiler2_mir`** — `LoweringContext::new` / `new_for_let` run **per function**, and each rebuilt `populate_from_package` (which lowers every class field type across all packages, `lower.rs:1190`) plus `ResolvedAliases::for_package` (which re-runs `find_recursive_aliases` over the whole project). M functions × N classes = **O(N²)** again.

### Fix

Memoize the whole-project work behind package/project-keyed salsa queries, reusing the manual `unsafe impl salsa::Update` (via `PartialEq`) pattern already established for `PackageItems`:

- `project_expansion_maps(db, project)` — `ppir/lib.rs:165`
- `package_lowering_data(db, pkg_id)` — `mir/lower.rs:957`. `LoweringContext` now **borrows** the schema maps and `resolved_aliases` (`&'db`) instead of rebuilding/cloning them per function. (The `resolved_aliases` borrow was completed in commit 3 — see *Follow-up*.)

### Result

| files | before | after |
|------:|-------:|------:|
| 35 | 44s | 20s |
| 55 | never finished | ~30s |

Scaling is now ~linear. **The bytecode snapshot test passes unchanged** — output is byte-for-byte identical, so this is a pure performance change. 1576 `baml_tests` lib tests pass.

---

## Problem 2 — `spawn` deadlocks against GC (commit `fix(engine): …`)

This is the subtle one. Once compilation was fast, the full 1614-test run started **hanging ~50% of the time**, always with the *same* shape: the tokio runtime driver parked in `block_on` and **every worker thread idle/parked** — i.e. a lost-wakeup, not a CPU spin.

### The BAML test that triggered it

`crates/baml_tests/baml_src/ns_cancel_cascade/cancel_cascade.baml`:

```baml
function cancelled_child_future_state_is_cancelled() -> baml.future.FutureState {
    let slow   = spawn { baml.sys.sleep(60000); 42 };  // task S: sleeps 60s
    let waiter = spawn { await slow };                  // task W: awaits S
    let _      = waiter.cancel();                        // cancel W (not S)
    waiter.state()
}
```

It passed **5/5 in isolation** but hung ~50% in the full run — the classic fingerprint of a *concurrency* bug that needs accumulated load, not a logic bug in the test. Two facts narrowed it down:

- All threads parked ⇒ a lost wakeup in async machinery, not a synchronous lock.
- It correlated with **garbage collection**: forcing GC to run on *every* allocation (temporarily setting the Gen0 threshold `10_000 → 1` in `bex_heap`) turned the flake into a **100% reproducible** hang on a tiny repro. That was the key to pinning it.

### Background: the heap-permit model

The engine coordinates GC with a `HeapPermitManager` backed by a **single tokio `Semaphore`** (`bex_heap/src/heap_guard.rs`):

- Each running VM mutator holds **one** `ActiveHeapPermit` (one semaphore permit).
- Stop-the-world GC parks everything by draining the **entire** semaphore at once:

```rust
// bex_heap/src/heap_guard.rs:227
pub async fn request_park(&self) -> HeapGuard<'_> {
    let permits = self.active
        .acquire_many(MAX_PERMITS)   // <-- wants ALL permits; completes only when
        .await                        //     every ActiveHeapPermit has been released
        ...
}
```

The crucial property: **tokio's `Semaphore` is fair (FIFO)**. Once `acquire_many(MAX_PERMITS)` is queued, any later `acquire()` (even for 1 permit) queues **behind** it and cannot be granted until the big request is satisfied and released.

### The bug

`spawn` allocated the child's heap `Future` by taking a **second, fresh** permit *while the parent task that issued the `spawn` still held its own permit*. The parent awaits `spawn_thread` inline, so both permits live on the same logical flow:

```rust
// OLD — bex_engine/src/lib.rs, spawn_thread_setup (deleted in this PR)
let permit = self.heap_permit_manager.new_permit(()).await;
let permit = permit.acquire().await;          // <-- 2nd permit, while parent still holds its 1st
let (future_id, future_ptr) = {
    let mut guard = self.futures.acquire(permit.proof()).await;
    guard.new_future(child_cancel.clone())     // allocate the child Future
};
drop(permit);
```

Now interleave a GC park (which, under real workloads, fires whenever heap pressure crosses the threshold — hence the flakiness, and 100% under forced GC):

```mermaid
sequenceDiagram
    participant P as Parent task<br/>(holds permit P_main)
    participant G as GC (request_park)
    participant S as Semaphore (fair FIFO)

    P->>P: executing spawn { ... }
    Note over G,S: heap pressure → GC starts
    G->>S: acquire_many(MAX_PERMITS)
    S-->>G: queued — waits for ALL permits<br/>(P_main still held by Parent)
    P->>S: acquire() for the child-future permit
    S-->>P: queued BEHIND GC (fair) — blocked
    Note over P,G: 🔒 deadlock cycle
    Note right of P: Parent won't release P_main<br/>until spawn returns
    Note right of P: spawn can't return<br/>until it gets the 2nd permit
    Note right of G: GC can't grant the 2nd permit<br/>until it finishes, which needs P_main
```

So: **GC waits for the parent's permit → the parent waits (fairly, behind GC) for a second permit → the parent won't release the first until it gets the second.** Cycle. All tasks suspend; every worker parks. The 60s `sleep` is a red herring — the hang is immediate (it reproduced with `PASS=0`).

### The fix

There is no reason to take a *new* permit to allocate the child future: the parent is **already** holding an active permit at the `spawn` site. Allocate the future there, under the parent's permit, and hand the `future_id` to `spawn_thread`:

```rust
// NEW — bex_engine/src/lib.rs, VmExecState::Spawn dispatch (~lib.rs:2462)
let child_cancel = cancel.child_token();
let future_ptr = {
    let mut guard = self.futures.acquire(thread.proof()).await;   // parent's permit
    let (future_id, future_ptr) = guard.new_future(child_cancel.clone());
    drop(guard);                                                  // dropped before the await below
    Arc::clone(self)
        .spawn_thread(child_cancel, parent_errors_arc, closure, spawn_name, call_id, future_id)
        .await?;
    future_ptr
};
thread.vm.stack.push(Value::object(future_ptr));
```

`spawn_thread` now only builds the child VM and **registers** the child's permit via `new_permit` (which takes the holders mutex but does **not** acquire a semaphore permit), then fires the task. The child's permit is acquired later, on the spawned task — never nested under the parent. **One permit per task on the spawn path**, so the deadlock cycle cannot form. `spawn_thread_setup` is deleted.

Why this is safe to audit:
- `new_future` is synchronous and only needs a `PermitProof` to prove GC isn't running — the parent's `thread.proof()` satisfies that just as well as a fresh permit did.
- The `FutureManagerGuard` is dropped **before** the `spawn_thread().await`, so no non-`Send` guard crosses a yield point (same pattern already used elsewhere in `run_thread_event_loop`).
- `new_permit` only contends on the holders mutex, and `request_park` is explicitly ordered (semaphore first, then holders mutex) to not deadlock against it.
- The parent holds its permit across the whole dispatch, so GC cannot move `future_ptr` during the await.

### Verification

The bug is flaky, so I verified with the forced-GC stress harness (turns the race into a deterministic signal) and then with the real threshold:

| scenario | before fix | after fix |
|---|---|---|
| minimal `spawn/cancel` repro, **GC forced every alloc** | 6/6 hang | **0/10 hang** |
| full 1614-test corpus, real GC threshold | ~50% hang | **0/8 hang** (1614 passed, 0 failed each) |

(The temporary `gc.rs` threshold change was only a debugging aid and is **not** part of this PR.)

---

## Follow-up — borrow `resolved_aliases`, and what's *not* worth optimizing (commit `perf(mir): …`)

Review feedback flagged two remaining per-`LoweringContext` (i.e. per-function) costs. I profiled a full `--list` compile of the corpus before touching either, and the result decided each:

> The whole MIR lowering path (`package_lowering_data` / `LoweringContext::new` / `lower_function`) is **~11 of ~3600 samples**. Compile time is dominated by TIR `infer_scope_types` (`render_scope_diagnostics → infer_scope_types`). So neither of these is a measurable cost on the current corpus.

**Fixed: `resolved_aliases` cloned per context → borrowed.** The other five package-invariant schema maps were already borrowed (`&'db`) from `package_lowering_data`; `resolved_aliases` (a `HashMap` + `HashSet`) was the one I'd left as a per-context clone for expedience. Now it's borrowed too. Whole-struct passes (`&self.resolved_aliases`) became `self.resolved_aliases` (the field is already a reference); `.aliases` sub-field accesses and `.convert(...)` calls are unchanged (they auto-deref through the borrow). Not a measurable speedup here, but it's a strict reduction in per-context allocation, completes the borrow-not-clone design, and is asymptotically `O(contexts × aliases)` for alias-heavy projects.

**Skipped (measured non-issue): making `build_class_type_tags` a tracked query.** It showed up as **0 samples**. Unlike `populate_from_package`, it does no type lowering — just cached `file_item_tree` reads and `TypeName→i64` inserts. Memoizing it would add a wrapper + `Update` impl + a project-keyed query, and it's **bytecode-affecting** (it assigns the global type-tag numbering that must match the emitter), so it's risk for no measurable benefit. Worth revisiting only if a profile on a larger/real project shows it mattering.

---

## Test status

- `bex_engine`, `bex_vm` unit tests — pass
- `cancel_cascade`, `spawn_array_race`, `spawn_parallel`, `spawn_semantics`, `spawn_specialization` integration tests — pass
- `baml_tests --lib` (1576 tests) — pass
- bytecode snapshot (`baml_tests --test baml_src`) — pass, **unchanged**
- `cargo fmt` + `cargo clippy -D warnings` — clean (enforced by pre-commit)

## Reviewer notes / blast radius

The engine change is on the **hot `spawn` path** and touches core GC/permit concurrency, so it deserves a careful read despite the small diff. The repro is flaky by nature; I'd suggest a CI loop running the corpus a handful of times to build confidence. The compiler change is performance-only and guarded by the byte-identical bytecode snapshot.

## Summary by CodeRabbit

* **Refactor**
  * Memoized package-level schema and type-alias data to eliminate redundant per-function recomputation and speed up compilation.
  * Centralized project-wide expansion maps for consistent, more efficient expansion across files.
  * Adjusted spawn/allocation flow so child futures are allocated at spawn sites, improving heap-permit handling and runtime stability.

* **Chores**
  * Updated size-gate thresholds and CI artifact size metadata.


hellovai:
Verified both findings against the current code and profiled before changing anything (sampled a full `--list` compile of the 55-file corpus).

**1. `resolved_aliases` cloned per `LoweringContext` → fixed** (`afc6ee2`).
The other five package-invariant schema maps were already borrowed (`&'db`) from `package_lowering_data`; `resolved_aliases` was the one I'd left as a per-context clone for expedience. Now borrowed too. Whole-struct passes (`&self.resolved_aliases`) became `self.resolved_aliases` (the field is already a reference); `.aliases` / `.convert(...)` sites are unchanged (auto-deref). Bytecode snapshot byte-identical, 1576 lib tests pass, full corpus passes.

Honesty note: this isn't a measurable speedup on the current corpus — profiling shows the whole MIR lowering path (`package_lowering_data` / `LoweringContext::new` / `lower_function`) is **~11 of ~3600 samples**; compile time is dominated by TIR `infer_scope_types`. I took it anyway because it's a strict reduction in per-context allocation, completes the borrow-not-clone design, and is asymptotically O(contexts × aliases) for alias-heavy projects.

**2. `build_class_type_tags` per `LoweringContext` → skipped (measured non-issue).**
It showed up as **0 samples** in the profile. Unlike `populate_from_package`, it does no type lowering — just cached `file_item_tree` reads and `TypeName→i64` map inserts. Making it a tracked query would add a wrapper + `Update` impl + a project-keyed query, and it's **bytecode-affecting** (it assigns the global type-tag numbering that must match the emitter), so it carries real risk for no measurable benefit. Happy to revisit if a profile on a larger/real project shows it mattering.

PR #3645, opened 2026-06-02
Deslopify TIR: A workflow experiment

antoniosarosi:
## What

An experiment in using **multi-agent workflows** to shrink `baml_compiler2_tir` while keeping behaviour identical — first two *mechanical* de-slopify passes, then a *semantic* pass that built a verified model of the crate and executed major refactors against it.

**Result: `baml_compiler2_tir` shrank 30,609 → 26,698 lines (−3,911, −12.8%)** with **no tests removed**, **131/131 crate + 4,991/4,991 workspace tests passing**, the full ~1,234-snapshot insta suite intact, and the workspace `cargo clippy --all-targets --all-features -- -D warnings` clean.

## Stage 1 — Mechanical passes (`8fb72ec6`, `b9c5d7c0`): −2,415 lines

A pipelined workflow (per-file → call-graph groups → whole-crate, each analyze→fix, gated by compile + tests after every wave) hunting duplication, dead code, redundant helpers, comment rot, needless abstraction, dead surface, and magic strings. Big files (`builder.rs`, ~13k lines) got section-mapped deep analysis so nothing was skimmed. Safety rails: workspace-wide usage check before any deletion (salsa/trait dispatch aware), test-integrity gates that restore any wrongly-deleted test from git.

Highlights: unified duplicate throws walkers, `#[derive(salsa::Update)]` over hand-written unsafe impls, deleted dead modules/variants/params (cycle_detector, 4 never-constructed `TirTypeError` variants, dead pub params) with cross-crate call-site updates.

## Stage 2 — Semantic pass (16 commits, `3a420db2..fafd96de`): −1,496 lines

A two-phase pipeline:

1. **Survey (read-only):** ~47 agents built a semantic model of the crate — per-file/per-type-cluster models, the data-flow map, and a **189-claim invariant register adversarially verified against the code**, distilled into 77 frozen behavioral invariants. Six lenses then proposed 29 major refactors, judged on value/safety/tractability; one was rejected outright when a judge found a concrete counterexample in the code.
2. **Execute (serial):** each approved refactor ran as plan → implement → adversarial verify → checkpoint-commit-or-revert. The gate per refactor: compile + unit tests + the **full insta snapshot suite with zero drift** + workspace clippy + a hostile diff review against the frozen invariants. **16/16 landed, 0 reverted.**

Snapshot policy: byte-identical snapshots, with one reviewed relaxation — purely *wording* changes to diagnostics (same kind, same span, same count) could be accepted per-snapshot by the verifier. Exactly **7 snapshots** changed across the whole run, all wording-only, all on the final commit (collapsing the two drifted duplicate diagnostic-message channels into one).

Headline refactors: deleted the provably-always-default `TyAttr` payload from all 23 `Ty` variants (−708 alone); `body: &ExprBody` un-threaded from 63 methods/209 call sites into builder state; the per-body inference map set (duplicated across 4 structs + a 14-tuple) collapsed into one `InferenceMaps`; `DPat`/`WitnessPat` unified; single-use dataflow framework inlined; duplicate Tarjan SCC impls unified; diagnostics routed through a sink.

## Why this is trustworthy

- Every refactor is an isolated checkpoint commit, individually verified before the next built on it.
- The insta suite acts as a behavioral oracle: agents were forbidden from accepting snapshots; only the adversarial verifier could, per-snapshot, under the wording-only rule.
- The invariant register (`model.md` / `invariants.md`, kept outside the repo) was verified by falsification — several "invariants" turned out to be corrected/false claims and are documented as bugs-to-preserve rather than silently "fixed".

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic messages for type errors and missing-return cases with clearer, more consistent wording.
  * Enhanced type inference error reporting with better formatting.

* **Documentation**
  * Added comprehensive documentation for TIR compiler invariants and semantic model.


aaronvg:
i wonder if we can get better reesults if we run `/simplify` in claude as well.

And also if we formally design how subtyping rules should be modeled across exhaustiveness checks, match, variable assignments, implicit casts etc.

PR #3692, opened 2026-06-04
TOML stdlib parser + ZonedDateTime/Plain* date-time family

antoniosarosi:
Lands two tightly-coupled pieces in one PR, per [BEP-021 v3](https://beps.boundaryml.com/beps/21): the `baml.toml` parser needs the new time types for lossless datetime interop, so they ship together.

## 1. `baml.time` date/time family (BEP-021, Slack-decided design)

**New stdlib types** (`ns_time/`): `ZonedDateTime`, `PlainDateTime`, `PlainDate`, `PlainTime`, `TimeZoneOffset`, plus `system_timezone()`. TC39 Temporal taxonomy and naming: `Plain*` = anything without a timezone.

- `ZonedDateTime` = absolute instant (`bigint` ns) + fixed offset **or** IANA identifier. Parses RFC 3339 and RFC 9557 (`2026-03-18T13:04:27-07:00[America/Los_Angeles]`); serializes the way Temporal does. Zoneless strings are rejected — they belong to `PlainDateTime.parse`.
- Civil types use the BEP's compact encodings (`PlainDateTime`: civil ns-as-if-UTC `bigint`; `PlainDate`: epoch days; `PlainTime`: ns since midnight).
- `PlainDateTime.to_zoned(tz, disambiguation)` implements TC39 DST semantics (`compatible`/`earlier`/`later`/`reject`); `reject` is built on earlier/later agreement in the BAML layer, so the io surface stays minimal.

**Timezone database = host io, per the Slack decision:**
- **Native** (`sys_native`): resolves IANA identifiers against the host's zoneinfo via `jiff` — Deno-style, no bundled tzdata except on platforms with no host db (Windows).
- **WASM** (`bridge_wasm`): delegates per-operation to host JS `Temporal` (no caching). Hosts without `Temporal` get `Unsupported`.
- Only IANA resolution touches the db; fixed-offset and `Plain*` operations are pure (parse/format/calendar math in `bex_vm` natives via the `time` crate).

**Deferred:** operator overloading (`ZonedDateTime + Duration`, …) pending BEP-021 Open Question 5.

## 2. `baml.toml` parser (continues @2kai2kai2's WIP)

`baml.toml.Table.parse` (native, `toml` crate) + BAML-level `to_json`/`from_json`/`item_to_json`/`item_from_json`. TOML's four datetime kinds map losslessly per the BEP interop table, replacing the string-backed `Datetime` placeholder:

| TOML kind | BAML type |
| --- | --- |
| Offset Date-Time `1979-05-27T07:32:00-07:00` | `ZonedDateTime` (fixed offset) |
| Local Date-Time `1979-05-27T07:32:00` | `PlainDateTime` |
| Local Date `1979-05-27` | `PlainDate` |
| Local Time `07:32:00` | `PlainTime` |

## Plumbing

- `AsBexExternalValue for Arc<BigInt>` so io fns can return `bigint`.
- Generated-code allows for 10-arg `from_components` and `__glue__format_zoned`.
- Fixes pre-existing clippy lints across the workspace (newer-clippy debt) so `cargo clippy --workspace --all-targets --all-features -- -D warnings` and the pre-commit hook are green again.

## Tests

- `baml_tests/tests/datetime.rs`: 22 tests — parse/format round-trips, components, pre-epoch civil math, fixed-offset `to_zoned`, IANA offset resolution (DST-aware), the canonical 2011 `America/Los_Angeles` gap/overlap disambiguation cases, `UnknownTimezoneError`/`AmbiguousTimeError`, and all four TOML kinds in one document.
- `baml_tests/tests/toml.rs`: updated for the new mapping (8 tests, 1 pre-existing ignore blocked on the VM `continue`/`break` bug).
- Full `baml_tests` suite: 1576 passed / 0 failed. Stdlib snapshots refreshed.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive date & time types (PlainDate, PlainTime, PlainDateTime, ZonedDateTime) with parsing, formatting, component accessors, conversions, and JSON serialization.
  * Timezone support: fixed UTC offsets and IANA timezones with DST-aware disambiguation and host/WASM timezone resolution.
  * TOML parsing and conversion to JSON, including datetime mappings.

* **Tests**
  * Added extensive runtime tests covering date/time behavior, timezone edge cases, and TOML handling.

## Bugs found & fixed along the way

1. **Workspace clippy debt (fixed).** A newer local clippy flags ~12 pre-existing sites across `baml_compiler2_{hir,tir,mir,emit}`, `bex_sap`, `bex_vm`, `bex_project`, `tools_sap_visualizer`, and the `baml` wrapper bin (unused vars, trailing commas, collapsible matches, `for_kv_map`, `unnecessary_sort_by`, `duration_suboptimal_units`, an f32-literal future-compat warning). All fixed, so `cargo clippy --workspace --all-targets --all-features -- -D warnings` and the pre-commit hook are green again.
2. **`AsBexExternalValue` missing for `Arc<BigInt>` (fixed).** io functions could take `bigint` arguments but not return `bigint` — the return conversion impl simply didn't exist. Added in `bex_external_types`; the tz io helpers need it.
3. **`baml_cli describe` listing + bytecode snapshots (refreshed).** Stale against the new stdlib surface and canary's `T | null` lowering — snapshot updates, not code bugs.

## Pre-existing bugs found, worked around, NOT fixed here

4. **VM test-block local boxing bug** (known; see the `ns_closures/closures.baml` header). Locals in a `test` block are closure-captured into cells; a boxed local passed to a **union-typed parameter** (which now includes every `T?` since the `Ty::Optional` → `T | null` lowering) silently arrives as **`null`**, and a boxed local hitting `match` class arms matches nothing (`baml.panics.Unreachable`). Initially misdiagnosed as a canary regression — it isn't; plain functions are unaffected (verified `baml.sys.exec` + `ProcessOptions` works through a function). Workaround, per the established convention: the affected test bodies are hoisted into top-level helper functions (`plain_date_to_plain_datetime_with_time`, `toml_item_to_json_array`). The silent-null mode makes this worth prioritizing — a ready-made repro now lives in `ns_time/datetime.baml`.
5. **Default parameter values on `$rust_function`s leak `OmittedArg`** (discovered here — this PR is the first stdlib use of defaults on natives). Defaults are filled by the *callee's BAML prologue*; native functions have no prologue, so an omitted argument reaches the native glue as the `OmittedArg` sentinel and panics (`TypeError { expected: ..., got: OmittedArg }`). **Workaround used:** every public defaulted function is a thin pure-BAML wrapper (whose prologue fills defaults) over an all-required private native — `PlainTime._from_components`, `PlainDateTime._from_components`, `PlainDate._to_plain_datetime`. Suggest a compiler follow-up: either fill defaults for native callees too, or reject defaults on `$rust_function` parameters at compile time.

## Language quirks encountered (documented in code, no action needed)

- `match` bindings don't narrow optionals (`null => ... / let x => ...` leaves `x: T?`) — handled with documented unreachable `?? fallback`s.
- `int` doesn't promote to `bigint` in field initializers (method args and operators do promote) — handled with `0n + x`.
- `baml.json.*` doesn't resolve in `catch` patterns from inside the `baml` package; `root.json.*` does.


antoniosarosi:
@coderabbitai resolve

PR #3698, opened 2026-06-04
fix: thread inferred generic type args to callee frames at runtime

hellovai:
## What was wrong

A generic call whose type args are **inferred** rather than written ran its callee with `T = unknown`. TIR inferred the bindings, used them to compute the result type, then threw them away; MIR seeded `frame.type_args` from receiver class args ++ **explicit `<...>` args only**. On current canary (`d968182`, verified pre-merge):

| Program | Wrong (canary) | Right (this PR) |
|---|---|---|
| `function tn<T>(a: T[]) -> string { reflect.type_of<T>().to_string() }` — `tn([1,2,3])` | `"unknown"` | `"int"` |
| Same, class static: `Holder.tn([1,2,3])` | `"unknown"` | `"int"` |
| `let it: IterA<int, never> = ArrA.of([1,2,3]); it.collect().length()` (2 implementors) | **0** — wrong implementor runs | **3** |
| `match (ArrBag.make([5,6,7])) { let a: ArrBag<int> => ... }` | falls to `_` arm — instance carries `class_type_args=[unknown]` | matches `ArrBag<int>` |
| `fwd<int>(true \|\| false)` where `fwd<T>(x: bool) -> bool { x }` | **`null`** (pre-existing emitter miscompile, surfaced by this work) | `true` |

The dispatch fallout is the nasty part: instances built by inferred static ctors are born with `class_type_args = [unknown]`, every pinned `is_type C<int>` dispatch guard fails, and the interface-dispatch switch's **last candidate arm is an unguarded else** — so whichever implementor is last in hash order silently runs. That made the bug look "fragile" (renames, extra declarations in the file, or method names flip the arm order and mask it), and it is why #3667's regression suite stayed green: those tests pass by luck of arm order — their instances carry the same broken `[unknown]` args (now pinned by a direct `IsType` regression test that fails without this fix).

## What the fix does

**1. TIR** (`baml_compiler2_tir`): `check_call_inner` records the post-inference bindings per call expr in `call_inferred_type_args` (sorted `Vec<(Name, Ty)>`), plumbed through `ScopeInference` / `DefaultParameterInference` with the same save/restore discipline as `call_plans` (parameter defaults + inline lambda inference).

**2. MIR** (`baml_compiler2_mir`): at the call-lowering merge point, `emit_inferred_call_type_arg_ops` emits those bindings as leading `LoadType` frame args in the callee's De Bruijn order (`callee_frame_generic_params` = owner params ++ fn params, mirroring `enclosing_generic_params`). Conservative gates preserve the status quo when seeding can't be done soundly: explicit `<...>` args, sys-ops (positional Rust glue), function values (no resolution), misaligned receiver prefixes, all-unknown seeds. Bindings nesting a TypeVar **outside** the caller's generic params — a rigid `Self` from a Self-pinned call — are demoted to `unknown`: naive templating lowers them to `void`, which would seed `Self[]` as `void[]` (caught by adversarial review; guarded by `nested_self_inferred_binding_not_voided`).

Also: the path-callee branch no longer prepends a stray `Constant::Null` receiver to **self-less static** methods (`ArrA.of(...)`). The VM's arity-based pop silently tolerated the orphan, but it corrupts the leading type-arg window once calls carry seeds.

**3. EMIT** (`baml_compiler2_emit`): pre-existing `ShortCircuit` miscompile, reproducible on canary without this PR. The emitter assumed the destination local was always stack-carried; when the stack-carry pass rejects the carry (e.g. a `LoadType` operand pushes before the use), the taken path never stored the slot while `eval_rhs` did — a half-applied phi, so `a || b` fed into such a call reads an unwritten local. The taken path now stores whenever the destination is not stack-carried (`&&` via a small trampoline). Bonus: the previously-failing `if_else_assigned_then_passed_to_call` shape now passes.

## Verification

- Clean A/B on the same canary base: all five wrong behaviors above reproduce on `origin/canary` and flip with this PR.
- In-BAML suite **1646/1646**; MIR/codegen/bytecode/format snapshot targets green; workspace Rust tests green; clippy/fmt clean.
- Snapshot churn is exactly the expected categories: `+load_type` seeds at inferred generic call sites, `-const null` stray receivers, store/load pairs where short-circuit carries are now correctly rejected.
- 4 new regression tests in `ns_inferred_generic_type_args` (free-fn reflect, instance `IsType` after inferred ctor, the lazy-iterator collect shape with 2 implementors, nested-`Self` demotion).
- Adversarially reviewed (multi-agent: TIR soundness / MIR De Bruijn alignment / emit fix / blast radius): 2 confirmed findings, both addressed in this PR.

## Known remaining (pre-existing, intentionally out of scope)

- The interface-dispatch switch does not thread a dispatched method's **own** inferred fn-level generics (interface method with its own `<U>` reading `U` at runtime).
- Function **values** (`let f = ArrA.of; f([1])`) stay unseeded — no resolution at the call site.
- The dispatch switch's last candidate arm is still an unguarded else; a guarded-miss panic would be a better failure mode than silent fallthrough if type args ever regress.
- Test-block `let` bindings consumed by a later statement in the block lower to `Constant(Null)` — separate lowering bug discovered during this work.

## Perf note

Every inferred generic call now carries `LoadType` seeds. The existing bench corpus is **monomorphic and cannot observe this path**, so a green bench run is not evidence — this PR adds a `compute/generic-apply-inferred-1m` speedtest workload that exercises it; a release-mode A/B (total cycles + allocs/iter) against canary is the meaningful pre-merge measurement.

## Release Notes

* **New Features**
  * Generic function and method calls now automatically infer missing type arguments at call sites and properly propagate them to callees at runtime, enabling more flexible generic code without explicit type specifications.

* **Tests**
  * Added comprehensive test coverage for inferred generic type argument dispatch, including interface methods, default methods, and nested generic scenarios.

### DIFF
--- a/baml_language/crates/baml_compiler2_emit/src/emit.rs
+++ b/baml_language/crates/baml_compiler2_emit/src/emit.rs
@@ -2223,12 +2223,32 @@ impl<'ctx, 'obj> StackifyCodegen<'ctx, 'obj> {
             Terminator::ShortCircuit {
                 operand,
                 is_and,
-                destination: _,
+                destination,
                 eval_rhs,
                 join,
             } => {
                 // Legacy-style short-circuit using JumpIfFalse (peek, no pop).
-                // The destination local is PhiLike — value stays on TOS, no store/load.
+                //
+                // The short-circuit (taken) path leaves the operand value on TOS
+                // and jumps to the join. The `eval_rhs` block computes and stores
+                // the result via its own trailing `destination = <rhs>` statement.
+                // Both paths must agree on where the result lives at the join:
+                //
+                //  * When `destination` is stack-carried (PhiLike etc.), the join
+                //    consumes the value straight off TOS, and the `eval_rhs` store
+                //    is also elided (EvalNoStore) — so the taken path correctly
+                //    leaves the value on TOS and we emit no store here.
+                //  * Otherwise `destination` is a real slot: the `eval_rhs` store
+                //    writes the slot (StoreSlot) and pops, so the join reads via
+                //    `LoadVar`. The taken path must therefore also store its TOS
+                //    value into the slot — otherwise the slot is never written on
+                //    the short-circuit path and the join reads uninitialized
+                //    state. We emit that store on the taken path before the join.
+                let store_on_taken_path = !matches!(destination, Place::Local(l)
+                if matches!(
+                    pull_semantics::local_store_behavior(self.analysis.classifications[l]),
+                    pull_semantics::LocalStoreBehavior::KeepOnStack
+                ));
                 self.emit_operand_pull(operand);
 
                 if *is_and {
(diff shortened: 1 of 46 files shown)


aaronvg:
superseded by this (which adds the type args in the CallPlan instead) 
https://github.com/BoundaryML/baml/pull/3702

PR #3701, opened 2026-06-04
feat(stdlib): baml.iter lazy iterator protocol (WIP / status snapshot)

hellovai:
> **Draft / status snapshot.** Stacked on #3698 (`hellovai/iterators2`). Base it there; do not merge into `canary` until #3698 lands. This PR documents where the `baml.iter` lazy-iterator protocol stands and what compiler bugs still block it.

## What this adds

A pure-BAML `baml.iter` namespace under `crates/baml_builtins2/baml_std/baml/ns_iter/`:

- **`iter.baml`** — `Done` sentinel, `Iterable<T, E>` / `Iterator<T, E> requires Iterable<T, E>` interfaces, lazy adapter methods (`map`/`filter`/`filter_map`/`flat_map`/`peekable`/`step_by`/`chain`) and consumer methods (`collect`/`reduce`/`count`/`some`/`every`/`find`).
- **`utility.baml`** — `ArrayIterator<T>`, `Range`, `Repeat<T>` sources.
- **`adapters.baml`** — `Map`/`Filter`/`FilterMap`/`FlatMap`/`Flatten`/`Peekable`/`StepBy`/`Chain` adapter classes.

All `implements` blocks are **in-body**. The out-of-body `implements<T> Iface for Class<T>` form does **not** dispatch correctly for inferred generic instantiations (see Known issues), so everything is written in-body.

Also adds `#![allow(clippy::iter_not_returning_iterator)]` in `bex_vm/src/package_baml` — the BAML `iter()` method generates native glue literally named `iter` returning `Value`, which trips that lint on generated code.

## Status

Verified against this branch (`cargo test -p baml_tests --test baml_src`, namespace `iter`):

### ✅ Working
- `ArrayIterator.new(...)` and `Range.new(...)` sources
- consumers: `collect`, `reduce`, `count`
- `chain`

These pass because #3698 fixes inferred-generic type-arg threading **for in-body implementors**, and `Chain<T, E>` implements `Iterator<T, E>` (no union error arg).

### ❌ Blocked — each on a distinct, still-open compiler bug
| Feature | Blocking bug |
|---|---|
| `.map()` / `.filter()` / `.filter_map()` / `.flat_map()` | **Union error-arg interface coercion.** An adapter whose interface instantiation has a union error arg — `Iterator<R, E \| E2>` — fails to subtype-coerce. Even an explicit `Map<int,int,never,never>{}` is rejected as `Iterator<int,never>`. This is the main blocker for the closure adapters. |
| `.step_by()` | Its `next()` uses a **C-style `for` loop**, which is separately buggy in the VM. |
| `[1,2,3].iter()` | The `T[]` **blanket `implements`** (out-of-body, on a builtin array type) does not attach `.iter()` — `type int[] has no member iter`. |
| out-of-body `implements<T> Iface for Class<T>` + inferred ctor | Still misdispatches (the in-body conversion is the workaround used here). |

The blocked cases are kept in `ns_iter/iter.baml` as commented-out tests with the bug each waits on.

## Repro for the headline remaining bug

Minimal, verified repro for the **union error-arg coercion** issue is the natural failure of `it.map(f).collect()`; the earlier inferred-static-ctor dispatch bug (now fixed by #3698) has a standalone repro in the task notes.

## Not done / out of scope here
- `for (let x in iterable)` desugaring (TIR/MIR) — deferred.
- Native `$rust_function` `next()` + `ForIter` opcodes (perf) — deferred.
- Python/TypeScript host projection — deferred.
- Native `next()` for the closure adapters depends on the union-coercion fix first.


aaronvg:
implemented

PR #3704, opened 2026-06-06
Use ring crypto by default

sxlijin:
## Summary by CodeRabbit

* **Chores**
  * Updated cryptographic backend dependencies across multiple components, switching the default TLS provider from AWS-based to Ring-based cryptography for improved compatibility and consistency.


aaronvg:
the ecosystem is moving to aws-lc-rs? why change it for everything by default?

PR #3765, opened 2026-06-15
prof: cut BEX tracing producer + consumer overhead

antoniosarosi:
## Summary

Reduces the runtime overhead of the BEX tracing system on both sides of the ring: the **producer** (the VM encoding/pushing a raw record per call into a lock-free SPSC ring) and the **consumer** (the background thread transcoding to protobuf and writing per-engine `.bamlprof`). Profiling is default-on, so this is paid by every run.

Six commits, each measured and committed on its own merit. **No `.bamlprof` format change**; byte-identical wire output (guarded by a differential test). All `bex_events` unit tests, `bex_engine`'s `prof_gate`, the loom/miri concurrency suites, and clippy `--all-features` are green.

## Commits

| Commit | Side | Change |
|---|---|---|
| `e7daa17` | consumer | Reuse one encode buffer instead of a `Vec::with_capacity` **per event** (was a malloc + a redundant `encoded_len` walk per record). |
| `85bf08b` | producer | Encode records **directly into the ring slot** (`Ring::push_with` / `Buf::with_slice_mut`) — drops the intermediate stack buffer, the 41-byte zeroing, and one of two copies. Loom keeps per-byte tracked stores. |
| `d8ce1f2` | producer | Encode fixed fields with `ptr::write_unaligned` instead of `copy_from_slice` — the latter did **not** inline at `opt-level="s"` (the shipped profile): ~78 instructions/record → ~15, no libcall. |
| `d4eb542` | consumer | Convert ticks→ns with a **precomputed reciprocal multiply-shift** instead of a per-record u128 **divide** (~20-40 non-pipelined cycles). Exact for rate (1,1); ≤1 ns rounding for measured TSC rates, monotonic. |
| `d55970` | consumer | **Hand-roll the protobuf** for the two hot events (`CallFunction`/`EndFunction`) in one pass with back-patched 1-byte length prefixes, instead of prost (which walks this nested-oneof message **3×**). Cold records stay on prost. Same technique as Perfetto's protozero. |
| `905d6a7` | consumer | **One `write` per drained range** instead of one per event: accumulate the whole range (one ~256 KiB segment slice) into the writer's buffer, then a single write straight to the `File`. Drops the 8 KiB `BufWriter` (~4,300 → ~135 write syscalls / 2M-event run). |

### Consumer — ~2.4× throughput
`prof_drain_throughput`: **7.7M → ~18M events/s/core.** The reciprocal removes the per-record divide; the hand-rolled encoder removes ~2/3 of the prost cost (prost was size-walking the message twice and byte-walking once); the batched range write removes the per-event syscall churn.

### End-to-end, single process (clean isolation)
Same binary, `BAML_PROFILE` toggled, real disk IO, **profile dir wiped before each run** so there is no cross-run accumulation. Medians, ms:

| workload | `off` (tracing disabled) | `discard` (producer floor, no IO) | `full` (real IO) |
|---|---|---|---|
| pure-call-1m (1M trivial calls) | 111.8 | 131.5 (+18%) | 169.3 (**+51%**) |
| call-chain-100x10k | 82.3 | 101.7 (+24%) | 125.9 (**+53%**) |

Decomposition on these adversarial call-storm microbenches: **producer ring-emit ~+18-24%**, **consumer transcode+write+fsync ~+30%**. The producer cost is at the encode/push floor (the clock read itself is only ~1-6 ns/pair — verified by a stub experiment; the rest is encoding + the ring commit, both irreducible without dropping records). The consumer (~18M ev/s) nearly matches the producer's ~18-20M ev/s burst, leaving a small drain tail plus the one-file write/fsync.

### Corpus medians
Across the full speedtest corpus, the same-binary **tracing-on producer cost is +3.6% median**; **total full-pipeline +8.2% median**. Real-shaped workloads (loops, strings, classes) are well under +10% — only the call-storm microbenches (millions of trivial calls, which no LLM-bound BAML program emits) are high.

> Note on a measurement pitfall: a multi-leg speedtest run *without* cleaning `.bamlprof` between sample-processes accumulates GBs and triggers NVMe dirty-page writeback throttling, inflating the `full` leg to +200-500%. The single-process numbers above (dir cleaned per run) are the representative cost; the inflation is a harness artifact, not a per-run cost.

## Follow-ups (not in this PR)

- **Per-segment compression (lz4).** The remaining ~+30% consumer+IO on record-dense workloads scales with `.bamlprof` byte volume; per-segment lz4 (~1.8-3× smaller files) is the lever, viable now that the consumer has CPU headroom. It's a **format-version bump**, so it's kept out of this no-format-change PR (separate branch/PR). The batched-range write here (`905d6a7`) is its structural carrier.
- **Record-shrink / delta-ts** — redundant with a block compressor on disk; revisit only for the ring overflow budget under sustained multi-second storms.
- **Leaf single-push fusion** — evaluated and rejected (~3-4 ns/pair for a high-risk unwind/cancel/drain change).
- **`opt-level=3`** — evaluated and reverted (speeds the VM ~17% but not the tracing delta; risks the wasm size budget).
- Fidelity levers (sampling / leaf-elision) are the only thing that *bounds* a pathological call-storm; out of scope here.

## Verification

- `cargo test -p bex_events` (53, incl. `hand_encoder_matches_prost` differential byte-equality and `reciprocal_matches_exact_divide`)
- `cargo test -p bex_engine --test prof_gate` (decodes the written bytes unchanged through `read_bamlprof`)
- loom (`--cfg baml_loom`) 34/34 + miri 33/33 on the ring/`push_with` change
- clippy `--all-targets --all-features` clean; CI mirror under `rustup run 1.93.0`

## Summary by CodeRabbit

* **Refactor**
  * Improved profiling event recording and ring emission to encode directly into ring slots, reducing per-event overhead.
  * Streamlined profile writing by reusing a shared scratch buffer and optimizing the hottest event encoders.
  * Updated consumer processing to buffer and flush by committed ring ranges, including heartbeat persistence.
  * Accelerated tick-to-nanosecond conversion using a precomputed multiply-shift reciprocal scale.
* **Tests**
  * Added/updated assertions to verify reciprocal-based tick conversion matches exact conversion within 1ns and updated refinement expectations.

### DIFF
--- a/baml_language/crates/bex_events/src/prof/clock.rs
+++ b/baml_language/crates/bex_events/src/prof/clock.rs
@@ -79,8 +79,23 @@ mod imp {
         meta: ClockMeta,
     }
 
+    static ANCHOR: OnceLock<Anchor> = OnceLock::new();
+
+    /// The process clock anchor. Fast path is a single `OnceLock::get`
+    /// (Acquire load + branch) that inlines into `now_ticks` — and thus
+    /// into the VM's call hot path — so the raw counter read can pipeline
+    /// with the surrounding encode/push work rather than hiding behind a
+    /// cross-crate call. The one-time init is `#[cold]` and out-of-line.
+    #[inline]
     fn anchor() -> &'static Anchor {
-        static ANCHOR: OnceLock<Anchor> = OnceLock::new();
+        match ANCHOR.get() {
+            Some(a) => a,
+            None => anchor_init(),
+        }
+    }
+
+    #[cold]
+    fn anchor_init() -> &'static Anchor {
         ANCHOR.get_or_init(|| {
             let (source, ns_per_tick) = detect();
             let kind = match source {
--- a/baml_language/crates/bex_events/src/prof/clock.rs
+++ b/baml_language/crates/bex_events/src/prof/clock.rs
@@ -350,6 +350,14 @@ pub struct TickConverter {
     seg_base_ns: u64,
     /// ns-per-tick as (numer, denom).
     rate: (u64, u64),
+    /// Fixed-point reciprocal of `rate` for the hot path: `numer << SCALE_SHIFT
+    /// / denom`, so `to_ns` is a `dt * scale >> SCALE_SHIFT` multiply-shift
+    /// instead of a per-record u128 divide (~20-40 non-pipelined cycles). Kept
+    /// in sync with `rate` at every install site. Exact for rate (1,1)
+    /// (scale = `2^SCALE_SHIFT` → ns == dt); ≤1 ns rounding for measured TSC
+    /// rates, which is below tracing-timestamp resolution and preserves
+    /// monotonicity.
+    scale: u128,
     /// Previous segment `(base_ticks, base_ns, rate)`, kept after a
     /// refinement so pre-switch ticks still convert on their original line.
     prev: Option<(u64, u64, (u64, u64))>,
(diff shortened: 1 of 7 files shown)


antoniosarosi:
Addressed both CodeRabbit nitpicks in 690634752 (doc-only): `encode_to` now states the buffer-size safety precondition (UB, not panic, on the unchecked `write_unaligned` path) with the satisfying callers; `prof_push_record` doc updated to the direct-ring-slot path (the stack-buffer/zeroing description was stale).

PR #3814, opened 2026-06-20
Diagnose postfix `!` (non-null assertion) instead of cascading `if` without `else` errors

ATX24:
## The error

Writing `xs.at(0)!` (a TypeScript/Swift/Kotlin-style non-null assertion) inside an `if` branch produced a misleading error cascade that blamed the `else` keyword and the closing braces — never the `!`:

```baml
function f(xs: int[]) -> int {
    if (xs.length() > 0) {
        xs.at(0)!
    } else {
        0
    }
}
```

Output (before the fix):

```
error: `if` without `else` cannot be used as a value; add an `else` branch   (E0001)
error: Expected expression, found '}'                                          (E0010)
error: Expected expression, found else                                         (E0010)
error: Expected '}', found EOF                                                  (E0010)
```

None of these point at the `!`. The accurate message only surfaced when the expression was isolated outside an `if`.

## Root cause

In the Pratt expression loop `Parser::parse_expr_bp` (`baml_language/crates/baml_compiler_parser/src/parser.rs`), `!` lexes as `TokenKind::Not`, which is only handled as a **prefix** unary operator in `parse_prefix`. When `!` appears in **postfix** position (after a complete expression), the loop had no branch for it and simply `break`ed, leaving the `!` unconsumed.

The statement loop in `parse_block_expr` then re-entered `parse_stmt` → `parse_expr` → `parse_prefix`, which *does* treat `!` as a prefix operator. It consumed the `!` and tried to parse an operand from the following `}`/`else`, yielding `Expected expression, found '}'` etc. The malformed then-block (no tail value) then made the `if` look like an `if` without a usable `else`, producing the `E0001` "if without else" error in the TIR (`baml_compiler2_tir/src/infer_context.rs`). Hence the cascade.

## The fix

Added a branch in `parse_expr_bp`'s postfix loop for `TokenKind::Not`: a dangling postfix `!` is now consumed in place and reported with a single targeted diagnostic at the `!` token:

> `unexpected '!'; BAML has no non-null assertion operator — unwrap optionals with '?? <default>' or 'if (let x) = opt { ... }'`

Because the `!` is consumed where it appears, the statement loop never re-interprets it as a prefix operator, so the `if`/`else` structure parses cleanly and the spurious `E0001`/`E0010` cascade disappears. `!=` is unaffected (it lexes as a single `NotEquals` token, not `!` followed by `=`).

The `!self.has_newline_ahead()` guard ensures a `!` at the *start of the next line* is still treated as a legitimate prefix unary operator on a fresh statement (BAML separates statements by newlines, not semicolons).

## Verification

Commands run from inside `baml_language/`:

- `cargo build -p baml_compiler_parser` — compiles cleanly.
- `cargo test -p baml_compiler_parser postfix_bang` / `not_equals` — parser unit tests pass, including:
  - `postfix_bang_emits_targeted_diagnostic_without_cascade` — asserts exactly one diagnostic, that it mentions the non-null assertion operator, that no diagnostic blames `else`/`}`, and that the `IF_EXPR` keeps both then/else blocks.
  - `not_equals_is_unaffected_by_postfix_bang_handling` — guards against regressing `!=`.
  - `prefix_bang_on_next_line_is_not_treated_as_postfix` — guards the newline rule.
- `cargo test -p baml_tests postfix_non_null_assertion` — end-to-end snapshot project under `projects/broken_syntax/postfix_non_null_assertion/` passes (lexer/parser/diagnostics).
- `cargo fmt` — clean.

New diagnostics output for the reproduction (after the fix), pointing precisely at the `!`:

```
  [type]  error: type mismatch: expected int, got int | null   (main.baml:6)
  [parse] error: unexpected '!'; BAML has no non-null assertion operator — unwrap optionals with '?? <default>' or 'if (let x) = opt { ... }'   (main.baml:7:17, points at the `!`)
```

Note: this adds a new parser diagnostic, so captured snapshots in the wider suite may shift on the first CI run; regenerate with `cargo insta accept --all` if so.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parsing of postfix `!` so it’s detected as an invalid non-null assertion operator and reports exactly one targeted diagnostic at the `!` location, avoiding misleading downstream errors.
  * Ensured `!=` parsing is unchanged and prefix `!` on the following line continues to be handled correctly.

* **Tests**
  * Added/updated regression fixtures and unit tests covering `xs.at(0)!`, `!=` behavior, and valid prefix `!` cases to prevent future parsing regressions.

<div><a href="https://cursor.com/agents/bc-b3e1add0-0c3e-4dfd-8417-3e5673dc75c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b3e1add0-0c3e-4dfd-8417-3e5673dc75c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parsing of postfix `!` so it’s treated as an invalid non-null assertion operator and produces exactly one targeted diagnostic at the `!` location, preventing misleading cascading parse errors.
  * Kept `!=` handling unchanged, and ensured a prefix `!` starting on the next line still parses correctly.

* **Tests**
  * Added/updated regression fixtures and unit tests covering postfix `xs.at(0)!`, `!=`, and valid prefix `!` cases to prevent future parsing regressions.


ATX24:
if let x = opt { ... } should be paranthesis around this
if (let x = opt) { ... }
^

ATX24:
no, you misunderstand. 

I want the error message to be 
  [parse] error: unexpected '!'; BAML has no non-null assertion operator — unwrap optionals with '?? <default>' or 'if (let x) = opt { ... }'   (main.baml:7:17, points at the `!`)

make sure that we fix the PR description as well

PR #3819, opened 2026-06-21
fix(python): rebuild generated models so class order can't break Pydantic (#793)

### LINKED ISSUE #793: Pydantic error when creating BAML classes with composition
One bug I noticed recently due to the way baml_client/types.py is sorted alphabetically it leads to AttributeError: 
__pydantic_parent_namespace__
```baml
class B {
}

class A {
  property1 B
}
```

```python
# baml_client/types.py
class A(BaseModel):
  Property1: "B"

class B(BaseModel):
  ...
```

It leads to __pydantic_parent_namespace__. It can easily be fixed by changing the names of the classes in baml. Using pydantic 2.1.0 and Python 3.9.16

# Possible fix
Always generate classes based on composition order (topo-sort).

GouravSingal-code:
## What & why

Fixes #793.

The generated `baml_client/types.py` lists classes alphabetically, so a class can reference another that is defined **later** in the file via a string forward reference (e.g. `field: "Other"`). On Pydantic 2.1.0 this can fail to resolve at import/use time:

```
AttributeError: __pydantic_parent_namespace__
# (or) `X` is not fully defined; you should define ..., then call `X.model_rebuild()`
```

Minimal repro from the issue:

```baml
class B {}
class A { property1 B }   // A sorts first, references B which is defined later
```

## Approach

Per @baml-sam's guidance in the issue, this **forward-declares instead of reordering**. Topologically sorting classes can't work once types are split across files with circular references, so instead the generator now emits a `model_rebuild()` for **every class at the end** of `types.py` and `stream_types.py`, after all classes are defined. Forward references then resolve regardless of declaration order:

```python
class A(BaseModel):
    property1: "B"
class B(BaseModel):
    ...

# Model rebuilds (2)
A.model_rebuild()
B.model_rebuild()
```

- Pydantic **v1** uses `update_forward_refs()`.
- Files with no classes (enum-only / alias-only) emit nothing.

## Changes

- `engine/generators/languages/python/src/generated_types.rs` — rebuild template + `render_py_model_rebuilds()`.
- `engine/generators/languages/python/src/lib.rs` — appends the block to `types.py` and `stream_types.py`.
- `engine/generators/languages/python/src/ir_to_py/classes.rs` — unit tests.
- Regenerated Python codegen snapshots (the `# Model rebuilds` block). Pre-existing, unrelated snapshot drift is intentionally left out of scope.

## Testing

- `cargo test -p generators-python` — 168 pass (incl. 3 new unit tests + the codegen suite).
- Verified the generated output resolves forward references on the reporter's stack (**Python 3.9 + pydantic 2.1.0**) and on **latest pydantic (2.13.4)** and **pydantic v1 (1.10)** — no regressions, no deprecation warnings. `model_rebuild()` is a no-op for already-complete models, so there's no import-time cost for models without forward refs.

## Notes / scope

- Python-only. TS/Go/Ruby/Rust use structural typing and don't have this forward-reference ordering problem.
- BAML's dynamic types (`TypeBuilder`) are native-Rust, not Pydantic models, so they're unaffected.

## Release Notes

* **Bug Fixes**
  * Improved generated Python SDK modules (`types.py` and `stream_types.py`) to stabilize Pydantic models after all model definitions, resolving string forward-reference issues across many model shapes.
  * Rebuilds are emitted as explicit ordered calls, while recursive models are intentionally omitted to avoid eager recursion.
* **Tests**
  * Added generator-level tests validating the emitted “model rebuild” logic for both Pydantic v1 and v2, recursive skipping behavior, and empty input handling.


sxlijin (approved):
LGTM, thanks!

GouravSingal-code:
Thanks @sxlijin  🙏 The only red checks left are the two Vercel preview deploys (beps, promptfiddle2). They fail with "Authorization required to deploy" — that just happens because the PR comes from a fork, it's not a problem with the change. Should be good to merge once someone can authorize or skip those checks.

sxlijin:
Sorry, I thought I clicked the "approve workflows" button when I did the first review. Looks like integ tests are failing substantially, something about adding the model rebuilds triggered it. I'm not sure a priori what this would be, it's probably something in how pydantic models recursive classes.

GouravSingal-code:
Hi @sxlijin — you were right, it was model_rebuild() recursing on recursive type aliases (RecursiveAliasDependency). I made the rebuild best-effort (try/except per model), so recursive models are left as-is and the normal cases still resolve. Tested locally: the integ-tests now collect with 0 recursion errors (was 26). could you re-run the workflows when you get a sec? Thanks!

sxlijin:
Instead of just doing a blanket try-except, let's make client codegen only generate rebuilds for non-recursive classes. There's existing cycle detection logic, we should piggyback off it.

GouravSingal-code:
@sxlijin Done — switched to your suggested. Instead of the blanket try/except, the generator now only emits model_rebuild() for non-recursive classes,using the IR's existing cycle detection (finite_recursive_cycles structural_recursive_alias_cycles, following TypeIR::dependencies() so transitive cases like RecursiveAliasDependency → JsonValue are caught). Recursive models are skipped and left for Pydantic to resolve lazily.

sxlijin:
Until a human engages with my PR comments, I'm closing this PR.

This PR reimplements cycle detection, instead of piggybacking off the original.

sxlijin:
Awesome, thanks!

PR #3823, opened 2026-06-22
feat(openai-responses): surface response_id + document previous_response_id passthrough

divyvasal:
## What

Enables **resumable / stateful multi-turn** use of the OpenAI **Responses API**
(`provider "openai-responses"`) — without BAML holding any session state.

Today the Responses API's server-side conversation state (via `previous_response_id`)
isn't reachable from BAML: the response `id` is parsed but dropped, so a caller has
no way to chain turns. This PR closes that, keeping BAML stateless (the caller
supplies and chains ids; BAML never tracks or auto-injects them).

## Changes

**1. Surface the response id (output side).**
`parse_openai_responses_response` already deserializes `ResponsesApiResponse.id` but
discarded it. Added `response_id: Option<String>` to `LLMCompleteResponseMetadata`
and populate it from the Responses API `id`. It's `None` for every other
provider/endpoint (chat/completions, Anthropic, Google, Vertex, AWS, streaming).

**2. Document `previous_response_id` / `store` passthrough (input side).**
These already forward to the request body via the existing `properties.clone()` in
the `ResponsesApi` `build_body` — added an in-code comment so it's discoverable.

**Together:** read `response_id` from turn N's metadata → pass it as
`previous_response_id` into turn N+1 → server-side session resumes.

```baml
client<llm> CoralResp {
  provider openai-responses
  options { model "gpt-..." }
}
// turn 1: capture response_id from the run metadata
// turn 2: set previous_response_id = <that id> (e.g. via ClientRegistry options) → resumes
```

## Design note

This deliberately keeps BAML **stateless / pure-function**: BAML forwards what the
caller provides and surfaces what the API returns — it does **not** track response
ids or auto-inject `previous_response_id`. That's left to the caller (or an
application-layer/proxy that maps a session → last id), consistent with BAML's
existing design philosophy.

## Tests

Updated the three `parse_openai_responses_*` unit tests to assert the surfaced id.

> ⚠️ Opened as **draft**: I wasn't able to run the full `engine` Rust build/test
> suite in my environment. The change is mechanical (one new `Option<String>`
> metadata field, populated at the Responses parser, `None` at all other
> constructor sites), but please let CI validate the build/tests. Happy to adjust
> field placement / naming to match maintainer preference.

## Summary by CodeRabbit

* **New Features**
  * Added response tracking capability across all supported language model providers. Response identifiers are now captured and can be referenced in subsequent API calls to maintain conversation context and continuity in multi-turn interactions.

* **Tests**
  * Updated test cases and fixtures to validate the new response tracking functionality across all provider implementations.

### DIFF
--- a/engine/baml-runtime/src/internal/llm_client/mod.rs
+++ b/engine/baml-runtime/src/internal/llm_client/mod.rs
@@ -319,6 +319,11 @@ pub struct LLMCompleteResponseMetadata {
     pub output_tokens: Option<u64>,
     pub total_tokens: Option<u64>,
     pub cached_input_tokens: Option<u64>,
+    /// Provider-assigned response id, when the API returns one (e.g. the OpenAI
+    /// Responses API `id`). Surfaced so stateless callers can capture it and pass
+    /// it back as `previous_response_id` on the next turn to resume server-side
+    /// state. `None` for providers/endpoints that don't return a response id.
+    pub response_id: Option<String>,
 }
 
 // This is how the response gets logged if you print the result to the console.
--- a/engine/baml-runtime/src/internal/llm_client/primitive/anthropic/response_handler.rs
+++ b/engine/baml-runtime/src/internal/llm_client/primitive/anthropic/response_handler.rs
@@ -87,6 +87,7 @@ pub fn parse_anthropic_response<C: WithClient + RequestBuilder>(
         request_options: client.request_options().clone(),
         model: response.model,
         metadata: LLMCompleteResponseMetadata {
+            response_id: None,
             baml_is_complete: is_done(&response.stop_reason),
             finish_reason: response.stop_reason.clone(),
             prompt_tokens: Some(response.usage.input_tokens),
@@ -224,6 +225,7 @@ mod tests {
             model: model_name,
             request_options: client.request_options().clone(),
             metadata: LLMCompleteResponseMetadata {
+                response_id: None,
                 baml_is_complete: true,
                 finish_reason: Some("end_turn".to_string()),
                 prompt_tokens: Some(321),
--- a/engine/baml-runtime/src/internal/llm_client/primitive/aws/aws_client.rs
+++ b/engine/baml-runtime/src/internal/llm_client/primitive/aws/aws_client.rs
@@ -1019,6 +1019,7 @@ impl WithStreamChat for AwsClient {
                     model: self.properties.model.clone(),
                     request_options,
                     metadata: LLMCompleteResponseMetadata {
+                        response_id: None,
                         baml_is_complete: false,
                         finish_reason: None,
                         prompt_tokens: None,
@@ -1460,6 +1461,7 @@ impl WithChat for AwsClient {
                 request_options,
                 model: self.properties.model.clone(),
                 metadata: LLMCompleteResponseMetadata {
+                    response_id: None,
                     baml_is_complete: matches!(
                         response.stop_reason,
                         bedrock::types::StopReason::StopSequence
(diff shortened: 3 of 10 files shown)


divyvasal:
Validated locally with the repo toolchain (cargo 1.94.1):
- `cargo check -p baml-runtime` → compiles clean (no errors; no new warnings on the touched files).
- `cargo test -p baml-runtime parse_openai_responses` → 5/5 pass, including the three responses-parser tests updated to assert the surfaced `response_id`.

Marking ready for review. @coderabbitai review

aaronvg:
we do provide the full http response via the Collector. This would be the preferred approach for grabbing this value.

divyvasal:
Thanks @aaronvg — confirmed. `collector.last.calls()[-1].http_response().body` already carries the full Responses API payload (incl. `id`), so the new `response_id` metadata field is redundant — I'll drop it.

And since `previous_response_id` / `store` already forward to the request body via the openai-responses options passthrough, resumable sessions actually work end-to-end **today** with existing primitives: set `previous_response_id` in the client options (e.g. via `ClientRegistry`) and read the prior turn's id from the Collector. No runtime change required.

Happy to convert this into a short docs example for that pattern (Responses-API resume via options + Collector) if that's useful — otherwise I'll close this. Let me know which you'd prefer.

divyvasal:
To frame the actual goal: we're after **resumable-session parity with the OpenAI SDK**. There you pass `previous_response_id=...` and read `response.id` directly off the result:

```python
r1 = client.responses.create(input=...)
r2 = client.responses.create(input=..., previous_response_id=r1.id)
```

In BAML today: setting `previous_response_id` via the openai-responses options works, and the id is reachable via the Collector (`http_response().body`) — so it's **functionally** there. The only parity gap is ergonomic: OpenAI gives `response.id` first-class, whereas BAML requires parsing the raw HTTP body.

Would you be open to surfacing the Responses `id` as a **first-class accessor** (e.g. on `LLMCall` / `FunctionLog`, or in `FunctionLog.metadata()`) so callers don't parse raw JSON? I'll happily re-scope this PR to that Collector-side accessor (and drop the `LLMCompleteResponseMetadata` field from the original diff). If you'd rather keep `http_response().body` as the canonical surface, I'll add a short docs example for the resume loop and close. Whichever you prefer.

divyvasal:
Closing — resumable-session parity is achievable with BAML's existing primitives, so no runtime change is needed:
- **Set** `previous_response_id` / `store` → openai-responses client options (per-turn via `ClientRegistry`); forwarded to the request body by the existing passthrough.
- **Read** the response id → the Collector (`collector.last.calls()[-1].http_response().body` → `id`), per @aaronvg.

Thanks @aaronvg for the pointer to the Collector. Closing the metadata-field change as redundant.

divyvasal:
Reopening to pursue the one real gap: **ergonomic parity** for the response id. Functional parity already works (set `previous_response_id` via options + read via Collector). The only delta vs the OpenAI SDK is that there you get `response.id` first-class, whereas BAML requires parsing `http_response().body`.

@aaronvg — would you accept a thin first-class accessor for it, and if so which surface do you prefer?
1. `LLMCall.response_id()` (parsed from the response body), or
2. include `response_id` in `FunctionLog.metadata()`, or
3. keep `http_response().body` as canonical (then I'll just add a docs example + re-close).

I'll scope the diff to whichever you pick (and keep it Collector-side — dropping the original `LLMCompleteResponseMetadata` change). Just don't want to wire it across the Python/Ruby/Go/Rust clients before confirming the surface you'd merge.

sxlijin:
Closed and rejected: this is an ai-only drive-by pull request.

divyvasal:
Hi @aaronvg could you confirm if BAML has background mode like OpenAI or not? Is there a plan to support in the future or it shall forever remain stateless?

PR #3824, opened 2026-06-22
Add `baml.math.sum`, `baml.math.mean`, and `baml.math.median` aggregation builtins

ATX24:
## Summary

Adds three **native VM** aggregation functions to the `baml.math` namespace, operating over `float[]`:

- `sum(values: float[]) -> float throws never` — arithmetic sum, folded left-to-right from `0.0`; `sum([])` is `0.0` (never throws).
- `mean(values: float[]) -> float throws root.errors.InvalidArgument` — `sum(values) / values.length()`; throws `InvalidArgument` on an empty array.
- `median(values: float[]) -> float throws root.errors.InvalidArgument` — sorts a **copy** (the caller's array is not mutated) using IEEE-754 `totalOrder` (`f64::total_cmp`), returns the middle element for odd counts or the mean of the two middle elements for even counts; throws `InvalidArgument` on an empty array.

## Implementation

These are **native VM functions, not pure BAML.** Each is declared in the stdlib with `$rust_function` and the `//baml:vm` marker in `crates/baml_builtins2/baml_std/baml/ns_math/math.baml`, and implemented in Rust (~96 lines) in `crates/bex_vm/src/package_baml/math.rs` (the `BamlNamespaceMath` trait). Every function carries a `///` docstring with parameters, return value, errors, and examples.

The API is `float[]`-based because this compiler deliberately has no `int <: float` subtyping and no implicit int→float widening; the `sum` docstring shows how to map integer data to floats first (`ints.map((x: int) -> float { x * 1.0 })`). I placed these in the `baml.math` namespace rather than as methods on `Array<T>` because `Array<T>` is generic with no numeric bound.

## Relationship to B-492

This **partially** addresses B-492 ("Add `Array.fold`/`reduce` and `Array<int|float>.sum` to stdlib"): it delivers `float[]` aggregations under `baml.math`. It does **not** add the `int[]` sum, an `Array<int|float>.sum()` method, or the `Array.fold` / `Array.reduce` method surface requested there — those remain a follow-up, so B-492 stays open.

## Tests

12 `test` blocks in `crates/baml_tests/baml_src/ns_math/math.baml`:

- `math_sum_basic`
- `math_sum_single`
- `math_sum_negatives`
- `math_sum_empty_is_zero`
- `math_mean_basic`
- `math_mean_single`
- `math_mean_empty_throws`
- `math_median_odd`
- `math_median_even`
- `math_median_unsorted_input`
- `math_median_does_not_mutate_input`
- `math_median_empty_throws`

Snapshots regenerated (additive — the three new functions): the `describe` builtin listings and the `__baml_std__` HIR/MIR/codegen + package-items snapshots.

## Commands (run from `baml_language/`)

- `cargo test -p bex_vm` — native math + VM.
- `cargo test -p baml_tests` — includes the 12 new `math_*` tests and the regenerated snapshots.

## Summary by CodeRabbit

* **New Features**
  * Added `sum`, `mean`, and `median` helpers to the math toolkit for working with float arrays.
  * `sum` now returns `0.0` for empty arrays.
  * `mean` and `median` provide clear results for valid inputs and report invalid empty-input cases.

* **Bug Fixes**
  * `median` now preserves the original array order while calculating results.
  * Improved handling of numeric array values, including sorted and even-length medians.


ATX24:
Should probably be using rust functions here. No reason to use baml. Much faster.

PR #3834, opened 2026-06-23
feat(stdlib): nested-override ToString + obj.to_string() sugar

codeshaunted:
## Summary

Two string-conversion features on the opt-in `baml.ToString` interface (no magic methods), plus a rendering bug fix.

### 1. `baml.ToString` honors overrides at every depth
`string.from(value)` previously rendered containers/structs structurally, ignoring a *nested* value's `to_string` override. It now walks the value tree honoring overrides at any depth (collect override-bearing nodes pre-order -> dispatch each through the VM -> splice results back by a pre-order counter).

The runtime shim behind `string.from` also honors the two non-instance builtin implementors, `type` (`baml.TypeValue`) and `uint8array` (`baml.Uint8Array`), so `string.from(x)` and `x.to_string()` always agree.

### 2. `obj.to_string()` sugar
A 0-arg `to_string()` call on a value whose type has **no** real `to_string` method now lowers to `string.from(recv)`, so any value can be rendered with `.to_string()`.

Implemented operator-style (like `==` -> `baml.ops.equals_equals`), not as an AST rewrite:
- **TIR** types the fallback call as `string` (the AST node stays a method call).
- **MIR** lowers it to `string.from(recv)`.
- **Throws analysis** treats it as `throws never`.

So diagnostics keep showing `.to_string()` (never `string.from`), a real `to_string` from any interface (including user-defined ones) is dispatched normally rather than hijacked, captured/generic/**optional** receivers work (a null optional renders as `null`, since `string.from` is total over null), and it honors nested overrides identically to `string.from`.

### 3. Bug fix: quote map keys in structural rendering
The structural renderer now emits map keys quoted -- `{"a": (1, 2)}` instead of `{a: (1, 2)}` -- so keys containing `:`/`,` stay unambiguous and match how nested strings are quoted.

```baml
point.to_string()      // sugar -> string.from(point)
string.from(myMap)     // {"a": (1, 2)}  (nested Point override honored)
```

## Note on FromString
An earlier revision of this branch also added a `baml.FromString` interface + `string.to<T>(s)` parser. It was **removed**: it had no stdlib implementors, its primitive parsing overlapped `int.parse`/`float.parse`, structured deserialization is already `baml.json.from_json`, and ToString/FromString are not truly symmetric (every value has a display form; most have no canonical single-string form to parse back). It can return when there is a concrete need for generic, type-driven parsing of custom string forms.

## Tests
- `to_string_interface.rs` -- nested overrides (array / class-field / map / array-in-class / mixed-sibling alignment), quoted map keys, builtin dispatch.
- `tostring_sugar.rs` -- primitives, string, array, plain class, override dispatch, nested-override parity with `string.from`, chained/generic/captured-closure receivers, receiver-error not swallowed, user-interface `to_string` not hijacked, optional receivers (present + null).

Full workspace `nextest` passes; clippy clean.

## Summary by CodeRabbit

* **New Features**
  * Added `.to_string()` support that respects custom `baml.ToString` overrides, including nested values, arrays, maps, optionals, and generic cases.
  * Improved rendering so `string.from(...)` and `.to_string()` behave consistently, even for operator-style sugar calls.
* **Bug Fixes**
  * Ensured fallback handling only applies when a real `to_string` implementation is missing or untyped, so true overrides always win.
  * Prevented receiver evaluation errors from being masked by `.to_string()` fallback.
  * Updated structural output formatting (including map entry key rendering) to match override behavior.
* **Tests**
  * Added extensive `.to_string()` and override-dispatch test coverage.


codeshaunted:
@coderabbitai full review

PR #3847, opened 2026-06-25
fix(zed): pin language server download to engine release tag

hellovai:
## Problem

The Zed extension fails to start the BAML language server with:

> no asset found matching `"baml-cli-baml-language-0.12.2-nightly.20260625.d-aarch64-apple-darwin.tar.gz"`

The extension resolved the `baml-cli` binary via `latest_github_release`, which returns the newest release across the **whole** repo. But the repo publishes two independent release lines:

| Release line | Tag | Asset naming |
|---|---|---|
| Engine / `baml-cli` | `0.223.0` | `baml-cli-0.223.0-<target>.tar.gz` ✅ (what the extension wants) |
| `baml-language-*` nightlies | `baml-language-0.12.2-nightly.20260625.d` | `baml-language-…-<target>.tar.gz` |

The nightly line currently holds GitHub's "latest" pointer, so `release.version` became the nightly tag and the `baml-cli-{version}-{target}` template produced an asset name that doesn't exist.

## Fix

Pin the lookup to the engine release tag — the extension's own `CARGO_PKG_VERSION` (kept in sync with `extension.toml` via `tools/versions/zed.cfg`) — using `github_release_by_tag_name` instead of `latest_github_release`. The extension version and the language-server binary it downloads now stay locked together, and the nightly line is ignored entirely.

Also corrects a stale `Cargo.lock` version (`0.218.0` → `0.223.0`).

## Verification

`cargo check` passes clean. With the extension at `0.223.0`, the asset resolves to `baml-cli-0.223.0-aarch64-apple-darwin.tar.gz`, which exists on the engine release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Language-server downloads now use a fixed release version instead of always following the latest GitHub release.
  * This helps keep the installed engine aligned with the app version, reducing mismatches and unexpected updates.

### DIFF
--- a/engine/zed/Cargo.lock
+++ b/engine/zed/Cargo.lock
@@ -746,7 +746,7 @@ dependencies = [
 
 [[package]]
 name = "zed_baml"
-version = "0.218.0"
+version = "0.223.0"
 dependencies = [
  "log",
  "zed_extension_api",
--- a/engine/zed/src/lib.rs
+++ b/engine/zed/src/lib.rs
@@ -23,6 +23,13 @@ const HARDCODED_EXTENSION_CONFIG: HardcodedExtensionConfig = HardcodedExtensionC
 
 const GITHUB_REPO: &str = "BoundaryML/baml";
 
+// The engine (`baml-cli`) release we download the language server from is tagged
+// with this exact version. We pin to it rather than using `latest_github_release`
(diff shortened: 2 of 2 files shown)


sxlijin:
#3853 should take care of this

PR #3848, opened 2026-06-25
Migrate Zed extension to baml-language toolchain

j-hartshorn:
### Problem

Installing the BAML Zed extension fails with:

```
no asset found matching "baml-cli-baml-language-0.12.2-nightly.20260625.e-aarch64-apple-darwin.tar.gz"
```

`engine/zed/src/lib.rs` formatted the asset name as `baml-cli-{release.version}-{target}` from `latest_github_release("BoundaryML/baml")`. The auto-published `baml-language` toolchain releases are **not** flagged as GitHub pre-releases, so `/releases/latest` now returns a `baml-language-*` tag. The extension then builds `baml-cli-` + `baml-language-<version>` + `-<target>`, a doubled-up name that no release asset matches.

### Fix

Replace the bespoke GitHub-release download with the same thin-client model the VSCode extension uses, as a layered resolver in `language_server_command`:

1. **Explicit override** — `lsp.baml-language-server.binary.path` (+ `arguments`).
2. **`baml` wrapper on PATH** — `worktree.which("baml")`, run `baml lsp`. The wrapper resolves the project's pinned toolchain, so the LSP always matches `baml generate`.
3. **Managed download fallback** — fetch `pkg.boundaryml.com/manifest/v1/{channel}.json`, read `version` + `artifacts[<target>].url`, `download_file` the toolchain archive, run `bin/baml-cli lsp`. Pin-aware: honours `BAML_VERSION` / `baml.toml [toolchain] version`, otherwise the `canary` channel head.

`language_server_initialization_options` drops to `None` — the new server ignores initialization options (`bex_project` `on_request_initialize` only reads workspace roots). The unused `log` dependency is swapped for `toml`; version bumped `0.223.0` → `0.224.0`.

### Validation

- Builds to `wasm32-wasip2`.
- Canary toolchain `0.12.1`: `baml-cli check` against an existing `baml_src` project exits 0; `baml-cli lsp` completes an `initialize` handshake and advertises hover/completion/semantic-tokens.

### Notes

- The `[features] debug` block in `Cargo.toml` predates this change and is no longer wired to anything (the old `LocalBuild` source is gone). Left untouched to keep this PR focused; happy to remove it separately.
- The download fallback can match an exact `baml.toml` version pin, but not a `channel =` pin (the wrapper resolves channels against machine-local state an editor can't see). Installing the `baml` wrapper — path 2 — is the way to guarantee an exact channel-pin match.

### DIFF
--- a/engine/zed/Cargo.lock
+++ b/engine/zed/Cargo.lock
@@ -474,6 +474,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_spanned"
+version = "0.6.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf41e0cfaf7226dca15e8197172c295a782857fcb97fad1808a166870dee75a3"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "simd-adler32"
 version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


sxlijin:
Dammit. Thanks for flagging this for us, need to figure out what the best way to handle this is. The new baml-language toolchain is not ready for prime time yet, and should not impact the Zed extension.

sxlijin:
#3853 is up to fix this

j-hartshorn:
Thanks for the attention on this one, and the fix for the underlying issue (much more simply than this PR!)

PR #3849, opened 2026-06-25
feat(testing): migrate testset test runner from Rust into BAML stdlib

antoniosarosi:
## What

Migrates the `baml test` **testset** driver out of Rust and into the `testing` stdlib (Linear **B-255**). `TestRegistry.run_filtered(include[], exclude[])` (plus `list_filtered` for `--list`) now does discovery → lazy expansion → filtering → concurrent leaf execution (`spawn`/`await`) → **runner-honoring** aggregation, returning a single flat `FlatTestReport`. The Rust driver (`test_command.rs`) shrinks to compile → engine → one call → print.

> **Reconciled with #3855** ("Honor testset runners for filtered tests"), which merged into `canary` mid-review and overlaps these files. This PR **adopts #3855's behavior** while keeping the BAML-first architecture (the tolerated-failure classification was moved out of #3855's Rust and into BAML). See the **[design note](https://github.com/BoundaryML/baml/pull/3849#issuecomment-4810058691)** for the full architecture decision (Option A — BAML owns orchestration, Rust is a thin presenter) and why Option B is structurally blocked by user-authored runners.

## Behavior

- **Both filtered and unfiltered** runs honor testset runners (`PassRate`, `Quorum`, `Retry`); the exit code follows the aggregate outcome. *(Previously, filtered runs bypassed runners — that changed with #3855 and this PR matches it.)*
- **Tolerated failures**: a failing leaf whose enclosing testset runner still passes (e.g. under `PassRate`) is counted separately as *tolerated* — kept out of the hard `failed` count and does **not** fail the command. The output makes this explicit: `PASS testing::* [outcome=pass; N tolerated failure(s)]` and `aggregate passed — X passed, N tolerated failure(s), Z total`.
- Per-testset expansion-failure containment: a testset whose generator body throws *or* panics becomes one error — named `suite/(failed to expand)` so `-i suite::*` / `-x suite::*` still select/exclude it — while the other tests keep running.

## Changes

- **`testing/registry.baml`** — `run_filtered`/`list_filtered`; glob filtering (`select_names` + helpers, mirroring `TestFilter`); `run_selected` honors runners over the selected subset; `flatten_with_tolerated`/`count_leaves` classify tolerated vs. hard failures **in BAML** (replacing #3855's Rust `extract_leaf_summary` tree-walk).
- **`testing/types.baml`** — `FlatTestReport { outcome, passed, failed, tolerated, total, failed_names[], messages[] }` (aggregate-only; `messages[]` carries every failing run's message for `Quorum`/`Retry`). The per-leaf `FlatTestResult` was dropped.
- **`baml_cli/src/test_command.rs`** — thin presenter (~875 lines, down from canary's 1287): reads one flat report and prints it; **no longer walks** the nested `TestReport | TestSetReport` tree across the FFI boundary. The legacy `test Foo { functions args }` LLM-test path + compile/engine setup stay in Rust.
- **`baml_cli/src/test_filter.rs`** — regex → literal `*`-glob, so legacy and testset tests filter identically and match the documented `*` wildcard contract (the old regex silently gave `.`, `(`, `[`, … regex meaning).

## Testing

- `exit_code_e2e`: **33/0** — incl. #3855's filtered-runner-honored, leaf-runs-under-parent-`PassRate`, and mixed tolerated-failures-kept-out-of-failed scenarios, plus unfiltered PassRate, aggregate-fail-with-zero-failed, and failed-child-name.
- `baml_cli` lib: **252/0** — filter glob parity against the old regex unit tests + flat-report parse/print.
- `__testing_std__` fixture compiles **diagnostics-clean**; HIR/TIR/MIR/codegen, `bytecode_format`, and `describe` snapshots updated.

## Deferred (follow-ups — not in this PR; tracked in the design note)

1. **Drop the stateful registry handle** — fold `collect_tests` into one stateless `run`/`list` call (touches the engine, not just the stdlib).
2. **Migrate the LSP** off the chatty `serialize`/`expand_set`/`run_test` protocol onto the flat boundary (until then the registry has two consumer contracts).
3. **Delete legacy `test { functions … args … }` + the Rust `TestFilter`** — collapses filtering to a single BAML implementation.

## Notes

- Adversarially reviewed; CodeRabbit findings addressed.
- `catch_all_panics` isn't wired into the parser yet, so panic containment uses the equivalent `baml.panics.Panic =>` arm (same as `run_test`). Wiring the keyword is a possible follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added include/exclude pattern filtering for the CLI `test` command, enabling leaf-only selection and `--list` output.
  * Test execution now uses a new flat report summary for clearer, consolidated PASS/FAIL results and messages (including tolerated outcomes), with legacy tests handled separately from testsets.

* **Bug Fixes**
  * Standardized testset expansion failure reporting (including panics) into consistent outcomes and error formatting.
  * Updated test filter matching to consistent, anchored `*`-glob semantics aligned across the CLI and testing layer.


antoniosarosi:
## Reconciled with #3855 + architecture decision (Option A: BAML owns orchestration)

This PR was rebased onto #3855 ("Honor testset runners for filtered tests"), which merged while this was in review and overlaps the same files. Rather than ship a half-migration, we picked a deliberate target architecture. Summary of the decision + what landed vs. what's intentionally deferred.

### The decision
There are only two clean targets for the test runner:
- **(A) BAML owns orchestration**, Rust is a thin presenter — one call in, one flat record out, the CLI never walks BAML object graphs.
- **(B) Rust owns orchestration**, BAML is pure leaf functions.

**(B) is structurally blocked:** `TestSetRunner`s (`PassRate`, `Quorum`, `Retry`) are *aggregation-over-children authored in BAML by users*. If Rust owned aggregation we'd have to relocate a user extension point into the host, or make leaves return subtrees (→ back to A). So with user-authored runners, **A is the only clean end state.** #3855 moved the other way (re-implemented the whole aggregation tree-walk + tolerated logic *in Rust*); this PR completes the BAML direction instead.

### What this PR does
- **Adopts #3855's behavior** (filtered runs honor testset runners; a failing leaf whose runner still passes is a *tolerated* failure, kept out of the hard `failed` count). Validated against #3855's e2e tests (`test_filtered_testset_run_honors_pass_rate_runner_for_selected_set`, `test_filtered_testset_leaf_runs_under_parent_runner`, `test_mixed_testset_run_keeps_tolerated_failures_out_of_failed_total`, unfiltered PassRate).
- **Keeps the BAML-first architecture.** `TestRegistry.run_filtered(include[], exclude[])` does glob-filter → `run_selected` (the runner-honoring path) → `flatten_with_tolerated`, returning a single flat `FlatTestReport`. The **tolerated-failure classification moved from Rust into BAML** (`count_leaves`/`flatten_with_tolerated`, the `tolerated_by_parent` walk) — replacing the Rust `extract_leaf_summary` tree-walk.
- **Result:** `test_command.rs` is a thin presenter again (~875 lines vs canary's 1287; **0** `class_matches`, the CLI no longer walks the nested `TestReport | TestSetReport` tree; **O(1)** FFI crossings for the testset path vs. the prior `O(P+L)` serialize/expand ping-pong).
- **Orthogonal fixes kept:** `test_filter.rs` regex → literal `*`-glob (the old regex gave `.`/`(`/`[` regex meaning and dropped invalid patterns); **expansion-panic containment** — a `baml.panics.Panic` arm on `testset_child`/`testset_child_selected` (the existing `catch_all` caught `throw`s but not panics, so a panicking testset body aborted the whole run).

### Deferred (follow-ups — not in this PR)
These are the remaining steps to make Option A *fully* clean. They're larger / out of scope here and tracked separately:
1. **Drop the stateful registry handle** — fold `collect_tests` into one stateless `run`/`list` call. Touches the engine (`bex_engine`), not just the stdlib.
2. **Migrate the LSP** (`bex_project/src/bex_lsp/multi_project`) off the chatty `serialize`/`expand_set`/`run_test` protocol onto the flat boundary. Until then the registry has **two consumer contracts** (CLI = one flat call; LSP = the old ping-pong).
3. **Delete legacy `test { functions … args … }` + the Rust `TestFilter`.** Legacy is on its way out; until it's removed it keeps a separate sequential Rust execution path and its own `TestFilter`, which is the one remaining spot where filtering exists in two languages. Deleting legacy is what finally collapses that to a single (BAML) filter implementation.

antoniosarosi:
**Re: the two findings on the selection path (both from #3855) — addressed in 948022607.**

**`testset_children_selected` (leaf-name vs. whole-testset selection) — fixed.** Removed the exact `_name_selected(ts.name, names)` branch for testsets and select purely by `_has_selected_descendant`. `select_names` returns leaf paths and a testset only ever contributes `name/…` descendants, so an exact `ts.name` match could only fire on a same-named top-level `test` (tests/testsets dedupe in separate collections) — wrongly running the whole testset. Descendant-only selection is also strictly more correct for the filter semantics.

**Expansion-error catch breadth (275–289) — fixed via the label, catch kept broad on purpose.** You're right that the catch covered the runner, so a panicking `TestSetRunner` was mislabeled `…/(failed to expand)`. I kept the broad `catch_all` deliberately: narrowing it (running `run_testset` outside the catch) would let a runner panic escape `run()` into the parent's `run_children_parallel`, which hits `baml.sys.panic("unexpected test child failure")` and **aborts the entire run** — trading a contained, if mislabeled, error for a whole-run crash. The defect was the *label*, not the containment, so `expansion_error_report` now uses the generic `…/(testset error)`. `collect_subtree_names`, which is purely an expansion step, keeps the precise `…/(failed to expand)`.

PR #3904, opened 2026-07-02
Fix generic match exhaustiveness so `null` can't leak into a non-nullable `T` (B-633)

antoniosarosi:
> ⚠️ **ON HOLD — do not merge / enqueue.** Blocked on @2kai2kai2's TIR type‑algebra + interface‑soundness rework, which is an explicit **ongoing multi‑PR series** (#3896: *"this PR is part of an ongoing process so further PRs will be forthcoming"*, runtime *"partially uses the canonical type algebra"*; #3852: interface‑soundness code *"currently mostly unused"*). This PR changes **TIR exhaustiveness** (`union_targets_for_pattern`) **and** relies on **reified interface membership** (`type_implements`) — both squarely in that rework's path. It must land only **after** the series completes, then be **rebased and re‑verified** (the interface special‑casing here may be subsumed by the unified interface soundness). Converted to Draft until then.

## 📍 Status — where we stopped (resume from here)

The approach is **CodeRabbit‑approved**. Pushed commit `ea1446c53` implements the two‑part fix (documented below). **Two items remain**, both saved as uncommitted WIP in the worktree `~/orca/workspaces/baml/antonio-b-633`:

1. **Closure‑value edge in the reified `is_type` test — WIP fix on disk, not yet committed / verified.**
   Running the *full* CI suite (not just `baml_tests`) surfaced a real panic: `value_matches_reified_ty` reused `value_concrete_runtime_ty`, which is interface‑**receiver‑only** and `unreachable!`‑panics on Function/Closure/Future values (`bex_vm/src/vm.rs` ~1996). Two `bex_engine` tests hit it — `while_let_closure_captures_fresh_cell_per_iteration` and `test_closures_in_loop_vars` — because a **closure value reaches the reified test**.
   WIP fix (uncommitted, `vm.rs` +46/−7): make the test **total** via a new `is_reified_type_subject` guard (only primitives / class‑instances / enum‑variants / `type` / array / map are valid `value_concrete_runtime_ty` subjects) plus type‑tag checks for `Function`/`Future` targets.
   ⚠️ **Still needs verification** that this makes the 2 tests **pass with correct results** (not merely stop panicking) — i.e. confirm the closure reaching the test is a legit *refutable* `let x: T`, and not an *irrefutable* bind that shouldn't emit a type test at all (if the latter, gate it out in lowering instead of tolerating it in the VM).

2. **`baml_cli` describe snapshot regen.** The stdlib `iter.baml` change (added `null => Done {}` arm to `ArrayIterator.next`) shifts the `describe baml` listing; `baml_cli__describe_command_tests__render_builtin_package_listing.snap` needs regenerating (a `.snap.new` is staged).

**Resume checklist (once Kai's series lands):**
- Rebase `antonio/b-633` onto the new `canary`; reconcile `union_targets_for_pattern` with the reworked TIR (the fix may simplify or partly disappear).
- Finish item 1; confirm the 2 closure tests pass *with correct results* + soundness (`E0062`, no null‑leak) + no‑regression (`iter_array_nullable_elements` → `[3,1,4,3]`, `iter_flatten` → `[1,2,3,4]`).
- Run the **full** CI locally: `cargo nextest run --all-features -E 'not package(baml_tests) and not package(/^sdk_test_/) and not binary(=pack_e2e)'` **and** `cargo test -p baml_tests -- --skip parser_stress` → 0 failed; regen `baml_cli` (+ any flagged) snapshots; `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo fmt`.
- `gh pr ready 3904` (un‑draft) and enqueue.

---

## Bug

A generic `match` arm `let x: T` over a union scrutinee `T | <concrete>` (where `T` is a rigid type variable) was treated as covering the concrete member too. So this was wrongly deemed **exhaustive** and compiled, letting `null` bind into a non-nullable `T` at runtime:

```baml
function bad_pop<T>(b: Box<T>) -> T {
    match (b.items.pop()) {   // .pop() : T | null
        let x: T => x         // no `null` arm, yet accepted
    }
}
```

The concrete control `match (v: int?) { let x: int => x }` correctly errored `E0062`. The bug was specific to a type-variable arm. It also produced a **symmetric false `E0063` "unreachable arm"**: in `match (v: T?) { let x: T => …, null => … }` the explicit `null =>` arm was reported unreachable.

(Note: the ticket title "pop() erases the ?" is inaccurate — `pop()`/`.at()` correctly return `T | null`; the defect was in exhaustiveness/reachability dispatch.)

## Root cause

**Static** — `union_targets_for_pattern` (`baml_compiler2_tir/src/builder.rs`) decides which union members a pattern targets via a *symmetric* structural overlap, and `atoms_overlap` reports a `TypeVar` as overlapping *everything*. So a bare `let x: T` pattern claimed the concrete `null`/`string` member, making the match exhaustive and shadowing the explicit concrete arm.

**Runtime** — once `let x: T` no longer over-claims (so it is not the exhaustive-last catch-all), its runtime type test was emitted, but a bare `TypeArgRef` fell through to a constant-`false` in codegen (`emit.rs`, "these don't arise from pattern matching today"). With the test always false, the `let x: T` arm never matched — which would break every array iterator (`ArrayIterator.next` relied on the old catch-all).

## Fix

1. **Static (directional overlap)** — `builder.rs`: a bare top-level `TypeVar` *pattern* claims only the union's `TypeVar` member(s), never a concrete member. Mirrors the existing member-side rule. This restores `E0062` and removes the false `E0063`.
2. **Reified runtime test** — a `let x: T` arm now resolves `T` against the frame's realized type args and tests the value against the resolved type, reusing the machinery B-634 (#3902) introduced for class match arms:
   - `lower.rs`: route a bare `let x: T` through `tir2_to_dispatch_guard_template` (covariant `TypeArgRefOrWildcard`), the same path as the `Class<T>` arm.
   - `emit.rs`: emit a reified `is_type` (a `Type` constant carrying the template) instead of constant-false.
   - `vm.rs`: `IsType` substitutes the template from `frame.type_args` and runs an **interface-aware** value-membership test (`is_subtype_of` for concrete/union targets; `type_implements` when `T` resolves to an interface).

This single approach fixes **both** the soundness hole **and** keeps generic iteration working:
- `is_type(null, int?)` → **true** → `let x: T` binds it (nullable `T` yields its stored `null`).
- `is_type(null, int)` → **false** → falls to the `null` arm (no leak into `int`).
- `is_type(ArrayIterator, Iterable)` → **true** via `type_implements` (a class value matches an interface `T` it implements) → `flatten`/interface-typed iterators keep working.

The one stdlib change: `ArrayIterator.next` gains the now-required `null` arm (unreachable at runtime — stored nulls match `let value: T`).

## Tests

- `diagnostic_errors/generic_typevar_match` — `E0062` for `let x: T` over `T?`/`T | string` via `.pop()`, `.at()`, bare `T?`, and `T | string`.
- `compiles/patterns_new` — both arm orders (`let x: T` first and `null` first) and `Done`-first compile with no `E0062`/`E0063`.
- `ns_patterns_new_runtime` e2e — both arm orders + `.pop()`/`.at()` round-trip; `null` routes to the `null` arm for non-nullable `T` and binds to `let x: T` for nullable `T`.
- `match_union_typevar` — updated the reachability regression to assert concrete arms after a bare `let x: T` stay reachable.

## Before / after

| case | before | after |
| --- | --- | --- |
| `match (v: T?) { let x: T => x }` | compiles (unsound) | `E0062` missing null |
| `match (v: T?) { let x: T => …, null => … }` | false `E0063` | accepted, both orders |
| `iter_array_nullable_elements` (`int?[]`) | regressed to end-early | `[3, 1, 4, 3]` |
| `iter_flatten` (interface-typed) | regressed to `[]` | `[1, 2, 3, 4]` |

## Verification

- `cargo test -p baml_tests -- --skip parser_stress` → 2796 passed, 0 failed (incl. B-634's `ns_generic_match_subtype` / `generic_match_typevar_arm`).
- `cargo clippy --all-targets --all-features -- -D warnings` clean on `baml_compiler2_tir`, `baml_compiler2_mir`, `baml_compiler2_emit`, `bex_vm`.
- Rebased onto current `canary` (after B-634 #3902 and B-611 #3903).

https://linear.app/boundaryml2/issue/B-633

## Summary by CodeRabbit

* **New Features**
  * Improved runtime matching for generic type variables, so `let x: T` now works correctly alongside `null` and other concrete cases.
  * Added support for resolving reified type checks during runtime dispatch.
* **Bug Fixes**
  * Fixed cases where empty arrays, out-of-bounds access, or nullable inputs could route to the wrong match arm.
  * Preserved reachability of concrete match arms when a generic arm is present.
* **Tests**
  * Added regression coverage for generic pattern matching, nullable unions, and exhaustiveness behavior.


antoniosarosi:
Parking as draft — holding until @2kai2kai2's type-algebra / interface-soundness rework series lands. This is a TIR exhaustiveness fix (`union_targets_for_pattern`) plus a reified runtime `is_type` with interface membership (`type_implements`), and that overlaps the ongoing rework (#3896: "further PRs will be forthcoming"; runtime "partially uses the canonical type algebra"; #3852 interface-soundness code "currently mostly unused"). Once the series completes I'll rebase onto the new TIR and re-verify — the runtime interface special-casing here may be subsumed by the unified interface soundness. Approach is CodeRabbit-approved; only remaining local work is a closure-value edge in the reified test + a `baml_cli` describe snapshot regen.

PR #3905, opened 2026-07-02
fix: correct React streaming hook data types for nullable fields

### LINKED ISSUE #2622: [bug] React hooks with streaming always return non-nullable types
### Product

BAML

### Describe the bug

From the docs:

Image

But the React hooks are always returning the non-nullable types.

### Reproduction Steps

Given the following BAML:

```baml
class MessageTest {
  // MessageTest won't stream until type is known
  type "error" | "success" | "info" @stream.not_null

  // MessageTest will only stream when title is known
  title string @stream.done @stream.not_null

  // Timestamp will only appear when fully complete, until then it will be null
  timestamp string @stream.done

  // will stream token by token, and include completion state
  content string @stream.with_state

  // will stream token by token
  other string
}

function GetMessageTest(input: string) -> MessageTest {
  client CustomGPT5
  prompt #"
  "#
}
```

It generates two types:

```ts
// types.ts
export interface MessageTest {
  type: "error" | "success" | "info"
  title: string
  timestamp: string
  content: string
  other: string
  
}

// partial_types.ts
    interface MessageTest {
      type: "error" | "success" | "info"
      title: string
      timestamp?: string | null
      content?: StreamState<string | null>
      other?: string | null
    }
```

Then the hook's return type is:

```ts
// const data: {
//     type: "error" | "success" | "info";
//     title: string;
//     timestamp: string;
//     content: string;
//     other: string;
// } | undefined
// const data: MessageTest | undefined

const { data } = useGetMessageTest({ stream: true });
```

Shouldn't it be `partial_types.MessageTest`? Because it is actually returning null in the nullable fields, which is causing `Cannot access property on ...` bugs.

### BAML Version

0.211.2

### Language/Framework

React

### LLM Provider

None

### LLM Model

_No response_

### Operating System

macOS

### Browser

Firefox

### Code Editor

VS Code

mvanhorn:
## Summary
The generated React hooks typed the streaming `data`/`onData` values as `FinalDataType | StreamDataType`. During streaming the value is always the (partial) stream type, so the union forced consumers to narrow against a `FinalDataType` that never occurs mid-stream, and nullable fields surfaced incorrectly.

## Changes
When `stream` is not `false`, type `data` and `onData` as `StreamDataType<FunctionName>` rather than the union. Also set the render package to `baml_client.react` while generating the server streaming types so type references resolve in the right namespace, restoring the previous package afterward. Regenerated integ-test client reflects the new types.

## Testing
Added `useAaaSamOutputFormat.streaming.test.tsx` asserting the streaming data type.

Fixes #2622

## Summary by CodeRabbit

* **Bug Fixes**
  * React hook TypeScript typings now align precisely with the selected `stream` mode: streaming uses partial types for both `onData` and returned `data`, while non-streaming uses the final result type (removing the previous partial-or-final union).
  * Updated generated React server streaming type definitions to reference `partial` payload types across many schemas and test outputs.
* **Tests**
  * Added TypeScript compile-time assertions covering streaming vs non-streaming hook output shapes and `onData` callback typings.


aboqasem (approved):

aaronvg:
neat! Thanks Matt :)

mvanhorn:
Appreciate it, @aaronvg - the streaming hook types match nullable fields properly now.

aaronvg:
@mvanhorn how are you using BAML? Curious

mvanhorn:
Thanks for the merge, @aaronvg! To answer your question: I'm not running BAML in production - I ran into the nullable-fields typing gap while kicking the tires on the React streaming hooks for AI side projects, and it was annoying enough to fix properly. Really like the structured-output approach, so I'll keep poking at it.

PR #3915, opened 2026-07-04
Fix Windows LSP/playground project path handling

jubjub727:
**Summary**
- Normalize Windows drive-letter paths before converting them into BAML VFS paths.
- Fix native source discovery so VFS-style Windows globs load all project `.baml` files from disk.
- Add Windows regression coverage for URI-style drive paths, backslash paths, parent traversal, and source glob loading.

**Codex Comment**
The bug was caused by Windows paths crossing between three path formats: native paths like `D:\...`, URI/VFS-style paths like `/d:/...`, and mixed forms like `\d:\...`. BAML's VFS path logic uses `/` as its separator, so raw backslash paths could not walk to parent directories correctly. Separately, `/d:/...` could be treated as relative and joined against the server cwd, producing duplicated paths like `/d:/d:/...`.

The playground/LSP source loader also built VFS-style globs such as `/d:/project/baml_src/**/*.baml`, then passed them directly into native filesystem matching on Windows. That meant sibling files like `clients.baml` could be skipped, leaving only the open in-memory file loaded. In practice, valid client declarations appeared as unresolved names because their defining file was absent from the project view.

This fix keeps the translation at the filesystem/VFS boundary: Windows-only normalization before VFS conversion, and Windows-only conversion between VFS drive globs and native disk paths during source discovery. Non-Windows behavior is left unchanged.

**Author Comment**
I used Codex to investigate the root cause of these issues. Its understanding of the issue looks accurate on the surface and the fix appears to work on my machine. I haven't manually confirmed the actual root cause analysis myself but its work looks good to me.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform glob and file lookup behavior by normalizing platform-specific paths before matching.
  * Ensured Windows drive-root and separator formats are interpreted correctly for VFS path resolution.
  * Returned matching entries using consistent, VFS-friendly path formatting (forward slashes and correct leading `/` handling).
* **Tests**
  * Added Windows-focused coverage for drive-path and separator edge cases, plus a non-Windows sanity check.

### DIFF
--- a/baml_language/crates/baml_lsp_server/src/native_vfs.rs
+++ b/baml_language/crates/baml_lsp_server/src/native_vfs.rs
@@ -3,7 +3,7 @@
 //! Wraps `vfs::PhysicalFS` and implements `BulkReadFileSystem` by walking
 //! only the glob's base directory (not the entire filesystem).
 
-use std::io::Read;
+use std::{borrow::Cow, io::Read};
 
 /// Wrapper around `vfs::PhysicalFS` that implements `BulkReadFileSystem`.
 #[derive(Debug)]
@@ -74,8 +74,9 @@ impl bex_project::BulkReadFileSystem for NativeVfs {
         // Extract the base directory from the glob (everything before the
         // first `*` or `?` wildcard). This lets us walk only the relevant
         // subtree instead of the entire filesystem.
-        let base_dir = glob_base_dir(glob);
-        let pattern = glob_to_regex(glob);
+        let native_glob = native_glob_for_fs(glob);
+        let base_dir = glob_base_dir(&native_glob);
+        let pattern = glob_to_regex(&native_glob);
 
         let base = std::path::Path::new(&base_dir);
         if !base.is_dir() {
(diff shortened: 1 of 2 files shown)


sxlijin (commented):
Ahh, so `vfs` requires Unix-style paths. That's annoying - doing a quick skim it looks like https://docs.rs/path-slash/latest/path_slash/ might be the best crate to use for this?

I don't love the idea of reimplementing windows path normalization multiple times; I'd rather have a new crate `baml_language/crates/utils_windows/` for that purpose (will need a cargo-stow update to make `utils_` allowed as well).

jubjub727:
Absolutely agree. The only reason I made this fix is because I wanted to get it working on my machine so I could go back to working on the project I'm making with BAML. I mainly made this PR as a way of communicating the specific behaviour I was experiencing and the potential root cause behind it rather than as a serious long term proposal for fixing these path bugs. I'm also very thankful that you were able to review this and find a relevant library here as that gives me a very easy starting point and clear direction to go from.

As I mentioned in discord I think a more detailed dive into not just this specific path handling but instead how paths are handled throughout the entire project including the various layers not just Rust is the most appropriate path forward. I'm happy to go off, figure this all out and come back to you with a detailed report of my findings and how I'm thinking of solving it more widely if that sounds good to you.

As much as I feel like I'm perfectly capable of solving a task of this scope on my own I am also cognizant that I'm unfamiliar with the nuances of the code base and how that can lead to accidentally creating a bunch of work for your team later on. Programming languages are hard and deserve patience, respect and discipline imo :)

jubjub727:
Closing this now. The better solution was included in the first batch of fixes here: https://github.com/BoundaryML/baml/pull/4103

PR #3923, opened 2026-07-06
Make Server.serve honor task.cancel() (B-681)

antoniosarosi:
## What

`baml.http.Server.serve`'s accept loop now `select!`s on the future's own cancel token, so a `task.cancel()` on a spawned `serve` stops the accept loop cooperatively and settles `Cancelled` — token-based, rather than relying solely on the sys-op dispatcher dropping the still-pending future.

```rust
let accepted = tokio::select! {
    biased;
    // B-681: observe the future's own cancel token directly.
    () = cancel.cancelled() => {
        return Err(VmRustFnError::Panic(VmPanic::Cancelled));
    }
    result = state.listener.accept() => result,
    Some(_) = conns.join_next(), if !conns.is_empty() => continue,
};
```

On cancel, returning from `serve` drops `conns` (aborting in-flight connection tasks via `JoinSet`'s `Drop`) and `_serving_guard` (releasing the "serving" flag so the same `Server` can be re-served). The listener stays bound (it lives in the `Server`), so the port is retained for a restart. `biased` checks cancel first on every poll, so a hot accept loop can't starve it.

## How `sleep`'s pattern was mirrored — and an important finding for B-682

`baml.sys.sleep` does **not** observe a cancel token itself. Its cancel-awareness is entirely **engine-wrapped**: the engine races *every* async sys-op's future against the thread cancel token in a biased `tokio::select!` at `bex_engine/src/lib.rs:4423-4427`, and drops the still-pending future when cancel wins. `task.cancel()` fires that exact token (`FutureManager::cancel_future` → `Future::settle_cancelled` → `self.cancel.cancel()`, `bex_vm_types/.../future.rs:450`), and `serve` receives a clone of the same token as its `cancel` argument.

That means **`serve` already honored `task.cancel()` on canary** via the engine's select — verified empirically: after `task.cancel()`, a fresh request to the still-bound port times out (the loop stopped accepting) rather than being serviced. This PR makes `serve` observe its own token **cooperatively** (the explicit, per-op form of that mechanism) instead of depending only on the dispatcher dropping it. On canary the engine's biased cancel arm still wins the outer race and drops the future first, so behavior is unchanged and all existing tests stay green; the change hardens `serve` to be self-sufficient and encodes the token-based-cancellation contract in the op.

This is the tactical, per-op stopgap. Because the general mechanism is already engine-level (the `lib.rs:4423` select), the proper central solution is tracked in **B-682**, which can supersede / remove this per-op arm.

## Test

Added `server_serve_honors_cancel` to `ns_http_server/http_server.baml`: `spawn { server.serve(...) }`, `fetch` (proves the server came up), `task.cancel()`, then `(await task) catch (Cancelled) { 0 }` — asserting the awaiter observes `Cancelled` promptly (not a hang). Mirrors `cancel_from_handle_then_await_catches_cancelled`.

## Verification

- `cargo test -p baml_tests` — 2009 baml tests passed / 0 failed; bytecode snapshot updated for the `http_server` namespace (the added test function only; no stdlib snapshots moved).
- `cargo nextest run --all-features -E 'not package(baml_tests) and not package(/^sdk_test_/) and not binary(=pack_e2e)'` — 4152 passed / 0 failed / 33 skipped.
- `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- `cargo doc --no-deps --all-features` — succeeds.
- `cargo fmt` — no changes.

Fixes B-681. Note: tactical per-op fix; the general engine-level solution is tracked in B-682.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server shutdown behavior so cancellation is handled promptly instead of leaving the server hanging.
  * Confirmed the server can be reached before cancellation and then exits cleanly when stopped.
  * Preserved server restart readiness after cancellation by ensuring cleanup completes correctly.

### DIFF
--- a/baml_language/crates/baml_tests/baml_src/ns_http_server/http_server.baml
+++ b/baml_language/crates/baml_tests/baml_src/ns_http_server/http_server.baml
@@ -185,6 +185,26 @@ test "server_reserve_after_cancel" {
     assert.equal(server_reserve_after_cancel(), "first/second")
 }
 
+// A spawned `serve` honors `task.cancel()` (B-681): the accept loop selects on
+// the future's cancel token, so cancelling stops it and `await task` settles
+// `Cancelled` promptly (a catchable `baml.panics.Cancelled`) rather than hanging.
+// The `fetch` before the cancel proves the server actually came up first.
+function server_serve_honors_cancel() -> int {
+    let server = baml.http.Server.bind("127.0.0.1:0");
+    let task = spawn {
+        server.serve((req: baml.http.Request) -> baml.http.Response {
+            baml.http.Response.new(200, { }, "ok".to_utf8())
+        })
+    };
+    let _ = baml.http.fetch("http://" + server.addr + "/").text();
+    let _ = task.cancel();
+    (await task) catch (e) { baml.panics.Cancelled => 0 }
+}
+
+test "server_serve_honors_cancel" {
+    assert.equal(server_serve_honors_cancel(), 0)
+}
+
 // A bound method used as the handler: `server.serve(greeter.handle)`. Proves the
 // callable path injects the receiver as `self` (so `self.prefix` is visible).
 class Greeter {
--- a/baml_language/crates/baml_tests/snapshots/baml_src/http_server.snap
+++ b/baml_language/crates/baml_tests/snapshots/baml_src/http_server.snap
@@ -51,26 +51,32 @@ function http_server.$init_test_ns_http_server_http_server(registry: testing.Tes
     call testing.TestCollector.register_test
     pop 1
     load_var registry
-    load_const "server_bound_method_handler"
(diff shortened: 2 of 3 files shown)


antoniosarosi:
Closing unmerged — superseded by the B-650 investigation. B-681's premise (that `serve` ignores `task.cancel()`) turned out to be incorrect: `serve` already honors cancellation via the engine-wrapped biased `select!` (bex_engine/src/lib.rs ~4423) that races every async sys-op against the thread cancel token (the same mechanism `sleep` relies on). This cooperative per-op arm is behavior-unchanged, and the actual B-650 fix (#3925 — gate the end-of-run wait on the true top-level root) did not need it. See B-682 for why the central mechanism already exists. No behavior lost.

PR #3927, opened 2026-07-07
Fix baml test hang: terminate .env self-referential variable expansion

hellovai:
## What

`baml-cli test` (and any command that loads a `.env`) could hang, ballooning memory to tens of GB, when the `.env` file contained a **self-appending variable** such as:

```
PYTHONPATH=$PYTHONPATH:./src
PATH=$PATH:/usr/local/bin
LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/opt/lib
```

Symptom reported: the `Loading environment variables from .env file` warning prints, then the run hangs with **no further output**, RAM climbing to ~57 GB. Present in older versions too (the warning line is new in 0.223; the expansion bug predates it).

## Why

`expand_variables` ran `expand_value` to a fixpoint (`while changes_made`), resolving `$VAR` against the file's **own current value**. A self-reference never reaches a fixpoint — each pass copies the value into itself, so its length **doubles every pass** (~2^n):

```
FOO=x$FOO
pass 30: 1.07 GB
pass 34: 17 GB
pass 35: 34 GB
pass 36: 68 GB  -> thrash / OOM, process wedged
```

Because the pip `baml-cli` runs the Rust CLI in-process via PyO3 (`baml_py:invoke_runtime_cli`), the ballooning process shows up as `python3.10`, and it never reaches any output-producing code — matching "killing it ends the run with no output".

## Fix

Rewrite expansion to standard `.env`/shell semantics: resolve each variable **once** (memoized), with transitive resolution of file-to-file references (`NESTED_VAR -> LOGS_DIR -> BASE_DIR` still works), and break self-reference cycles by resolving against the **process environment only**, never the variable's own in-file value.

- `PATH=$PATH:/foo` now appends to the real `PATH` (the intended behavior).
- Cycles (`A=$B`, `B=$A`) terminate in one bounded pass.

Also collect `content.lines()` once in `load_env_from_string` instead of `.nth(i)`/`.count()` per iteration, removing a latent O(n^2) parse (adjacent, not the hang).

## Tests

New regression tests in `dotenv/tests.rs`:
- self-append terminates and appends to the process value
- self-append with no process value stays literal (no growth)
- an `A<->B` cycle returns instead of hanging

All existing dotenv tests (interpolation, transitive `NESTED_VAR`, complex file, system env) still pass.

## Workaround for affected users
Remove the self-referencing line from `.env` (or any `.env` in a parent dir — the loader walks up to `/`), or run `baml-cli test --dotenv false`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `.env` variable interpolation to safely resolve self-references and cyclic references without hanging or unbounded expansion.
  * Missing variables now remain as the original `$VAR`/`${VAR}` text, while process environment values are used when available.
  * Optimized multiline `.env` parsing for more reliable line handling.
* **Tests**
  * Added regression tests for self-referential, unset, and two-variable cyclic interpolation scenarios.

### DIFF
--- a/engine/baml-runtime/src/cli/dotenv/mod.rs
+++ b/engine/baml-runtime/src/cli/dotenv/mod.rs
@@ -1,7 +1,7 @@
 mod tests;
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     env,
     fs::File,
     io::{self, BufReader, Read},
@@ -65,10 +65,13 @@ pub fn load_env_file(path: PathBuf) -> Result<HashMap<String, String>> {
 
 pub fn load_env_from_string(content: &str) -> Result<HashMap<String, String>> {
     let mut env_vars = HashMap::new();
+    // Collect the lines up front: indexing into `content.lines()` on every
+    // iteration would make this O(n^2) in the number of lines.
+    let lines: Vec<&str> = content.lines().collect();
     let mut line_index = 0;
 
-    while line_index < content.lines().count() {
-        let raw_line = content.lines().nth(line_index).unwrap();
+    while line_index < lines.len() {
+        let raw_line = lines[line_index];
         line_index += 1;
 
         let line = raw_line.trim();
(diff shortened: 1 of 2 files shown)


hellovai:
Addressed the review comment: fixed the misleading doc comment on `test_self_reference_without_process_value_terminates` — it now says the unset self-reference is left **literal** (`lookup_var` returns `None`), matching the assertion, instead of the incorrect "resolves to empty". Also applied rustfmt. Rebased onto latest canary.

PR #3940, opened 2026-07-08
fix(jinja): render map<K,V> as a JSON object shape in ctx.output_format (B-630)

antoniosarosi:
## Problem

A `map<string, T>` field rendered as the raw BAML type syntax inside the
`${ctx.output_format}` prompt hint:

```baml
class Review { scores: map<string, int> }
```

produced

```
scores: map<string, int>,
```

Since the model must answer with **JSON**, leaking the literal BAML type gives
it no example of the object shape it has to emit.

## Fix

The default map rendering is now a JSON object shape, consistent with how
classes (`{ field: T }`) and lists (`T[]`) already render:

```
scores: { "<string>": int },
```

In `baml_language/crates/sys_llm/src/types/output_format.rs` (the maintained
output_format renderer):

- The default `map_style` is now `ObjectLiteral` instead of `TypeParameters`.
- `ObjectLiteral` renders `{ "<key>": value }` (a quoted placeholder key that
  reads as "any string key") instead of the ambiguous `{ [key: K]: V }`.
- `map_style='type_parameters'` remains available as an opt-in escape hatch that
  keeps the literal `map<K, V>` syntax.

> Note: `engine/` is legacy/deprecated and intentionally not modified — only
> `baml_language/` is maintained.

## Tests

- Added `test_render_class_with_map_field_object_shape` (the exact ticket repro:
  `class Review { scores: map<string, int> }`) and
  `test_render_map_type_parameters_opt_in` (escape-hatch regression guard).
- Updated existing `sys_llm` map render assertions to the new object shape.

All `sys_llm` tests pass; `cargo fmt --check` and `cargo clippy` clean on the
touched crate.

Ticket: https://linear.app/boundaryml2/issue/B-630

Fixes B-630

## Summary by CodeRabbit

* **Bug Fixes**
  * Map schemas now default to a JSON-object-style representation, improving the readability of generated output.
  * Only the explicit `type_parameters` option now uses the alternate angle-bracket map format; all other values fall back to the standard object style.
  * Updated recursive and nested map outputs to match the new format consistently.

* **Documentation**
  * Clarified the available map rendering options and the default behavior in comments and examples.

### DIFF
--- a/baml_language/crates/sys_llm/src/jinja/output_format_object.rs
+++ b/baml_language/crates/sys_llm/src/jinja/output_format_object.rs
@@ -194,8 +194,8 @@ fn parse_hoist_classes_kwarg(kwargs: &Kwargs) -> Result<HoistClasses, minijinja:
 }
 
 /// Parse `map_style` kwarg:
-/// - Not present -> `TypeParameters` (default)
-/// - "`type_parameters`" -> `TypeParameters`
+/// - Not present -> `ObjectLiteral` (default JSON object shape)
+/// - "`type_parameters`" -> `TypeParameters` (opt-in `map<K, V>` escape hatch)
 /// - "`object_literal`" -> `ObjectLiteral`
 fn parse_map_style_kwarg(kwargs: &Kwargs) -> Result<MapStyle, minijinja::Error> {
     if !kwargs.has("map_style") {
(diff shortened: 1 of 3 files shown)


antoniosarosi:
@coderabbitai I've addressed the cross-layer consistency note: the new compiler's renderer (`baml_language/crates/sys_llm/src/types/output_format.rs`) was a port of the engine renderer and carried the same B-630 bug. It now defaults to `ObjectLiteral` and renders `{ "<string>": int }` byte-for-byte identical to the engine, preserving the byte-equivalence invariant documented in `llm_types.baml`. `map_style='type_parameters'` remains the opt-in `map<K, V>` escape hatch. Both layers are now aligned. Please re-review.

PR #3969, opened 2026-07-10
refactor(lsp): Phase 0 latency and mutex containment

rossirpaulo:
## Summary

Implements Phase 0 of the LSP latency and mutex containment design (`docs/design/lsp-latency-and-mutex-fix.md`):

- **Source revision and owned candidates** — monotonic `SourceRevision`, detached `CompilationOutcome` / `DiagnosticCandidate` / `EngineCandidate`, and revision-conditional engine commits so superseded rebuilds change no runtime state.
- **Coherent runtime state** — unified `RuntimeState` with collection epoch fencing, registry leases, and atomic run snapshots (`prepare_function_run`).
- **Durable versioned diagnostics** — `DiagnosticsFence` with compare-and-clear dirty tracking, stale-candidate discard, and bounded busy retry on native.
- **Bounded request waits and typed LSP errors** — 1s request-lane deadline maps to `ContentModified` / `RequestFailed` / `InternalError` instead of generic `-32001`.
- **Position encoding negotiation (C1)** — new `PositionCodec` for UTF-8/UTF-16 at the LSP boundary.
- **Debounced native tails (B3)** — single-flight engine rebuild and diagnostics/project-update debouncing.
- **Salsa memoization (B1)** — `file_annotations` and `semantic_tokens` as tracked queries.
- **Deferred profiling activation** — winning engine commit only activates profiling lifecycle.

Draft for early review and comparison with an alternate approach. Further performance work (snapshot reads off the source gate, startup prewarm) is planned after measurement.

## Test plan

- [x] `cargo test --lib -p bex_project`
- [x] `cargo test --lib` (workspace; SDK harness failures pre-existing)
- [x] `cargo check --target wasm32-unknown-unknown -p bridge_wasm`
- [x] `cargo clippy` (pre-commit hook)
- [ ] Extension smoke test against release `baml-cli` on sandbox project
- [ ] Compare startup `-32803` behavior vs alternate PR (later)
- [ ] Rerun `/tmp/baml-inlay-repro` harness after merge candidate selection

## Summary by CodeRabbit

* **Improvements**
  * Enhanced language-server responsiveness with debounced refresh of diagnostics and rebuild work while editing.
  * Improved correctness of positions, diagnostics, semantic tokens, inlay hints, code actions, and navigation across UTF-8/UTF-16 clients.
  * More reliable run launching and cancellation so actions target the correct project state; improved LSP error-code handling and added support for common notifications.
  * Smoother profiling lifecycle behavior for inactive/discarded runs.
[...]

PR #3976, opened 2026-07-11
feat(lsp): Phase 1 — bounded ingress, cancellation ownership, client gating

rossirpaulo:
**Stacked on #3969** (Phase 0). Do not merge before it; GitHub will retarget this to `canary` when #3969 lands. Together the two PRs deliver the LSP latency/mutex design (`docs/design/lsp-latency-and-mutex-fix.md`) through Phase 1; Phase 2 (salsa snapshot reads) remains behind the design doc's spike gate.

### Rust — bounded ingress and cancellation (Track B2/D2/D3)
- **IngressScheduler** (from the alternate implementation in #3961, relocated into `baml_lsp_server` where transport scheduling belongs): transport-neutral admission classes, lifecycle barriers, per-class byte/item budgets, response-ownership tokens, didChange tail-coalescing, browser takeover. Ported with fixes for every defect verified in review:
  - forced `NonCancellableWhenRunning` on `initialize`/`shutdown`/`executeCommand` can no longer be downgraded by a caller policy (`stronger_cancellation`);
  - a permanently oversized message is contained per-message — requests answer `-32803`, notifications drop with a warn; the session and process stay alive;
  - an oversized `$/cancelRequest` no longer livelocks the control lane;
  - pipelined `initialized` during `InitializeResponding` is admitted FIFO and accepted at dispatch (found via the latency harness; the reference dropped it, breaking pipelining clients).
- **LspRuntime**: stdio and browser `/api/lsp` sessions dispatch through one scheduler behind a process-owned mutex; `$/cancelRequest` is drained on the transport thread so it can claim a response while a handler runs — queued requests answer **`-32800`** immediately, running cancellable requests observe a dispatch-scoped token (checked at handler entry and immediately after the source guard; never connected to salsa unwinding). Browser takeover atomically revokes the old session's sink (tombstone), purges its pending responses, and replays document overlays.
- **Bounded output**: serialize-once `Arc<[u8]>` frames with a shared budget charge (accounting equals actual memory); per-session outbound budgets; `Lagged`/saturation close only the affected session; responses exceeding the reservation are replaced with a per-request `-32803`, never a session kill.
- **Bounded stdio framing**: 16 KiB header cap; >8 MiB bodies discarded in bounded chunks with a recoverable null-id error; abnormal exit paths return nonzero.
- **bex_project kept minimal**: `LspError::RequestCanceled` → `-32800` at the single serialization boundary (I7); connection-scoped LSP sessions (fresh position-encoding negotiation and workspace roots per connection, shared project registry) — fixes the global once-only encoding negotiation flagged in review.

### TypeScript — client gating (D6, D2 subset, Track A)
- **Playground run-gating**: when `is_bex_current` is false (or a run/preview was refused with `projectNotReady`), Run/test/preview actions disable with a "Preparing current build…" status and re-enable on the next current `ProjectUpdate`. No protocol version bump; closes the fail-closed D7 UX window #3969 introduces. Stale "using last successful build" copy removed.
- **WebSocketRuntimePort reconnect correctness**: fail-closed hello; bounded first-connect queue (64, drop-oldest); session-scoped commands issued while disconnected are dropped with synthesized rejections instead of replaying into a new session; a desired-runtime lease survives reconnects and is re-sent after every compatible hello (the reference's lease-loss bug is inverted and pinned by test).
- **Track A multi-root ownership**: `projectRoots.ts` (canonical/symlink-aware root resolution, nested projects, multi-workspace) + routing in the extension — excluding two review-verified reference bugs: closing the last `.baml` document no longer stops the owner server (kept the playground webview alive), and aggregate status preserves `error` when every client start fails.

## Known limits (deliberate)
- `ensureProjectRuntime`/`releaseProjectRuntime` client plumbing is dormant: typed, unit-tested, nothing sends them yet — the seam for the demand-gating server work (D5).
- A client that pipelines `didOpen` in the same instant as `initialized` (no ordering gap) can still have that notification dropped at admission — contract-conformant per the design doc; VS Code never does this.
- Oversized (>8 MiB) stdio requests answer with a null-id error (the id is unknowable without buffering the oversized body).
- The `pendingTestTarget` deep-link auto-run is converted to the preparing state during a rebuild window but not auto-retried.

## Validation
- `cargo test --lib -p baml_lsp_server` 67/67 (22 scheduler, 9 runtime, 7 framing/budget, plus existing) · `-p bex_project` 45/45 (+1 pre-existing ignored)
- `cargo clippy -p baml_lsp_server -p bex_project --all-targets -- -D warnings` clean · `cargo check --target wasm32-unknown-unknown -p bridge_wasm` clean · `cargo fmt --check` clean
- TS: playground 125/125, extension 13/13 (8 projectRoots), webview 19/19 (3 new gating), promptfiddle + all typechecks green
- Harness (`/tmp/baml-inlay-repro`, 400 fns, release build of this branch): `$/cancelRequest` now answers `-32800` (Phase 0: completed normally); typing inlay p50 2.0 ms / p95 51.7 ms; zero request errors; zero unanswered; 10-edit burst → 0 refresh storms.

https://claude.ai/code/session_01VEMxti5WQ8bCgJUuX9AQvo

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -1043,6 +1043,7 @@ dependencies = [
  "sys_native",
  "sys_ops",
  "sys_types",
+ "thiserror 2.0.18",
  "tokio",
  "tokio-tungstenite",
  "tower-http",
--- a/baml_language/crates/baml_lsp_server/Cargo.toml
+++ b/baml_language/crates/baml_lsp_server/Cargo.toml
@@ -54,6 +54,7 @@ reqwest = { workspace = true }
 rustls = { workspace = true, optional = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+thiserror = { workspace = true }
 tokio = { workspace = true, features = [ "sync", "rt-multi-thread", "net", "time", "macros" ] }
 tokio-tungstenite = { workspace = true }
 tower-http = { workspace = true }
--- a/baml_language/crates/baml_lsp_server/src/lib.rs
+++ b/baml_language/crates/baml_lsp_server/src/lib.rs
@@ -33,6 +33,8 @@
 //! `--all-features` (both enabled); prefer one when building the LSP binary.
 
 mod deadlock_watchdog;
+pub mod lsp_ingress;
+mod lsp_runtime;
 mod native_lsp_sender;
 mod native_vfs;
 pub mod playground_env;
@@ -47,6 +49,7 @@ pub mod playground_ws;
 use std::{
     collections::BTreeMap,
     fs,
+    io::{BufRead, Write},
     path::{Path, PathBuf},
     sync::Arc,
     time::{Duration, UNIX_EPOCH},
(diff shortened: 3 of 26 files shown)


rossirpaulo:
Superseded by #4000, which combines this PR with its base #3969 into a single PR against canary (with latest canary merged in and re-verified). All content of this PR is in #4000 unchanged apart from the canary reconciliation.

Issue #2666, opened 2025-10-27
[feat] add TOON to output formats


phischdev:
### Product

BAML

### Problem Statement / Use Case

AI is becoming cheaper and more accessible, but larger context windows allow for larger data inputs as well. LLM tokens still cost money – and standard JSON is verbose and token-expensive. 

### Proposed Solution

„Token-Oriented Object Notation is a compact, human-readable format designed for passing structured data to Large Language Models with significantly reduced token usage.“

It claims to be a drop-in replacement for JSON that typically uses 30–60% fewer tokens.

https://github.com/johannschopplich/toon

sneerin:
that would be awesome !

kienbui1995:
I think we should implement this soon to try to reduce token cost

aaronvg:
we are looking for contributors to drive this through.

We would need a spec of how to serialize BAML Types into this format in the prompt, like this document: https://gloochat.notion.site/XML-Outputs-218bb2d26216806fb4dfef242d150c51?source=copy_link 

https://crates.io/crates/toon-format/0.3.0

Then we can use an opensource crate to adapt our parser to turn the TOON output into json > back to the baml parser

mvanhorn:
Opened #3990 with a first pass at TOON output support: `ctx.output_format(format="toon")` renders a TOON schema in the prompt and the response is decoded back through the schema-aligned (`jsonish`) parser into the typed value. Built on the existing `toon` input filter (#2720). Happy to align the exact surface/spec with whatever @aaronvg had in the design doc.

PR #3990, opened 2026-07-12
feat(output-format): add TOON output format

mvanhorn:
## Issue Reference
- [x] This PR addresses #2666

@aaronvg mentioned on #2666 that you're looking for contributors to drive TOON support through, and asked for a spec of how BAML types serialize into the format in the prompt. This is a first pass at that. Happy to adjust the surface (client option vs. `ctx.output_format(...)` argument, spec details) based on what fits the design you had in mind.

## Changes
Adds TOON ([Token-Oriented Object Notation](https://toonformat.dev)) as a selectable output format:

- `{{ ctx.output_format(format="toon") }}` renders TOON-shaped schema instructions instead of JSON. `format` defaults to `"json"`, so existing prompts are unchanged.
- The `format` argument is validated at compile time in the jinja type checker (only `"json"` / `"toon"` accepted).
- Response parsing decodes TOON via the official `toon-format` crate and feeds it through the same schema-aligned (`jsonish`) coercion used for JSON, so preambles, indentation, tabular arrays, and recoverable array-length mismatches are all handled.
- Builds on the existing `toon` input Jinja filter (#2720) — this adds the output/parse side.

Rendering lives in `engine/baml-lib/jinja-runtime/src/output_format/`; parsing in `engine/baml-lib/jsonish/` (new `toon_parser.rs`). Docs updated in `fern/03-reference/baml/prompt-syntax/output-format.mdx`.

## Testing

- [x] Unit tests added/updated

Added round-trip tests covering class/enum/list/map/optional/union, TOON after a model preamble, and recoverable bad array lengths. Local runs:
- `jsonish`: 493 tests pass (incl. 10 new `test_toon` cases)
- `internal-baml-jinja`: 119 pass, `internal-baml-jinja-types`: 43 pass
- `cargo fmt --check`, `git diff --check`, and current-branch `baml check` all clean

## Screenshots

Simulated demo of the prompt-side contrast (JSON vs TOON schema for the same type) and the parse-back to a typed value:

TOON output format demo

## PR Checklist

- [x] I have made corresponding changes to the documentation

## Additional Notes

Adds one dependency (`toon-format` 0.5.0). The branch is a couple of commits behind current `canary`; I can rebase before merge. AI was used for assistance.

## Summary by CodeRabbit

* **New Features**
  * Added TOON as a supported option for `ctx.output_format` via a new `format` parameter.
  * Enabled schema and prompt rendering in TOON mode, including arrays, maps, enums, and nested classes.
  * Added TOON-aware response parsing and decoding, including recovery from preambles and formatting variations.
  * Extended validation to recognize `format="json"` and `format="toon"`.
* **Bug Fixes**
  * Improved TOON candidate extraction and array-length recovery.
* **Documentation**
  * Updated output-format documentation with TOON guidance.
* **Tests**
  * Added coverage for TOON rendering, parsing, decoding, and validation.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -1065,7 +1065,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
- "thiserror 2.0.12",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -1285,7 +1285,7 @@ dependencies = [
  "supports-color",
  "tempfile",
  "test-log",
- "thiserror 2.0.12",
+ "thiserror 2.0.18",
  "time",
  "tokio",
  "tokio-stream",
@@ -1618,7 +1618,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
- "thiserror 2.0.12",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
@@ -3238,7 +3238,7 @@ dependencies = [
  "rustls-pemfile 2.2.0",
  "serde",
  "serde_json",
- "thiserror 2.0.12",
+ "thiserror 2.0.18",
  "time",
  "tokio",
 ]
@@ -3259,7 +3259,7 @@ dependencies = [
  "rand 0.9.2",
  "serde",
  "serde_json",
- "thiserror 2.0.12",
+ "thiserror 2.0.18",
  "tokio",
 ]
 
@@ -3287,7 +3287,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
- "thiserror 2.0.12",
+ "thiserror 2.0.18",
  "time",
  "url",
 ]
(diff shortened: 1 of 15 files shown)


mvanhorn:
Both Major findings were already addressed in 35c8d2e; flagging it here since the review still reads as outstanding.

**Fence matching anchored to line starts.** `fenced_toon_blocks` no longer uses a bare `find("```")`. Pairing goes through `find_line_start_fence`, which only accepts a fence whose line contains nothing but whitespace before it, so a stray ``` in prose cannot open or close a block. Guarded by `ignores_inline_backticks_when_pairing_fenced_toon_blocks`, which feeds exactly that case and asserts the real TOON block still comes back.

**Full dedup, not just consecutive.** `candidates.dedup()` is gone; `deduplicate_candidates` uses a `HashSet` with `retain`, so all duplicates drop and first-seen order is preserved. Guarded by `deduplicates_all_candidates_in_first_seen_order` with a non-adjacent duplicate.

`cargo test -p jsonish --lib -- toon` is 12/12 locally, including both of those.

Issue #2409, opened 2025-09-03
[feat] Alternative structured output format (YAML, XML, TSV, CSV)


yyuyulm:
### Product

BAML

### Problem Statement / Use Case

Allow the generation of the structured output response to be other string-based formats such as YAML, TOML, XML, TSV, CSV, optionally. For example, YAML has better token efficiency than JSON, resulting in lower response time and better context utilization. TSV/CSV allows token-efficient generation of tabular data.

more:<br>[https://david-gilbertson.medium.com/llm-output-formats-why-json-costs-more-than-tsv-ebaf590bd541](<https://david-gilbertson.medium.com/llm-output-formats-why-json-costs-more-than-tsv-ebaf590bd541>)

[https://medium.com/better-programming/yaml-vs-json-which-is-more-efficient-for-language-models-5bc11dd0f6df](<https://medium.com/better-programming/yaml-vs-json-which-is-more-efficient-for-language-models-5bc11dd0f6df>)

### Proposed Solution

Add alternative structured output format as an option per prompt in `.baml` file or per project as CLI setting/argument?

### Alternative Solutions

*No response*

### Additional Context

*No response*

lukeramsden:
This would be a very interesting and powerful differentiator for BAML but I can't imagine it's simple - would require new versions of `jsonish` for each format. It's a good argument for separating out the various libraries that constitute BAML as a whole though

yyuyulm:
Maybe supporting strict parsing for those additional formats can be a first step, and fuzzy parsing can be added as a follow-up? Models are getting better at tool calling these days. Most of the failure cases I have encountered are models failing to generate structure output at all, rather than incorrect formatting.

aaronvg:
We created a design doc for XML format -- is XML something people are interested in? Thumbs up for this if you want XML

The tl;dr is that we can do a simpler version of xml schemas as outlined in the doc. We wouldnt need to do fancy parsing for this initially. Fuzzy parsing would come much later.

https://gloochat.notion.site/XML-Outputs-218bb2d26216806fb4dfef242d150c51

yonil7:
[...]

PR #3991, opened 2026-07-12
feat(output-format): add XML output format

mvanhorn:
## Issue Reference
- [x] This PR addresses #2409

@aaronvg mentioned on #2409 that there's a design doc for an XML output format and asked who'd be interested. This is a first pass — happy to reconcile the element/attribute mapping with whatever the design doc specifies before merge.

## Changes
Adds XML as a selectable output format alongside JSON:

- `{{ ctx.output_format(format="xml") }}` renders an XML-shaped schema in the prompt instead of JSON. `format` defaults to `"json"`, so existing prompts are unchanged.
- The `format` argument is validated at compile time in the jinja type checker.
- Response parsing decodes XML through the same schema-aligned (`jsonish`) coercion used for JSON, with preamble tolerance, malformed closing-tag recovery, null handling, Unicode aliases, and a depth limit. Existing JSONish behavior for XML-like reasoning/code text is preserved.
- Shares the output-format backend with the TOON work in #3990, so adding formats stays a single abstraction rather than bespoke code paths.

Rendering lives in `engine/baml-lib/jinja-runtime/src/output_format/` (new `xml.rs`); parsing in `engine/baml-lib/jsonish/` (new `xmlish.rs`). Docs in `fern/03-reference/baml/prompt-syntax/output-format.mdx`.

## Testing

- [x] Unit tests added/updated

Added XML fixtures covering class/enum/list/map/optional/union/recursion/aliases, preamble tolerance, and malformed closing tags. Local runs:
- `cargo test -p internal-baml-jinja -p jsonish --lib` — 120 + 492 tests pass
- `cargo clippy -p internal-baml-jinja -p jsonish --lib` clean
- current-branch `baml check` against an XML prompt fixture, `git diff --check` clean

## Screenshots

Simulated demo of the prompt-side contrast (JSON vs XML schema for the same type) and the parse-back to a typed value:

XML output format demo

## PR Checklist

- [x] I have made corresponding changes to the documentation

## Additional Notes

Motivation from the issue thread: some models emit more reliable structured output in XML than JSON (the Morph/Warp-Grep guide @yonil7 linked). Branch is aligned with current `canary`. AI was used for assistance.

## Summary by CodeRabbit

* **New Features**
  * Added `ctx.output_format(format=...)` to choose `none | "json" | "xml"`, including XML-structured prompt/output schema rendering.
  * Enabled XML deserialization into typed results (classes, enums, lists, maps, optionals, unions, and recursive types with aliases).
* **Documentation**
  * Documented XML formatting, XML type mapping rules, alias behavior, and XML element-name sanitization.
* **Bug Fixes**
  * Improved XML input robustness (tolerates preamble/whitespace and recovers from incomplete or slightly malformed XML).
* **Tests**
  * Expanded rendering and XML→structured decoding coverage, including Unicode/sanitization and additional error cases.


aaronvg:
This is the design doc: https://app.notion.com/p/gloochat/XML-Outputs-218bb2d26216806fb4dfef242d150c51?source=copy_link 

It'd be great to have a ton of tests for this kind of thing (recursive types, arrays, classes, unions). All of these are discussed in the design doc as well

mvanhorn:
Addressed the review in ba5350a — `output_format` now parses `format` as `Option<String>`, so an explicit `format=none` resolves to the default JSON format instead of failing at runtime (kept the type union so `format` stays optional). Updated the null-options test to cover it.

mvanhorn:
Thanks @aaronvg — I'll go through the design doc and make sure the element/attribute mapping and list encoding match it. The PR already covers classes, enums, lists, maps, optionals, unions, and recursion in tests; I'll expand the suite to exercise deeper recursive types and nested arrays/unions per the doc. One thing: the Notion link is gated for me — if there are specific mapping decisions there you want matched exactly (e.g. how unions or nullable fields serialize), a quick note here would help me align precisely.

mvanhorn:
Pushed 869ae00 with expanded tests — recursive types, arrays of classes/unions, nested arrays, nullable fields, class-valued maps, and malformed-but-recoverable XML. Adding the nested-array case actually surfaced a real bug (nested XML arrays were parsing as empty), now fixed in `xmlish`. I still owe a pass to reconcile the exact serialization against the design doc; if the union/nullable mapping there differs from what I have I'll match it — a pointer to the specific section would help since the Notion link is gated for me.

mvanhorn:
Went through the design doc and aligned the XML rendering to it in 94f7de6:
- nested classes render inside a `<ClassName>` wrapper
- class arrays use the `{ClassName}List` wrapper with `<ClassName>` elements; primitive/enum/union arrays use `<list><item>`
- optionals render inline as `value or null` (honoring `or_splitter` / `render_null_as`)
- class unions render as sibling `<ClassName>` blocks separated by a bare `OR`
- recursive classes reference their name inline, with enum/class hoisting and the doc's default prefixes

Parser updated to accept the new shapes; 633 tests pass. One question: the doc's canonical arg is `type=xml` (and `ctx.output_format_xml`), but I kept `format="xml"` here to match the TOON PR (#3990). Want me to switch to `type=xml`? If so, it'd probably make sense to move TOON to `type=toon` too for consistency.

mvanhorn:
Thanks for the thorough pass. Pushed `bb71619` addressing all five:

1. **Tag collisions** — distinct raw field names that sanitize to the same XML tag now reject the class with a clear error instead of emitting indistinguishable fields.
2. **Exhaustive type handling** — the union-variant catch-all resolves recursive aliases through the alias path, handles nested unions, and returns explicit errors for tuple/arrow/Top rather than serializing via `to_string()`; the scalar fallback is limited to genuine scalars.
3. **Union maps** — map variants now render under a `<map>` root so the parser's variant disambiguation recognizes them. Added render→parse round-trip tests for both `map | class` and `map | null`.
4. **Escaping** — the optional/null suffix is XML-escaped before insertion.
5. **Class descriptions** — rendered as a leading XML comment for root, nested, and expanded classes, matching the JSON renderer.

641 tests pass (`internal-baml-jinja` + `jsonish`), `cargo fmt` clean. @aaronvg let me know if any of this diverges from the design doc.

PR #3997, opened 2026-07-13
fix(graph): resolve root-qualified namespaced function calls

jubjub727:
## Author Comment

I found a bug with namespaces where referencing them with root results in graph view breaking. Fixed it by adding a single helper function that matches the name shape with what the downstream code consuming it expects by stripping the leading "root." from the name.

## Codex Summary

Cross-namespace BAML calls use root-qualified paths such as `root.task_routing.ChooseTask(...)`. These calls compile and run correctly, but the graph analyzer retained the leading `root.` while its function resolver expected the playground-qualified name `task_routing.ChooseTask`.

The lookup therefore failed to recognize the callee as an LLM function. Graph preparation intentionally preserves LLM calls while pruning ordinary helper calls without `//#` headers. Consequently, the misclassified call was removed and affected workflows could appear as only a root node.

This change removes one leading `root.` before graph lookup. The normalization applies both to requested graph entry points and to callees extracted from the AST. It is limited to static graph analysis and does not change runtime name resolution or displayed source labels.

The regression test verifies that:

- A root-qualified namespaced workflow can be used as a graph entry point.
- A cross-namespace `root.*` call resolves as `NodeType::LlmFunction`.
- The LLM call survives visualization pruning without a `//#` header.

Tested with:

```text
cargo test -p baml_project test_root_qualified_llm_call_is_resolved_and_rendered --lib
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workflow calls using root-qualified names so they resolve correctly.
  * Ensured namespaced language-model functions remain visible in workflow visualizations.

### DIFF
--- a/baml_language/crates/baml_project/src/db.rs
+++ b/baml_language/crates/baml_project/src/db.rs
@@ -508,6 +508,10 @@ impl ProjectDatabase {
         self.ast_control_flow_graph_impl(function_name, &mut ctx)
     }
 
+    fn normalize_graph_function_name(function_name: &str) -> &str {
+        function_name.strip_prefix("root.").unwrap_or(function_name)
+    }
+
     fn ast_control_flow_graph_impl(
         &self,
         function_name: &str,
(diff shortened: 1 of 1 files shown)


aaronvg:
Thanks for the PR!

sxlijin:
per discord discussion, closing this

PR #4016, opened 2026-07-14
perf(compiler2): ~32x faster cold compile — remove redundant work, memoize package-invariant queries

hellovai:
## TL;DR

Cold `check` + `emit` on the in-repo test corpus (78 files, 25,838 lines) goes from **16.0s → ~0.50s** (~32×). End-to-end `baml check` wall time on the same corpus goes from **~17s → ~0.55s**. Everything here is **single-threaded** — no parallelism, no on-disk cache. Every win is one of three things:

1. deleting work we were doing more than once,
2. turning package/project-invariant computation into a Salsa-tracked query so it runs once instead of once-per-scope/per-function, or
3. cutting allocation churn on the hot type-checking paths.

It also **fixes two user-visible bugs** found along the way (details below), and ships a new profiling harness (`crates/tools_compile_profile`) whose README documents the entire audit so the next person doesn't have to rediscover any of this.

## Why

`baml check` on a real project was taking ~17 minutes. The pipeline is Salsa-based, but a cold CLI run gets zero benefit from Salsa's memoization *across* invocations — so cold-path cost is pure single-run efficiency. Profiling (query-event counts + `samply` CPU sampling) showed the time wasn't in any one algorithm; it was the same few computations being repeated tens of thousands of times.

## How to review this PR

Each change is independent and documented at its call site with a comment explaining what was being recomputed and roughly what it cost. The full narrative, methodology, and per-change measurements live in **`crates/tools_compile_profile/README.md`** (section "The July 2026 cold-compile audit") — reading that first will make every diff hunk make sense.

### Stop recomputing package/project-invariant data (the big one)

| What | Before | After |
|---|---|---|
| Recursive-alias analysis (`find_recursive_aliases`, a whole-alias-graph DFS) | ran inside **every** subtype/type-equality check — 18k executions | runs once per package; result travels in the new `baml_type::ResolvedAliases` (alias map + precomputed recursive set), threaded by reference through TIR/MIR |
| Package alias map & impl-block list | rebuilt (deep `Ty` clones + project walk) inside every `infer_scope_types` execution (~16k) | `package_resolved_aliases` / `package_impl_locs` tracked queries |
| CST → AST lowering | re-lowered per file by **six** different consumers (~31% of cold CPU) | `file_ast(db, file)` tracked query |
| Project-wide class → type-tag map | rebuilt per *function* lowered in MIR (~13% of cold CPU) | `class_type_tags_for_project` tracked query |
| Callee generic params + bound lowering | re-derived at every call site (~10% of cold CPU) | `callee_generics_for_func` tracked query |
| Callee function body for direct calls | PPIR cloned the entire `ExprBody` arena per call site | `ppir::function_body` tracked, returns `Arc` |

### Stop doing the same inference twice

- **Lambda bodies were type-inferred twice**: once inline while inferring the enclosing function, then again from scratch by the standalone lambda-scope query. The inline pass now captures the lambda's complete inference tables (`NestedLambdaInference`, moved not cloned) and the standalone query projects them. This is also a **bug fix**: diagnostics inside lambdas were previously **reported twice**.
- Files with no `$stream` expansions had their semantic index built twice (identical results); PPIR's `file_semantic_index` now delegates to HIR's.
- The builtin stdlib was eagerly type-checked on every `baml check` (a fixed cost floor even for a 1-line project) even though it can't produce user-actionable diagnostics. Diagnostics collection now skips builtin files; builtin scopes user code touches are still inferred lazily.

### Make the remaining hot path cheap

- **Subtype/equality fast paths** (`baml_type::normalize`): reflexivity short-circuit; co-inductive assumption bookkeeping (deep clone + full-tree hash per step) now only happens in the arms that can actually regress a cycle (μ-types, type vars, projections) — the millions of purely structural steps skip it. The safety argument is written out in a comment at the site.
- **`heads_definitely_differ` filter**: MIR impl dispatch probes are overwhelmingly misses between types with different nominal heads — now decided without the two allocation-heavy normalization walks.
- **Interface-method-name pre-filter in MIR dispatch**: `dispatch_target_for_concrete` enumerated every impl block in the package closure for every method call *and field access*. A per-package `FxHashSet` of interface-declared method names rejects the common plain-method case in one lookup (`is_same_normalized_type` calls: 952k → 275k).
- **mimalloc** as the global allocator for `baml_cli` (and the profiler). System malloc was ~35% of remaining CPU — the workload is small-allocation heavy (`Ty` trees have deep value semantics). This is mitigation; the foundational fix (interning `Ty`) is scoped as future work in the README.

### CLI-side bug fix

- **Diagnostic rendering rebuilt ariadne's line index per diagnostic.** `render_diagnostics` constructed a fresh `SourceCache` — a line index over *every project file* — for each diagnostic. With 79 warnings that was 34% of total `baml check` wall time. Built once per batch now. (CLI 0.94s → 0.55s.)

### New: `crates/tools_compile_profile`

A standalone profiling harness that runs the exact `baml check` pipeline against any project directory and reports wall-clock per stage, Salsa query execution/cache-hit counts by pipeline phase, top queries, and a "suspect" table (high-exec, low-hit queries — i.e. caching gaps). Supports `--warm-runs`, `--repeat`, `--json`, `--check-only`, and pairs with `samply` for line-level CPU profiles. The README documents how to use it, how to read its output, and the full audit above.

## Correctness

- Full test suite passes. Three snapshots changed, all expected: two reflect the duplicate-lambda-diagnostics fix (diagnostics no longer emitted twice), one reflects lambda-scope inference now being projected (same types, same diagnostics, different provenance).
- The `package_resolved_aliases` query participates in a legitimate dependency cycle (aliases whose RHS is an associated-type projection resolve through inference, which consults the alias map); it now declares `cycle_initial` and seeds fixpoint iteration with the empty environment, mirroring `infer_scope_types`.
- No language-semantics changes: every optimization computes the same values, just fewer times.

## Numbers (test corpus, cold, median of 5)

| Stage | Before | After |
|---|---|---|
| check (diagnostics) | 8.1s | 0.32s |
| emit (bytecode) | 13.2s (worst measurement) | 0.17s |
| **pipeline total** | **16.0s** | **~0.50s** |
| `baml check` CLI wall | ~17s | ~0.55s |

## Explicitly out of scope (documented as future work in the README)

- **Parallelism** — untouched, remains available as a multiplier on top of this.
- **`Ty` interning / hash-consing** — ~25% of remaining self-time is still clone/drop/malloc of `Ty` trees with deep value semantics. Staged plan (intern `QualifiedTypeName` → intern `Ty` → arena `NormalTy` temporaries) is written up in the README.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Performance**
  * Improved cold compilation performance for `baml check` and `baml build`, particularly on macOS.
  * Reduced repeated work during type checking, interface resolution, and lambda analysis.

* **Bug Fixes**
  * Improved handling of recursive type aliases and subtype checks.
  * Prevented compiler diagnostics from including internal built-in files.
  * Reduced duplicate diagnostics for nested lambda expressions.
  * Improved diagnostic rendering efficiency across multiple files.

* **Developer Tools**
  * Added a standalone compiler profiling tool with timing, cache, query, JSON, and flamegraph-supporting reports.
  * Added comprehensive documentation and usage guidance for profiling compiler performance.

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -745,6 +745,7 @@ dependencies = [
  "indicatif",
  "insta",
  "libsui",
+ "mimalloc",
  "object 0.36.3",
  "regex",
  "reqwest",
@@ -833,6 +834,7 @@ dependencies = [
  "baml_compiler2_ppir",
  "baml_compiler2_tir",
  "baml_type",
+ "baml_workspace",
  "indexmap",
  "num-bigint",
  "rustc-hash 2.1.2",
@@ -853,6 +855,22 @@ dependencies = [
  "salsa",
  "smol_str 0.3.6",
  "text-size",
+]
+
+[[package]]
+name = "baml_compiler2_profile"
+version = "0.0.0-beta"
+dependencies = [
+ "anyhow",
+ "baml_compiler2_tir",
+ "baml_compiler_diagnostics",
+ "baml_project",
+ "baml_workspace",
+ "clap",
+ "mimalloc",
+ "salsa",
+ "serde",
+ "serde_json",
 ]
 
 [[package]]
@@ -988,6 +1006,7 @@ dependencies = [
  "baml_compiler_diagnostics",
  "baml_compiler_parser",
  "baml_compiler_syntax",
+ "baml_type",
  "baml_workspace",
  "bitflags 2.11.1",
  "indexmap",
@@ -4237,6 +4256,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.49"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a45a52f43e1c16f667ccfe4dd8c85b7f7c204fd5e3bf46c5b0db9a5c3c0b8e9"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4486,6 +4514,15 @@ dependencies = [
  "log",
  "objc",
  "paste",
+]
+
+[[package]]
+name = "mimalloc"
+version = "0.1.52"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d4139bb28d14ad1facf21d5eb8825051b326e172d216b39f6d31df53cc97862"
+dependencies = [
+ "libmimalloc-sys",
 ]
 
 [[package]]
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -858,22 +858,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "baml_compiler2_profile"
-version = "0.0.0-beta"
-dependencies = [
- "anyhow",
- "baml_compiler2_tir",
- "baml_compiler_diagnostics",
- "baml_project",
- "baml_workspace",
- "clap",
- "mimalloc",
- "salsa",
- "serde",
- "serde_json",
-]
-
-[[package]]
 name = "baml_compiler2_tir"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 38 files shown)


hellovai:
Superseded — closing in favor of the re-landed work, but **keeping this branch as the reference implementation** (do not delete).

Canary moved underneath this PR, so rather than rebase it we re-derived each still-orthogonal optimization from scratch against current `canary` and re-measured:

- **Profiler** (`tools_compile_profile`) landed separately as #4038.
- **Biggest finding not re-landed** — the recursive-alias hoist (16s → 4s) was independently captured by #4032, which deleted the old TIR `StructuralTy` algebra. The subtype/dispatch fast paths were re-derived onto the post-#4032 `baml_type::normalize` equivalence path instead.
- **Everything else re-landed** in #4058 as one PR, one commit per track: `file_ast` tracked query (A), package-level TIR queries + `callee_generics_for_func` (B), nested-lambda inference projection (C), MIR dispatch prefilter + `is_subtype_of` fast path (D), CLI mimalloc + ariadne SourceCache-once (E). Skip-builtin-diagnostics was dropped as superseded by #3924.

Result on the test corpus (cold, disk cache disabled, medians of 5): current canary `79aa0f44e` **2.264s** → #4058 **1.003s** (2.3x; check 2.8x, emit 1.9x). Full audit table and per-track detail are in #4058.

This PR's branch `perf/compiler2-cold-compile` stays as the historical reference.

PR #4031, opened 2026-07-14
feat(compiler2): allow binding to `..` in array patterns

codeshaunted:
Closes B-531.

Array rest patterns can now carry binding-shaped sub-patterns: `..let r`, `.._`, bind chains (`..let r: let s`), and a slice-typed ascription link (`..let r: int[]`). The binding is typed `elem[]` and binds a copy of the matched middle via `baml.Array.slice`.

## What is allowed
```baml
match (xs) {
    [let first, ..let mid, let last] => mid.length(),
    [..let all] => all.length()
}
```
Works in match arms, `let` (rest-only), `if let`, `for`, or-branches, and nested element patterns. The rest binds a shallow copy: mutating it does not touch the scrutinee and vice versa; elements are shared.

## What stays rejected (targeted diagnostic, expandable later)
- Type patterns / class destructures after `..` (`..int`, `..Box {}`): statically dead — the slice's runtime tag is always exactly `elem[]` and list type tests are invariant tag compares.
- Refutable shapes (`..([] | [_])`): the usefulness matrix (rustc's slice model) has no column for a rest constraint; TIR drops the rest sub-pattern's DPat while MIR emits its runtime test, so allowing these makes 'exhaustive' matches that fall through at runtime.
- Nested array rests (`..[a, ..r, b]`): sound via flattening, but a second spelling of the flat pattern (`[..[]]` = `[]`); the flattener is removed.

Rejected rests recover by binding their names as unknown, so arm bodies get one diagnostic instead of an unresolved-name cascade.

## Semantics fixes along the way
- Rest ascriptions are checked for **equivalence** with the slice type, not subtyping: `..let r: int[]` on `(int|string)[]` can never match at runtime, so it is now a type mismatch instead of a silently dead arm.
- MIR only materializes the slice when the rest sub-pattern needs it: plain `..let r` previously copied the middle twice (test phase + bind phase) and `.._` once; now once and zero. The pre-existing or-pattern MIR snapshot drops from 4 slice calls to 1.

## Tests
- `compiler2_tir/array_rest.rs`: 25 unit tests — typing, ascription equivalence, or-pattern name sets, refutability, exhaustiveness, per-shape rejections, nested-rest torture (`[..[..]]`, `[..[]]`, single-diagnostic depth-5 nesting), recovery.
- `tests/array_rest_binding.rs`: 16 runtime e2e — values, empty rests, fall-through, copy/shallow-copy semantics, or-arm dispatch, generics, if-let/for.
- MIR snapshots pin the slice projection (with/without suffix) plus a no-slice assert for `.._`.
- Corpus: `diagnostic_errors/patterns_array_rest_binding` (semantic; pins formatter output for the new syntax) and `broken_syntax/patterns_array_rest_parser_recovery` (parser garbage: double rests, `[...let r]`, `[..let]`, bare/wildcard ascriptions).

## Summary by CodeRabbit

* **New Features**
  * Expanded array pattern rest bindings, including wildcard, typed, chained, nested, and generic forms.
  * Added runtime support for capturing middle, tail, or full array slices.
  * Improved handling of rest bindings in matches, conditional destructuring, loops, and alternative patterns.

* **Bug Fixes**
  * Wildcard rests no longer create unnecessary slice operations.
  * Invalid rest patterns now produce clearer diagnostics with improved parser recovery.

* **Tests**
  * Added comprehensive compiler, diagnostic, parser-recovery, and runtime coverage for array rest patterns.


codeshaunted:
@coderabbitai full review

PR #4038, opened 2026-07-15
chore(tools): add tools_compile_profile — salsa/pipeline profiling harness

hellovai:
## Summary

- Adds `baml_language/crates/tools_compile_profile`, a standalone black-box profiling harness for the compiler pipeline (parse → HIR → PPIR → TIR → MIR → emit). It reports wall-clock per phase, per-Salsa-query execution / cache-hit / blocked counts, cold-run variance, cold-vs-warm comparison (`--warm-runs`), and JSON output.
- Ported verbatim from PR #4016 (`perf/compiler2-cold-compile`), with two adaptations: the `PROF_*` TIR audit counters (which only existed on that branch) are removed rather than ported, and the "July 2026 cold-compile audit" README section is annotated as a historical record predating #4032 / #3924.
- Workspace changes: `mimalloc = "0.1.52"` added to workspace deps (used only by this tool, as its global allocator — noted in the README), and `"compile"` added to the `tools` namespace `approved_prefixes` in `stow.toml`.

Provenance: #4016. The remaining findings from that audit are being re-landed as small independent PRs measured with this tool.

## Baseline on current canary (2660b8b9f)

Protocol: `cargo build --release -p tools_compile_profile`, then
`./target/release/tools_compile_profile baml_language/crates/baml_tests/baml_src --repeat 5`
with `BAML_NO_BYTECODE_CACHE=1` and `BAML_CACHE_DIR` pointed at a fresh temp dir (so #3924's disk cache is out of the picture). Corpus: 77 files, 25,212 lines. Apple Silicon, single-threaded, medians of 5 cold runs.

| phase | median |
|---|---|
| check | 1.089 s |
| emit (bytecode) | 1.330 s |
| **total** | **2.420 s** (min 2.385, mean 2.442, max 2.524) |

For comparison, #4016's pre-optimization baseline on this corpus was 16.0s and its final state ~0.50s. Canary's own movement (#4032 deleting the old TIR type algebra, #3924's caching work) already brought cold compile from 16.0s to ~2.4s; the still-orthogonal optimizations from #4016 target the remainder.

### Top 10 queries by executions (cold, representative run)

| exec | query |
|---|---|
| 15,590 | `infer_scope_types` |
| 2,262 | `function_in_scope_generic_param_bounds` |
| 2,256 | `function_body` |
| 2,131 | `callable_throws` |
| 625 | `class_generic_param_bounds` |
| 477 | `resolve_class_fields` |
| 419 | `function_parameter_defaults` |
[...]

PR #4043, opened 2026-07-15
perf(compiler2): tracked file_ast query — lower CST→AST once per file

hellovai:
## What

On `canary` the CST→AST lowering step (`baml_compiler2_ast::lower_file`) is a plain, **untracked** function that six different consumers each re-run from scratch for the same file:

- `baml_compiler2_hir::file_semantic_index`
- `baml_compiler2_ppir::file_semantic_index`
- `baml_compiler2_ppir::ppir_expansion_items`
- the two project-wide expansion-map collectors (`collect_block_attrs` / `collect_alias_bodies`)
- the `baml_lsp2_actions` check pass

Repeated CST traversal was ~31% of cold-compile CPU on the test corpus in the original audit. This PR memoizes lowering and removes two other redundant rebuilds.

## Changes

1. **`baml_compiler2_hir::file_ast(db, file)`** — a new `#[salsa::tracked]` query that lowers the CST once per file and shares `items` + lowering `diagnostics` + `env_var_refs`. All six consumers now read it instead of re-lowering. (`EnvVarRef` gains `PartialEq, Eq` so `FileAst` can use `PartialEq` for Salsa early-cutoff.)
2. **PPIR `file_semantic_index` delegates to HIR** when a file has no expansion items. The post-expansion index is byte-for-byte identical to the pre-expansion one for such files (same AST items, same builder, same file range), so rebuilding it was pure wasted work. The merged path is preserved as `file_semantic_index_expanded` for files that actually have `*$stream` companions. The delegation also unifies scope identity for expansion-free files, which cuts `infer_scope_types` executions (15,590 → 13,331).
3. **PPIR `function_body` is now `#[salsa::tracked]`** returning `Arc<FunctionBody>` (mirroring HIR's `function_body`). MIR lowering fetches the callee body at every direct-call site; the untracked version cloned the entire `ExprBody` arena each time.

Each changed site carries a short comment explaining what was being recomputed.

## Before / after (cold compile)

Measured with the `tools_compile_profile` harness from #4038 (not committed here), disk cache disabled via `BAML_NO_BYTECODE_CACHE=1`, on `crates/baml_tests/baml_src` (77 files, 25 212 lines). Numbers are medians of 5 fresh-database cold runs, before/after binaries run interleaved on a quiet machine (two full rounds, both consistent; quietest round shown):

| phase | before (canary 2660b8b9f) | after | delta |
|-------|--------------------------|-------|-------|
| check | 1.118 s | 0.752 s | −33 % |
| emit  | 1.361 s | 1.181 s | −13 % |
| total | 2.480 s | 1.934 s | −22 % |

Load-independent evidence, same direction:

- instructions retired (3 cold runs, `/usr/bin/time -l`): 93.7 G → 76.4 G (−18.5 %), stable across 3 interleaved rounds
- `file_ast` executes 131× (exactly once per file) instead of lowering inline ~6× per file
- `file_semantic_index_expanded` runs only 78× (files that really have `*$stream` expansions); the other 53 files reuse HIR's index
- `infer_scope_types`: 15,590 → 13,331 executions

## Tests

- `cargo test --workspace`: **all suites pass** (~190 binaries). Initial failures were environmental only and reproduced on clean canary too: disk-full during the run, the Python SDK venv resolving to Python 3.10 (fixed with `UV_PYTHON=3.12`), and the Node fixtures needing `sdk_tests/crates/typescript_node/setup.sh` run once; after that `sdk_test_python_pydantic2` (14/14) and `sdk_test_typescript_node` (15/15) pass.
- **Zero snapshot diffs** (no `.snap.new` files) — outputs are byte-identical, as required for this track.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean.
- `cargo fmt` check: clean.

## Provenance

Re-derived against current `canary`; reference implementation: #4016 (`perf/compiler2-cold-compile`), audit items 3, 9, 13.

### DIFF
--- a/baml_language/crates/baml_compiler2_ast/src/lower_expr_body.rs
+++ b/baml_language/crates/baml_compiler2_ast/src/lower_expr_body.rs
@@ -22,7 +22,7 @@ use crate::{
 };
 
 /// A reference to an environment variable found in source code (`env.VAR_NAME`).
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EnvVarRef {
     /// The variable name (e.g., `"OPENAI_API_KEY"`).
     pub name: String,
--- a/baml_language/crates/baml_compiler2_hir/src/lib.rs
+++ b/baml_language/crates/baml_compiler2_hir/src/lib.rs
@@ -98,6 +98,57 @@ pub fn compiler2_all_files(db: &dyn Db) -> Vec<baml_base::SourceFile> {
     files
 }
 
+// ── file_ast ──────────────────────────────────────────────────────────────────
+
+/// The CST → AST lowering output for one file: top-level items plus the
+/// lowering diagnostics and `env.*` references produced along the way.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FileAst {
+    pub items: Vec<baml_compiler2_ast::Item>,
+    pub diagnostics: Vec<baml_compiler2_ast::LoweringDiagnostic>,
+    pub env_var_refs: Vec<baml_compiler2_ast::EnvVarRef>,
+}
+
+// Safety: `FileAst` holds only plain (non-`'db`) data, so storing it by value
+// is sound. Manual `Update` impl uses `PartialEq` for Salsa early-cutoff.
+#[allow(unsafe_code)]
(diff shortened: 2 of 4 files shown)


hellovai:

PR #4044, opened 2026-07-16
perf(compiler2): interface-method-name pre-filter in MIR dispatch + is_subtype_of assumption fast path

hellovai:
Re-lands two slices of the cold-compile performance audit (items 12 and 7 of PR #4016), re-derived against current canary. Both are pure fast paths: dispatch results and subtype verdicts are unchanged.

## What was being recomputed

**1. MIR dispatch pre-filter (`baml_compiler2_mir/src/lower.rs`)**

`dispatch_target_for_concrete` enumerated every impl block in the package closure (`l1_impls_for_recv`, which probes each impl's pattern against the receiver, invoking alias normalization / subtype checks per probe) for **every** method call and field access — almost always just to conclude "no impl provides this member". Since #4032 deleted the TIR structural algebra, each of those probes goes through `AliasEquivCtx` / `baml_type::normalize`, making the per-call enumeration even more expensive.

Fix: collect every interface-declared method name (required + default, own package + dependency closure) once per package into the existing Salsa-tracked `PackageLoweringData`, and have `dispatch_target_for_concrete` return `None` in one hash lookup when the member name is absent. The set is package-wide, not per receiver type, so any name declared by **any** reachable interface still takes the full enumeration — the filter cannot change dispatch results.

**2. `is_subtype_of` assumption fast path (`baml_type/src/normalize.rs`)**

The co-inductive assumption bookkeeping (deep `NormalTy` clone + full-tree hash of the `(lhs, rhs)` pair, insert + remove) ran on **every** recursive step of the subtype check. Canary already had the reflexivity short-circuit, so only the bookkeeping split was needed.

Fix: split the structural rules into `is_subtype_of_inner`; the outer function performs the pair bookkeeping only for the *expanding* arms — `Mu`, `TypeVar`, `AssociatedTypeProjection` on the left, or `Mu` on the right.

**Termination argument** (also documented as a comment at the split): assumption tracking exists solely to terminate cycles, and a cycle can only regress through arms that expand a type — μ-unfolding (substitution can regenerate the same pair) and variable/projection bound lookup (a bound can mention the variable). Purely structural arms descend into strictly smaller subterms of finite trees, so no cycle can form through them; any infinite path must pass through an expanding arm infinitely often, and those still record assumptions (drawn from the finite subterm closure of the regular operands).

## Measurements

Profiler from PR #4038 (`tools_compile_profile`, not part of this PR), `crates/baml_tests/baml_src` (77 files, 25k lines), `BAML_NO_BYTECODE_CACHE=1`, fresh Salsa db per run, `--repeat 5`, Apple Silicon:

| | check (median) | emit (median) | total (median) |
|---|---|---|---|
| canary (2660b8b9f) | 1.454s | 1.791s | 3.250s |
| this PR | 1.358s | 1.295s | 2.675s |

Emit (where MIR lowering and dispatch run) drops ~28%; total cold compile ~18%.

## Testing

- `cargo test --workspace`: all compiler/runtime suites pass, including the 442-test `baml_tests` suite — no snapshot diffs, so dispatch results and subtype verdicts are byte-identical.
- Only failures were environmental, unrelated to this change: `sdk_test_typescript_node` (missing `node_modules` in the test sandbox — `tsc`/`vitest`/`attw` not installed) and 4 `sdk_test_python_pydantic2` cancellation tests (fixture uses `ExceptionGroup`, undefined on the sandbox's Python 3.10, plus asyncio-timing asserts).
- Pre-commit hooks (cargo fmt, workspace clippy `-D warnings`) pass.

Provenance: re-derivation of c1466f3cc (PR #4016) items 12 and 7.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type compatibility checks involving recursive types, type variables, and associated types, helping prevent incorrect results or excessive processing.
  * Optimized method dispatch checks by quickly rejecting methods that are not declared by applicable interfaces, improving compilation performance.

### DIFF
--- a/baml_language/crates/baml_compiler2_mir/src/lower.rs
+++ b/baml_language/crates/baml_compiler2_mir/src/lower.rs
@@ -1523,7 +1523,7 @@ use baml_compiler2_tir::{
     inference::infer_scope_types,
     resolve::{ResolvedName, resolve_name_at_in_scope},
 };
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHashSet};
 
 type ClassFieldIndices = IndexMap<TypeName, IndexMap<String, usize>>;
 type ClassFieldTypes = IndexMap<TypeName, IndexMap<String, RuntimeTy>>;
@@ -1646,6 +1646,12 @@ struct PackageLoweringData {
     enum_variants: EnumVariantIndices,
     interface_implementors: ImplementorsByInterface,
     resolved_aliases: ResolvedAliases,
+    /// Every method name any in-scope interface (own package + dependency
+    /// closure) declares, required or default. Fast pre-filter for
+    /// [`LoweringContext::dispatch_target_for_concrete`]: a member name absent
+    /// here can never dispatch through an interface impl, so the (hot) impl
+    /// enumeration is skipped for the overwhelmingly common plain-method case.
+    interface_method_names: FxHashSet<Name>,
 }
 
 /// # Safety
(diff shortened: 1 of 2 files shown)


hellovai:

PR #4075, opened 2026-07-17
perf(ci): faster cargo tests via dedup, opt-level, cache sharing + migrate exec tests to BAML corpus

hellovai:
## Goal

Get CI (\`CI - BAML Language\`) from ~12 min toward <5 min. The critical path was the two linux cargo jobs (~11 min each, 8-vcpu saturated). Almost all of that time is test *execution*, not compilation: the tests run the compiler thousands of times.

### `perf(ci)` — cut cargo-test wall time
- **Dedup**: the linux `cargo test` job re-ran `baml_tests`/`baml_cli`/`baml_lsp2_actions` that the snapshot job already covers on the same OS. Now excluded at the cargo level (`--workspace --exclude`), not just via a nextest `-E` filter, so their heavy test binaries aren't even built. The duplicated `pack_e2e` step is dropped (the snapshot job runs it).
- **Opt-level**: tests exercise the compiler thousands of times, so build it optimized under test — `CARGO_PROFILE_DEV/TEST_OPT_LEVEL=1` in CI, and `salsa` (the query engine that runs on every compile) at opt-level 3 in the dev profile.
- **Cache sharing**: the spawned-CLI e2e suites (`exit_code_e2e`, `pack_e2e`) hit a cold `<project>/.baml/cache` per test and recompiled the stdlib every time (~14s/spawn). They now share a `BAML_CACHE_DIR` under `CARGO_TARGET_TMPDIR`, so only the first invocation pays the stdlib compile. The cache is content-addressed and fingerprinted on the CLI binary, so sharing across tests/runs is safe.
- **Phantom build**: `exit_code_e2e` called `ensure_built()` (which shells `cargo build -p baml_pack_host`, ~10s/process even when fresh) despite never running `baml pack`; switched to `baml_cli()`.
- Both linux jobs bumped 8 → 16 vcpu (cost-neutral: they were CPU-saturated).

### `test(baml_tests)` — move exec tests into the BAML corpus
Each Rust-side `baml_test!` test recompiled a BAML project from scratch in its own nextest process (~0.7–2.5s); a `test { }` block in the `baml_src/` corpus compiles once with the whole project and runs for ~40ms.
- **370 tests migrated** across 43 files, one block per former Rust test (name preserved), each validated with `baml-cli test`.
- **Stays in Rust, documented at each site**: compile-diagnostics tests (`setup_test_db` + `collect_diagnostics`), host-FFI tests (`BexExternalValue`, VM introspection), host infrastructure (stub HTTP/TCP servers, symlink creation), `OptLevel::Two` dispatch, and `#[should_panic]` compile-time checks.
- The corpus runner snapshots each namespace's bytecode; new/extended namespaces add the corresponding `baml_src/*.snap`.

## Verification
- Full `cargo nextest run -p baml_tests`: **2819/2819 pass**.
- Local measurements (salsa-opt only, before CI opt-1): warm spawned-CLI test 14s → 1.4s; `exit_code_e2e` suite 574s CPU → 147s.

## Notes for review
- **First CI run is sccache-cold** for the opt-level change (one slow build per platform); it warms once canary refreshes the cache.
- **Follow-up, not in this PR**: feature-gate `baml_test!` behind an off-by-default `rust-baml-exec-tests` so new BAML *behavior* tests are written in the corpus, not Rust (compile-diagnostics tests keep running under a plain `cargo test`).
- The pre-existing slow/failing `python-pydantic2 (windows)` and `Test Go Client Windows Support` jobs are unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added expanded runtime coverage for PromptAst accessors and prompt/tagged-template interpolation behavior.
* **Bug Fixes**
  * Improved correctness verification for defer/cleanup, cancellation, error-cause chains, exception routing, string handling, sorting, GC stability, CSV decoding, networking, and filesystem error mapping.
* **Tests**
  * Added many new BAML regression suites (array rest bindings, interfaces/associated types, tagged templates, JSON conversion, SSE streaming, shell/CLI scenarios) plus additional compile/diagnostic assertions.
  * Improved CI reliability by isolating CLI/cache state, sharing CLI e2e cache, and redistributing workloads while excluding heavy packages from general jobs.
* **Chores**
  * Tuned local dev build optimization for faster interactive iteration.


hellovai:
@coderabbitai I've addressed this review — summary by comment. (The review predates commit 936a462b4, which is why it's still marked changes-requested.)

**Strengthened (commit 936a462b4)** — each was a faithful port of an already-weak Rust original; all now exercise the behavior their name claims, and all still pass (full corpus 2467/0):
- `rest_bind_chain_aliases_same_slice`: pushes through one chained binding and observes it via the other (→44), proving `r`/`s` alias one slice.
- `phase2_user_class_compare_with_explicit_binding_in_main`: added a helper that binds `high.compare(low)` to a local before returning.
- parity float inputs: added a representable `-0.0` for the ±0 total-order boundary.
- `form1_self_accesses_receiver_fields`: `describe` now returns `self.label` (a real field read).
- `generic_bound_runtime`: `gbr_first_name<T extends GbrNamed>(items: T[])` over a concrete `GbrDog[]`.

**Not changed — the "catch result in a local `let` may stay boxed" comments** (cancel_cascade L21-24, fs_from_rust, glob, csv_smoke, errorcontext, interfaces_associated_types, streaming_*): these are false positives here. That VM boxing quirk is specific to **test-block** locals; every flagged binding is inside a helper `function`, which is the documented workaround for exactly that quirk (see the ns_bigints header). These pass deterministically across repeated full-suite runs, which a real boxing failure would not allow.

**Not changed — `cancel_unblocks_await_on_non_descendant` "synchronize before cancelling"**: faithful port of the Rust original, which also cancels with no readiness signal. It verifies the outcome that matters — cancelling `waiter` makes `await waiter` observe `Cancelled` in <5s instead of blocking ~60s on `slow` — which holds whether cancel lands before or after the await parks. Forcing the parked-await path specifically would need a BAML readiness primitive.

@coderabbitai review

PR #4098, opened 2026-07-21
perf: linear cold compiles, and one warm cache for check/run/pack

hellovai:
## What this does

Six fixes that came out of benchmarking the BAML compiler head-to-head against tsgo (native TypeScript) and `go build` on generated, semantically-identical corpora from 1.5k to 295k LOC.

**Compiler core**
1. **Jinja type env memoized** (`jinja_base_types` salsa query per package+namespace) + skipped for files with no prompts/`template_strings`. It was rebuilt per checked file — O(N²) — and twice per compile via `get_bytecode`'s error gate. This was ~95% of the quadratic CPU.
2. **`ProjectDatabase` handle clones O(files) → O(1)** (Arc-wrapped file maps). The parallel drivers clone a handle per work chunk (~1.55N per compile); the deep map copies were the second quadratic (CPU + peak RSS).
3. **Cache-path check collector parallelized** — `check`/`run` routed through a serial loop while the identical parallel machinery sat next to it (only reachable with the cache disabled).
4. **`get_bytecode` no longer re-checks the whole project inside emit** — the LSP proves errors once and calls `get_bytecode_unchecked`. To keep that sound, the LSP diagnostics candidate now includes package-level diagnostics (cross-file name conflicts / namespace shadows), which previously **never reached the editor at all**.

**CLI cache surface**
5. **`baml pack` now uses the bytecode cache** (it never opened one — every pack was fully cold). Whole-program hit / per-file reuse / store, sharing run+check's key space. Packed `Program` is target-independent; emit determinism keeps reused images byte-identical.
6. **`baml check` now seeds the cache** after a clean check (only when it advances the manifest — no-op checks stay emit-free; a failed emit is logged, never a check failure). Check-only workflows finally reach the warm path.

Plus `sha2/asm` (hardware SHA for the ~230MB of unit hashing per warm run at large sizes; target-gated, wasm unaffected).

## End-user impact (measured, Apple Silicon, hyperfine medians)

**Cold compiles scale linearly now** (was ~quadratic: 5× the code cost 33× the CPU):

| project size | `check` before | `check` after | `run` before | `run` after | `pack` before | `pack` after |
|--:|--:|--:|--:|--:|--:|--:|
| 59k LOC | 1.56 s | **1.01 s** | 2.30 s | **1.04 s** | 1.75 s | **1.06 s** |
| 295k LOC | 21–46 s | **7.1 s** | 27–35 s | **7.3 s** | 40.5 s | **7.2 s** |

At 295k LOC: user CPU **125.7 s → 8.9 s (14×)**, peak RSS **5.3 GB → 3.6 GB**.

**One warm cache now serves every command** (59k LOC project):

```
baml check   1.0 s    (cold, seeds the cache)
baml check   0.5 s    (warm)
baml run     0.10 s   (whole-program hit)
baml pack    0.13 s   (hit — was a full cold compile, every time)
```

Re-pack of a 295k LOC project: **40.5 s → 0.61 s** (packed binary verified to execute).

For context at 59k LOC: cold check+emit+cache-store now lands at classic `tsc --noEmit` speed (1.01 vs 1.00 s) and ahead of `go build` (1.23 s); tsgo remains far ahead (0.13 s) — closing that further is warm-floor work (tracked below), not this PR.

**Trade-off made deliberately:** cold `check` at mid-size projects (~15–30k LOC) is ~10% slower than before because it now emits + stores once to warm the cache; every subsequent check/run/pack is then served warm.

**Editor fix riding along:** cross-file duplicate-name / namespace-shadow errors now appear in editor diagnostics (they were silently dropped by the LSP's per-file sweep).

## Not in this PR (known, profiled, next)

- The warm-path O(project) floor: the throws-gate safety compare re-parses every file, and `check` loads all unit bytes it doesn't need (~90% of warm no-op cost at 295k LOC). Design for fixing it (TIR-space throws compare + manifest-resident unit metadata) is under adversarial review.
- LSP single-sweep / changed-file-first publish.
- `baml test`'s separate `emit_test_cases` keyspace still doesn't share this cache.

Letting CI run the test suites for verification; diagnostics output was verified byte-identical (cold + warm) against the pre-fix binary on an error-heavy corpus during development.

## Summary by CodeRabbit

- **Performance**
  - Faster repeated project checks and packaging using bytecode caching with incremental, fresh diagnostics reused for later runs.
  - Improved hashing throughput on supported platforms with runtime-detected hardware-accelerated SHA-2, while keeping MSVC builds on the portable path.
  - Faster Jinja template validation via memoized, package-aware type information and early exits for irrelevant files.

- **Bug Fixes**
  - More complete cross-file package diagnostics (e.g., namespace conflicts) surfaced during compilation.
  - More consistent error-only diagnostic reporting.

- **Developer Experience**
  - More robust handling of empty project directories during project-mode compilation.

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -6883,6 +6883,16 @@ dependencies = [
  "cfg-if",
  "cpufeatures",
  "digest",
+ "sha2-asm",
+]
+
+[[package]]
+name = "sha2-asm"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b845214d6175804686b2bd482bcffe96651bb2d1200742b712003504a2dac1ab"
+dependencies = [
+ "cc",
 ]
 
 [[package]]
--- a/baml_language/Cargo.toml
+++ b/baml_language/Cargo.toml
@@ -303,7 +303,10 @@ rand_xoshiro = "0.6"
 num-bigint = { version = "0.4", features = [ "rand", "serde" ] }
 num-traits = { version = "0.2" }
 rsa = { version = "0.9", default-features = false, features = [ "std", "pem" ] }
-sha2 = { version = "0.10", features = [ "oid" ] }
+# `asm` enables the hardware SHA-2 backend (SHA-NI / ARMv8 crypto extensions,
+# runtime-detected). Without it the warm bytecode-cache path software-hashes
+# ~hundreds of MB of unit payloads per run.
+sha2 = { version = "0.10", features = [ "oid", "asm" ] }
 
 [workspace.lints.rust]
 unsafe_code = "warn"
--- a/baml_language/Cargo.toml
+++ b/baml_language/Cargo.toml
@@ -303,10 +303,11 @@ rand_xoshiro = "0.6"
 num-bigint = { version = "0.4", features = [ "rand", "serde" ] }
 num-traits = { version = "0.2" }
 rsa = { version = "0.9", default-features = false, features = [ "std", "pem" ] }
-# `asm` enables the hardware SHA-2 backend (SHA-NI / ARMv8 crypto extensions,
-# runtime-detected). Without it the warm bytecode-cache path software-hashes
-# ~hundreds of MB of unit payloads per run.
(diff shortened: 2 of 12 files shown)


hellovai:

PR #4100, opened 2026-07-21
perf: parallel cold checks that actually scale + flat LSP latency

hellovai:
Stacked on #4098. Fixes the "12 threads = 1.18× speedup" wall, diagnosed by comparing our salsa usage against ty/ruff's (our driver pattern was already identical to theirs — the problem was one query shape).

## What

1. **Prime wave**: every file's PPIR semantic index is computed across workers *before* the per-file check fan-out. Whole-package aggregates (`package_items`/`namespace_items`) fold over every file's index; on a cold db the first worker to claim that memo computed parse+lower+index for the *entire project* inline on one thread while all other workers parked on the memo's sync slot. Also retires the old first-file warm-up, which warmed the builtin package, not the user package.
2. **Bulk project load**: `add_or_update_files` writes the project file list once (the per-file path re-cloned the whole Vec and bumped the salsa revision per file — O(files²) during load); source reads are parallel with discovery order preserved (FileId assignment and diagnostic ordering unchanged).
3. **LSP**: the diagnostics candidate checks files through the shared order-preserving parallel checker under the held source guard, so every per-edit path gets the parallelism + prime wave.

## End-user impact (measured)

| | before | after |
|---|--:|--:|
| cold pure check, 295k LOC (12 threads) | 7.4 s (~1.2 cores) | **2.8 s** (2.4+ cores) |
| cold check+seed, 295k LOC | 10.5 s | **4.4 s** |
| LSP edit→diagnostics, 250 files | 464 ms | **172 ms** |
| LSP edit→diagnostics, 500 files | 1.7 s | **176 ms** |
| LSP edit→diagnostics, 1000 files | 5.5 s | **209 ms** |

Editor latency is now ~flat in project size.

## Verification

- Diagnostics byte-identical to the previous binary on the error-corpus (cold + warm).
- 198 tests pass across `baml_project` / `bex_project` / `baml_cli`.
- LSP load test: 8+ runs at 250/1000 files with 50-edit multi-file burst storms and 20 s of ~17 edits/s churn — zero crashes, stable RSS (~810 MB peak @1000), post-storm convergence every time. Disclosure: the first two 1000-file runs (only, immediately after a large build finished) showed a warm-up stall that has not reproduced in 8 subsequent attempts including exact-condition replays; the stress harness now captures server log traces to identify the stage if it recurs.

Not in this PR: ty-style snapshot+cancellation for the LSP (blocked on the `panic = "abort"` release profile — cancellation requires unwinding; needs a product decision), and the warm-edit floor (throws-gate + unit loads; separate reviewed design).

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -764,6 +764,7 @@ dependencies = [
  "libsui",
  "mimalloc",
  "object 0.36.3",
+ "rayon",
  "regex",
  "reqwest",
  "sdkgen_cpp",
--- a/baml_language/crates/baml_cli/Cargo.toml
+++ b/baml_language/crates/baml_cli/Cargo.toml
@@ -50,6 +50,7 @@ indicatif = { workspace = true }
 libsui = { workspace = true }
 mimalloc = { workspace = true }
 object = { workspace = true }
+rayon = { workspace = true }
 regex = { workspace = true }
 reqwest = { workspace = true, features = [ "blocking", "json", "rustls" ] }
 serde = { workspace = true }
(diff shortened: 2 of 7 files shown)


hellovai:

PR #4101, opened 2026-07-21
perf(cache): warm no-op checks stop re-proving the cache (draft — review before merge)

hellovai:
**Draft on purpose** — this touches cache-soundness invariants; walk through the risk section below before merging. Stacked on #4100.

## The problem

A warm no-op `baml check` at 295k LOC cost 7.2s — identical to a full recompute — because the warm path re-proved its own cache validity from scratch every run: the serve-time throws gate re-parsed every file (via `package_items`) just to convert types for a comparison, and `plan_reuse` read+hashed+decoded 78MB of unit payloads just to extract the tiny throws fragments it seeds from. Profiled: 70% gate, 20% unit loads.

## The fix (adversarially reviewed design)

The naive fix (compare throws in "fact space") was **killed in adversarial review** — it was either circular or unsound (alias-retarget hole). What's implemented is the reviewed survivor:

1. **No-delta gate skip.** When the dirty partition finds zero changed/added/removed files, the gate is provably a tautology: the values it would compare are byte-for-byte what the manifest was stored with (verbatim persist; content-addressed units written before the manifest), and the solve+conversion are deterministic. Comparing a value to itself cannot demote anything. **Every real edit still takes the full runtime-space gate** — the corner the adversary showed the gate genuinely protects (body-only alias retargets) is untouched.
2. **Seeds from manifest, not units.** `ManifestFile` carries a verbatim copy of each unit's throws-fragment blob; seeding reads that. A no-op invocation opens **zero** unit files. Units load (in parallel, order preserved) only when a delta means relink/re-emit will need them.
3. **Prime wave in the gate path** so the `package_items` derivation (still needed on edits and for package-level diagnostics) is a parallel fold, not a serial parse.
4. `bex_cache`: unit content-key is derived from the integrity digest entry-validation already computed — one hash pass per unit instead of two.

## Measured

| warm scenario | before | after |
|---|--:|--:|
| no-op check, 295k LOC | 7.2 s | **2.0 s** (zero unit reads) |
| no-op check, 59k LOC | 0.54 s | **0.30 s** |
| leaf-edit check | ~unchanged | emit-bound (it seeds the cache; the follow-up `run`/`pack` then hit in ~0.7 s at 295k LOC) |

## Risk analysis (what to scrutinize)

- **Where a bug would live:** the no-delta detection (`dirty == added == removed == ∅`) reuses the same dirty-partition logic every warm compile already trusts; a bug there was already a soundness bug before this PR. The new failure surface is *manifest-fragment drift from the unit copy* — mitigated by copying from the same in-memory bytes at store time, and by re-pointing **both** verify oracles (full `BAML_CACHE_VERIFY` compare + always-on 1-in-32 sampled re-derive) at the manifest copy, i.e. exactly what is served.
- **Deliberate behavior change:** a unit file deleted out from under a warm cache is not detected on a *no-op* run (nothing reads it); the next compile that needs it falls back to a byte-identical full compile. Slower once, never wrong.
- **Wire format:** old manifests fail borsh decode → treated as absent → one full recompile re-seeds. No mixed-version reads (manifest key includes compiler fingerprint).

## Verification run

- 66 cache unit tests green (incl. the poison-tripwire tests, re-pointed at the manifest copy; the missing-unit degradation test now exercises the delta path where detection still lives).
- `BAML_CACHE_VERIFY=1` honest-recompile byte-compare: PASS on no-op and after-edit.
- `BAML_CACHE_SAMPLED_VERIFY=1` forced on 7 consecutive warm invocations across edits: PASS.
- Diagnostics byte-identical to the pre-fix binary, cold and warm.

### DIFF
--- a/baml_language/crates/baml_cli/src/bytecode_cache.rs
+++ b/baml_language/crates/baml_cli/src/bytecode_cache.rs
@@ -671,6 +671,19 @@ pub(crate) struct ReusePlan {
     /// by rel_path. Rehydrated to serve those files' diagnostics without
     /// re-checking, and copied verbatim into the next manifest.
     pub(crate) clean_diagnostics: std::collections::BTreeMap<String, Vec<u8>>,
+    /// Clean files' `CallableThrowsFragment` blobs carried verbatim from the
+    /// previous manifest, by rel_path. The source of the `callable_throws`
+    /// seeds (so seeding needs no unit payloads) and of the next manifest's
+    /// fragment carry.
+    pub(crate) clean_fragments: std::collections::BTreeMap<String, Vec<u8>>,
+    /// True when the dirty partition found nothing changed: no dirty or added
+    /// files, and every manifest entry still present on disk. On this path the
+    /// serve-time throws gate is provably a tautology (the seeds being checked
+    /// are byte-for-byte the values the manifest was stored with, units are
+    /// content-addressed and written before the manifest, and the solve is a
+    /// deterministic pure function), so it is skipped — and unit payloads are
+    /// not loaded at all (nothing will be re-emitted or relinked from them).
+    pub(crate) no_delta: bool,
 }
 
 /// The result of the warm-database preamble ([`CacheContext::prepare_warm_db`]).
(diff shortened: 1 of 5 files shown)


hellovai:

PR #4103, opened 2026-07-21
fix(windows): harden path and URI handling (batch 1)

jubjub727:
## Author Comment

Went through and made the first 3 Windows fixes. Codex found what imo looks like a really elegant solution for replacing `walkdir`. Happy to make any requested changes. It looks good to me :)

## Codex Summary

This PR addresses the first batch of findings from a Windows path and URI audit across CLI discovery, the native LSP/VFS boundary, and the browser editor.

BAML file discovery previously used `walkdir`, skipped excluded directories only after they had already been yielded, followed directory links, and silently discarded traversal errors. It now uses `ignore::WalkBuilder` to prune ignored and hidden directories before descending, respect standard ignore files and negation rules, avoid following directory links, and return traversal failures to callers with context.

Native filesystem paths and internal VFS paths were also being treated as interchangeable strings. This introduces a shared `utils_windows` crate containing a type-safe native path wrapper and the conversion contract between native paths and slash-oriented VFS paths. The conversion preserves Windows drive and UNC roots and is consumed directly by both `bex_project` and `baml_lsp_server`.

The remote Monaco editor previously created a file URI from the server workspace root but continued comparing and joining files using the original native path string. The editor now converts the workspace root to a URI once and derives file URIs, sandbox containment, watcher paths, ancestor directories, and relative filenames from that representation. Workspace-relative file keys are normalized and absolute or escaping paths are rejected.

Path casing is deliberately preserved. Filesystem identity and case-only path aliases are a separate concern and are not addressed by this batch.

Regression coverage verifies:

- Standard ignore files, negation rules, hidden entries, missing roots, and linked directories.
- Windows drive, mixed-separator, verbatim, and UNC native/VFS path conversion.
- Native LSP project loading, discovery, watcher paths, and saves.
- Monaco drive and UNC roots, spaces, `%`, separator variants, containment boundaries, traversal attempts, and URI authority.

Tested with the relevant Rust unit and integration suites, WASM checks, TypeScript tests and typechecks, and a production webview build.

## Summary by CodeRabbit

* **New Features**
  * Added cross-platform workspace path handling for Windows, UNC, and non-Windows environments.
  * Improved workspace file discovery with ignore rules, symlink traversal, and deterministic results.
  * Enhanced editor workspace initialization, file opening, watching, and path validation.

* **Bug Fixes**
  * Prevented invalid or unsafe paths from escaping the workspace.
  * Improved handling of missing files, directories, encoded paths, and conversion errors.

* **Tests**
  * Added coverage for path normalization, sandbox containment, workspace discovery, and editor behavior.


baml_language/crates/baml_workspace/src/discovery.rs
```diff
@@ -1,64 +1,147 @@
...
-
-        // Skip hidden directories and common ignore patterns
-        if path.is_dir() {
-            let name = path.file_name().and_then(|n| n.to_str()).unwrap_or("");
-            if name.starts_with('.') || name == "node_modules" || name == "target" {
-                continue;
-            }
-        }
+    let mut builder = WalkBuilder::new(root);
+    builder.standard_filters(true).follow_links(false);
```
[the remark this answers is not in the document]
jubjub727: This seems like a code rabbit hallucination. Actual behaviour doesn't match what code rabbit claims here. Codex claims with source code evidence and runtime validation that this isn't an issue and following this suggestion would create unintended behaviour.

jubjub727:
Addressed the good code rabbit suggestions and the 1 hallucinated suggestion. Patching together the commits nicely here so there's only 1 commit per bug fix making it easy to reference later on.

baml_language/crates/baml_workspace/src/discovery.rs
sxlijin: we want to follow symlinks

baml_language/crates/baml_workspace/src/discovery.rs
```diff
@@ -1,64 +1,147 @@
...
 
-        // Collect .baml files
-        if path.is_file() && path.extension().and_then(|s| s.to_str()) == Some("baml") {
-            files.push(path.to_path_buf());
+    for entry in builder.build() {
+        let entry = entry?;
```
sxlijin: we should do our best to make progress in a scenario like this, the `filter_map(Result::ok)` before was deliberate

baml_language/crates/utils_windows/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+/// An absolute path in the host operating system's native path domain.
+///
+/// Native paths and VFS path strings are deliberately distinct. On
+/// Windows, VFS paths use `/`, prefix drive paths with `/`, and encode UNC
+/// roots as `/server/share`. Conversions between the two domains belong here
+/// so callers cannot accidentally pass an internal VFS spelling to
+/// [`std::path::Path`]. Component casing is preserved; filesystem identity is
+/// a separate policy owned by the project registry.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NativePathBuf(PathBuf);
```
sxlijin: what specifically does NativePathBuf as a newtype preclude, i.e. what states are unrepresentable because of NativePathBuf?

i'm wondering if having this as a trait makes more sense, since we end up using `std::path::Path` and `PathBuf` on both sides anyways

(i think there's two different ways to impl `NativePathBuf` that more obviously add type safety: (1) have it be `NativePath::UnixPath | NativePath::WindowsPath`, no cfg-only - this matters in e.g. the wasm case, where we run wasm on both windows and unix or (2) wrap VFS, ban direct access to it, and _only_ allow using VFS through NativePathBuf)

typescript2/pkg-editor/src/workspace-path.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  return parts;
+}
+
+/**
+ * Owns the boundary between a host-native workspace root and Monaco's file
+ * URIs. File-map keys remain portable, workspace-relative slash paths; all
+ * absolute identity and containment checks use the URI representation.
+ */
```
sxlijin, line 58: is there a way to set up a CI check that enforces that we always go through this? might be out of scope, just want to brainstorm a way to prevent bug-prone patterns

baml_language/crates/baml_lsp_server/src/native_vfs.rs
```diff
@@ -4,68 +4,103 @@
...
+use utils_windows::{NativePathBuf, NativePathError};
 
-/// Wrapper around `vfs::PhysicalFS` that implements `BulkReadFileSystem`.
-#[derive(Debug)]
-pub struct NativeVfs {
-    root: String,
-}
+/// Native filesystem adapter for BAML's slash-oriented VFS path domain.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct NativeVfs;
```
sxlijin, line 18: so this is always rooted at `/` now?

baml_language/crates/baml_lsp_server/src/native_vfs.rs
```diff
@@ -4,68 +4,103 @@
...
 impl NativeVfs {
     pub fn new() -> Self {
-        Self {
-            root: "/".to_string(),
-        }
+        Self
     }
 
-    fn inner(&self) -> vfs::PhysicalFS {
-        vfs::PhysicalFS::new(&self.root)
```
sxlijin, line 22: wtf why were we constantly constructing and deleting `vfs::PhsyicalFS` instances

baml_language/crates/baml_lsp_server/src/native_vfs.rs
```diff
@@ -4,68 +4,103 @@
...
+        NativePathBuf::from_vfs_path(path).map_err(Self::path_error)
     }
-}
 
-impl Clone for NativeVfs {
-    fn clone(&self) -> Self {
-        Self {
-            root: self.root.clone(),
-        }
+    fn path_error(error: NativePathError) -> vfs::VfsError {
```
sxlijin: should be possible to `impl From<NativePathError> for VfsError`

baml_language/crates/utils_windows/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "utils_windows"
```
sxlijin: might want to call this something else e.g. `utils_fs` if we're going to use it on all platforms

typescript2/pkg-editor/src/workspace-path.ts
jubjub727: I didn't know how to solve this but codex came up with the solution of enforcing low level Uri handling to workspace-path.ts so other files can't avoid it without changing the test first.

This is what codex gave as a potential solution here.

ChatGPT_6LDpxx1eqz

I don't know enough to confidently state that this is the appropriate solution but it does seem like a good solution to me. I'm not that familiar with modern typescript tooling so might be missing something important here.

baml_language/crates/baml_lsp_server/src/native_vfs.rs
jubjub727: It's deliberately handling the semantic conversions between a filesystem path and a vfs path. So filesystem paths are always given a proper /D:/stuff/something.baml path with / as the root when used for vfs but when converting back it becomes D:\stuff\something.baml as Windows expects.

Here's what codex had to say about this:

> It has always used the filesystem root: the previous root field was private and
hardcoded to "/". NativeVfs is now stateless because it converts absolute VFS
paths directly to their corresponding native absolute paths. On Windows that
means drive/UNC roots, rather than literally placing everything beneath "/".

Example behaviour:
- Unix: /repo/file.baml → /repo/file.baml
- Windows drive: /C:/repo/file.baml → C:\repo\file.baml
- Windows UNC: /server/share/file.baml → \\server\share\file.baml

baml_language/crates/baml_lsp_server/src/native_vfs.rs
jubjub727: This was apparently a helper that's no longer needed because the implementation has been changed. It did have a purpose but also it sounds like that purpose was a hacky fix and it's likely for the better that it's gone now. I can give you the full codex explanation if you want to know more but it's not really relevant now that the implementation is changed.

I believe this is directly related to the "/" root you brought up as well.

baml_language/crates/utils_windows/Cargo.toml
jubjub727: Yep agreed using the utils_fs name now

baml_language/crates/utils_windows/src/lib.rs
jubjub727: The actual underlying behaviour here seems to be a bit sticky. It's hard to get a clean solution that also addresses the nuance needed to make Windows work properly. I had Codex have a crack at it and it doesn't look terrible. Its solution is basically just ignoring native paths completely for wasm and ingesting them as strings on that platform which matches what wasm wants here.

I do agree that type safety is important here so I think it is worth getting a proper solution here instead of just accepting an easy solution.

I'm also moving away from the cfg heavy implementation here because it's a bit too fragile and makes the code harder to read imo. Going to split the windows behaviour and non windows behaviour into their own files so it's clearer what's going on here at a glance.

jubjub727:
I've implemented various changes and refactored things to minimize cfg spam and make things a bit more readable.

baml_language/crates/baml_lsp_server/src/native_vfs.rs
sxlijin: let's embed that info in the docstrings for `NativeVfs`

baml_language/crates/baml_path/src/native.rs
```diff
@@ -0,0 +1,90 @@
...
+mod platform;
+#[cfg(windows)]
+#[path = "native/windows.rs"]
+mod platform;
+
+/// An absolute, normalized path in the current host operating system's native
+/// path domain.
+///
+/// This type deliberately does not exist on WASM targets: browser/WASM paths
+/// belong to the VFS domain regardless of the client operating system.
```
sxlijin, line 16: we run the wasm build of baml on all of macos, linux, windows actually!!

when you run `baml playground` and it opens the playground in your browser, that's running a wasm build of baml. you can do this on macos, linux, windows - in all of these, the `baml playground` process is communicating with the open browser tab (the vscode extension uses the same model, just backed by the baml language server)

so it's important that the wasm build can work with both windows paths and macos paths, and we can't just cfg-select the implementation of `utils_fs` that we use

jubjub727: Are you sure this breaks that functionality? Graph view is completely working in the baml playground web server for me using this code. Is there a quick way we could test this and know which way round it needs to be?

Nothing is broken in playground from what I can see on my machine so either the code is right or the code is wrong and I messed up my local build/deploy or the bugs resulting from this behaviour are more complicated than being outright blockers. But this code does seem to work on my machine and I can't find any obvious bugs caused by it.

baml_language/crates/baml_lsp_server/src/native_vfs.rs
jubjub727: Done :)

baml_language/crates/baml_path/src/native.rs
jubjub727: I threw this conversation into codex to get a source grounded answer and it said this:
ChatGPT_woBUokEu4W

sxlijin: Ah, I was remembering the old playground architecture - that's my mistake! I'd forgotten we'd switched that model.

Given that, I'm down to merge this PR as-is then! Looks like there are still some more CI fixes to handle though, around formatting or clippy or something.

---

FWIW, from my codex session:

> 
> Short version: `utils_fs` is useful, but the abstraction boundary is still porous. I would keep it, while pushing for one concrete fix and a tightening follow-up.
> 
> The good part is that it centralizes a real, nontrivial conversion:
> 
> - Windows native paths such as `C:\repo` and `\\server\share` need an explicit mapping into the slash-based VFS namespace.
> - `NativeVfs` now performs that conversion at the physical-filesystem boundary instead of relying on `to_string_lossy()` and separator replacement.
> - The TypeScript workspace-path model similarly handles drive paths, UNC paths, containment, and URI encoding deliberately.
> 
> That is correctness work, not bloat.
> 
> The problem is that `bex_project` still has a competing, unvalidated path type: [`[FsPath(String)](https://github.com/BoundaryML/baml/blob/199bddf35bd985dd65373b0152f5145ebb3e4d91/baml_language/crates/bex_project/src/fs.rs#L11-L31)`](https://github.com/BoundaryML/baml/blob/199bddf35bd985dd65373b0152f5145ebb3e4d91/baml_language/crates/bex_project/src/fs.rs#L11-L31). Its constructor accepts any string, and `as_path()` reinterprets its VFS-form string as a host `std::path::Path`.
> 
> That makes it quite easy to bypass `utils_fs`, and the PR already does so in several places:
> 
> - Project identity and display paths go through `FsPath::as_path().to_string_lossy()`.
> - Relative-path operations use host `Path::strip_prefix` on VFS paths.
> - Protocol filenames are validated using host `Path::components()`, so behavior can differ depending on whether the runtime was compiled for Windows, Unix, or WASM.
> - Compiler database paths are populated with host `PathBuf`s constructed from VFS strings.
> 
> There also appears to be a concrete Windows bug in the cache-seeding path. [`[BexProject::new](https://github.com/BoundaryML/baml/blob/199bddf35bd985dd65373b0152f5145ebb3e4d91/baml_language/crates/bex_project/src/project.rs#L545-L559)`](https://github.com/BoundaryML/baml/blob/199bddf35bd985dd65373b0152f5145ebb3e4d91/baml_language/crates/bex_project/src/project.rs#L545-L559) passes a VFS-form root through `FsPath::as_path()` to `LspSeedCache::load_for_root()`. The seed loader then calls native [`[std::fs::canonicalize](https://github.com/BoundaryML/baml/blob/199bddf35bd985dd65373b0152f5145ebb3e4d91/baml_language/crates/bex_project/src/seed.rs#L250-L278)`](https://github.com/BoundaryML/baml/blob/199bddf35bd985dd65373b0152f5145ebb3e4d91/baml_language/crates/bex_project/src/seed.rs#L250-L278). On Windows, that means handing native filesystem APIs `/C:/repo` instead of `C:\repo`, so cache seeding will normally fail and silently fall back to a cold cache.
> 
> My review position would be:
> 
> 1. Keep `utils_fs` and the `NativeVfs` rewrite.
> 2. Fix the cache-seeding native-boundary violation before merging.
> 3. Make `FsPath` wrap `VfsPathBuf`, or remove it in favor of `VfsPathBuf`.
> 4. Remove or severely restrict `FsPath::as_path()`. VFS identity, joining, containment, relative paths, and display should use VFS semantics. Conversion to `NativePathBuf` should happen only immediately before native disk I/O.
> 
> A concise continuation for the review thread could be:
> 
> > I agree that the WASM build needs to understand paths originating from every host platform, so cfg-selecting one global path interpretation would be wrong. The `VfsPathBuf`/`NativePathBuf` distinction is useful for that reason.
> >
> > My concern is that the distinction currently ends at `BamlVFS`. `bex_project::FsPath` remains an unvalidated string with an `as_path()` method, so VFS paths can still be reinterpreted using the current compilation target’s `std::path` semantics. There are already several examples of that, including passing a VFS-form project root to `LspSeedCache::load_for_root`, which calls native `std::fs::canonicalize`; `/C:/repo` is not the Windows native path `C:\repo`.
> >
> > Could `FsPath` wrap `VfsPathBuf`, with conversion to `NativePathBuf` required explicitly at native-I/O boundaries? Otherwise the new abstraction fixes the main adapter but still makes Windows/VFS bugs easy to introduce elsewhere.
> 
> So: useful abstraction, imperfect integration—not pure bloat.

sxlijin (approved):

baml_language/crates/baml_path/src/native.rs
jubjub727: The things its flagging here might already be addressed in other findings I'm still working through. I'll do another codebase wide pass after all the fixes are in to make sure behaviour is consistent across the codebase.

~~For the CI change are you referring to the typescript CI change? I can go ahead and implement that in the next batch of fixes if you want. I'm just not 100% confident in the solution myself so don't really want to portray the solution as more than it is by trusting Codex. It looks like it's maybe worth while but I'm not entirely sure it's actually going to have the intended effect or if it might just be added bloat. We can just send it and hope Codex is right here imo because it's the type of thing that's easy to remove later on if it's annoying.~~ nvm I see what you mean now, I thought that error was just github being weird because the page was super laggy and bugged when I loaded it (not sure if firefox or github is to blame lol) but I see now what's causing it.

jubjub727: nvm I see what you mean now. I couldn't actually load the error when checking so I thought it was just github being weird but I got codex to actually read the logs and figure things out. Fixing it now.

It seems like the util_fs name is a problem.

PR #4105, opened 2026-07-21
perf: linear cold compiles, one warm cache for every command, flat editor latency

hellovai:
Consolidated stack (supersedes #4098, #4100, #4101 — their commits are all contained here, rebased onto current canary). Seven commits, reviewable independently, in dependency order.

## What this does

A ground-up performance pass on the compiler driven by head-to-head benchmarking against tsgo (native TypeScript) and `go build` on semantically identical generated corpora (1.5k–295k LOC), followed by root-cause profiling of every gap found.

### 1. `perf(compiler2)`: kill the two cold-compile quadratics; parallelize the cache-path check
- The package-wide Jinja type environment was rebuilt inside every `check_file` call (O(N²)) — and twice per compile via `get_bytecode`'s error gate. Now a memoized salsa query (`jinja_base_types`) + skipped entirely for files with no prompts.
- `ProjectDatabase` handle clones deep-copied two N-entry PathBuf maps; the parallel drivers clone ~1.55N handles per compile (quadratic CPU + peak RSS). Now Arc-wrapped, O(1).
- `baml check`/`run` routed through a serial diagnostics collector while the parallel one sat unreachable next to it. Now shared.
- The LSP proves the project error-free once (`get_bytecode_unchecked`), and its diagnostics candidate now includes package-level diagnostics — cross-file name conflicts previously **never reached the editor at all**.

### 2. `feat(cli)`: `baml pack` uses the bytecode cache; `baml check` seeds it
Pack was fully cold on every invocation (no cache participation at all); check reused but never wrote, so check-only workflows never warmed. Now one warm cache serves check → run → test → pack.

### 3–4. `build` + `fix(ci)`: hardware SHA-2 (`sha2/asm`, gated off MSVC), doc-link fix
The warm path hashes ~hundreds of MB of unit payloads per run; it was on the software SHA fallback.

### 5. `perf(compiler2)`: prime wave, bulk project load, parallel LSP sweep
Cold parallel check scaled 1.18× on 12 cores: every file's check demands whole-package aggregates, and the first worker to claim that memo computed parse+index for the entire project inline while all other workers parked on the sync slot (diagnosed against ty/ruff's salsa usage; interventionally verified). Per-file indexes are now primed across workers before fan-out. Project load stops re-writing the file list per file (O(files²)) and reads sources in parallel (FileId order preserved). The LSP candidate uses the shared order-preserving parallel checker.

### 6. `perf(cache)`: warm no-op checks stop re-proving the cache
A warm no-op check cost the same as a full recompute: the throws-gate safety compare re-parsed every file, and seeding read+hashed 78MB of unit payloads for metadata. Now: the gate is skipped when the dirty partition is provably empty (a value compared against itself), throws fragments live verbatim in the manifest so a no-op check opens **zero** unit payloads, and units load in parallel (manifest order preserved) only when a delta means relink will need them. Both verify oracles (`BAML_CACHE_VERIFY` full compare + always-on 1-in-32 sampled re-derive) now check the manifest copy — exactly what is served. Design was adversarially reviewed; the unsound variant (fact-space throws compare) was rejected and every real edit still takes the full runtime-space gate.

### 7. `refactor(cli)`: one shared `ProjectSession` for every project-loading command
check/run/test/pack/describe/grep/generate each hand-rolled a drifting subset of the same preamble — the root cause of "pack forgot the cache" and "describe never primes." One implementation; commands declare `CacheUse::{ReadWrite, ReadWriteTests, ReadOnly, Off}`. Introspection installs seeds only on a no-delta plan (provable tautology) and derives honestly otherwise.

## Measured impact (Apple Silicon, 12 cores, hyperfine)

| | before | after |
|---|--:|--:|
| cold check, 295k LOC | 21–46 s (126 s CPU, 5.3 GB RSS) | **6.6 s** (~13 s CPU, 2.7 GB) |
| cold check, 59k LOC | 3.3 s | **1.24 s** |
| warm no-op check, 295k / 59k | 7.2 s / 3.2 s | **2.0–2.6 s / 0.4–0.6 s** |
| `run` after check (59k / 295k) | full compile | **0.12 s / 0.65 s** (cache hit) |
| `pack` after check (59k / 295k) | 1.7 s / 40.5 s (always cold) | **0.19 s / 0.87 s** (cache hit) |
| LSP edit→diagnostics (250/500/1000 files) | 0.46 / 1.7 / 5.5 s | **179 / 192 / 216 ms** (flat) |
| `describe`, 295k / 59k LOC | 6.3 s / 716 ms | **2.6 s / 402 ms** |

Cold compile scaling is now ~linear in project size (was quadratic: 5× code cost 33× CPU), verified to 8000 modules.

## Verification

- Diagnostics byte-identical to the pre-change binary (error-heavy corpus, cold + warm) at each stage.
- Cache soundness: `BAML_CACHE_VERIFY=1` honest-recompile compare (no-op + after-edit) and forced `BAML_CACHE_SAMPLED_VERIFY=1` across edit sequences — all pass; 43 bytecode-cache oracle tests updated and green.
- LSP load-tested: 50-edit multi-file burst storms + sustained ~17 edits/s churn at 250/1000 files — zero crashes, stable RSS, convergence after every storm.
- Full stack green in CI (doc, Windows incl. MSVC sha2 gating, size gate, all SDK matrices).

## Deliberate trade-offs

- Cold `check` at mid-size projects (~15–30k LOC) is ~10% slower than before: it now emits + stores once to warm the cache; every later check/run/pack is served warm.
- A warm *edit* check re-emits and stores (that is what buys the instant run/pack); at 295k LOC the edit path remains ~6–9 s — the known remaining floor, tracked separately.
- A unit file deleted out from under a warm cache is not detected on a no-op run; the next compile that needs it falls back to a byte-identical full compile.

## Summary by CodeRabbit

* **New Features**
  * CLI commands now share a consistent, cache-aware project session workflow, including lenient read-only sessions for inspection when configuration is missing.
* **Performance**
  * Faster project loading and diagnostics checking via parallel per-file processing.
  * Improved bytecode-cache reuse by avoiding unnecessary payload hashing/loading, including a fast “no-delta” warm path.
* **Bug Fixes**
  * Fixed callable-throws fragment seeding and verification for warm “no-delta” runs.
  * Improved package-level diagnostics handling during compilation and error-gating behavior.
* **Chores**
  * Improved build robustness for cryptography acceleration on MSVC targets.

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -764,6 +764,7 @@ dependencies = [
  "libsui",
  "mimalloc",
  "object 0.36.3",
+ "rayon",
  "regex",
  "reqwest",
  "sdkgen_cpp",
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -6953,6 +6953,16 @@ dependencies = [
  "cfg-if",
  "cpufeatures",
  "digest",
+ "sha2-asm",
+]
+
+[[package]]
+name = "sha2-asm"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b845214d6175804686b2bd482bcffe96651bb2d1200742b712003504a2dac1ab"
+dependencies = [
+ "cc",
 ]
 
 [[package]]
--- a/baml_language/Cargo.toml
+++ b/baml_language/Cargo.toml
@@ -314,10 +314,11 @@ rand_xoshiro = "0.6"
 num-bigint = { version = "0.4", features = [ "rand", "serde" ] }
 num-traits = { version = "0.2" }
 rsa = { version = "0.9", default-features = false, features = [ "std", "pem" ] }
-# `asm` enables the hardware SHA-2 backend (SHA-NI / ARMv8 crypto extensions,
-# runtime-detected). Without it the warm bytecode-cache path software-hashes
-# ~hundreds of MB of unit payloads per run.
-sha2 = { version = "0.10", features = [ "oid", "asm" ] }
+# The hardware SHA-2 backend (`asm`: SHA-NI / ARMv8 crypto extensions,
+# runtime-detected) is enabled per-target in `bex_cache` — the `asm` feature
+# does not build under MSVC (sha2-asm ships GAS `.S` sources), so it must not
+# be turned on here unconditionally.
+sha2 = { version = "0.10", features = [ "oid" ] }
 
 [workspace.lints.rust]
 unsafe_code = "warn"
--- a/baml_language/Cargo.toml
+++ b/baml_language/Cargo.toml
@@ -314,7 +314,10 @@ rand_xoshiro = "0.6"
 num-bigint = { version = "0.4", features = [ "rand", "serde" ] }
 num-traits = { version = "0.2" }
 rsa = { version = "0.9", default-features = false, features = [ "std", "pem" ] }
-sha2 = { version = "0.10", features = [ "oid" ] }
+# `asm` enables the hardware SHA-2 backend (SHA-NI / ARMv8 crypto extensions,
+# runtime-detected). Without it the warm bytecode-cache path software-hashes
+# ~hundreds of MB of unit payloads per run.
+sha2 = { version = "0.10", features = [ "oid", "asm" ] }
 
 [workspace.lints.rust]
 unsafe_code = "warn"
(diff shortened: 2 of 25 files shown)


hellovai:
@coderabbitai resolve

All review items addressed in the latest push: the hawk resolver now goes through `mise which` (honors the pin and MISE_DATA_DIR), `store_unit_shared`'s dedupe probe byte-compares instead of re-hashing, and the seed test exercises a non-empty fragment through the disk round-trip.

PR #4143, opened 2026-07-23
Preserve structural VM values in diagnostics and JSON

aaronvg:
## Root cause

Two VM rendering paths used incomplete fallbacks for values whose runtime shape carried more information than the fallback understood:

- `ErrorContext.to_string()` executes inside a native call and cannot yield to user `ToString` implementations. It therefore rendered class errors as only their class name and replaced other non-string values with `<error>`, losing the values needed to diagnose caught failures.
- typed-union JSON serialization delegated to a runtime-shape walker instead of selecting a union member. That lost the member’s declared-type behavior, so class/media values could be dropped and non-JSON members such as `uint8array` could behave differently inside versus outside a union.

## Scope

- add a VM-heap-allocation-free structural value renderer for native diagnostic paths with depth, node, and active-cycle guards
- render ErrorContext class errors with consistent qualified names, brace spacing, and recursively rendered fields
- structurally render non-class thrown values instead of `<error>`
- keep ErrorContext rendering non-dispatching, so user `ToString` overrides are not invoked from the native call
- select the first union member containing the runtime value with the VM’s canonical ordered membership check, then serialize through that member’s typed path
- raise `JsonSerializationError` when a value matches no member, or when the selected member itself is not JSON-serializable (including `uint8array` without explicit encoding)
- add focused regressions for ErrorContext class/non-class values and unions containing class, enum, and media values

## User impact

Caught errors retain actionable structural detail without risking unbounded diagnostic recursion. `baml.json.to_string<T>()` now preserves typed-member behavior inside unions; consequently, union serialization can fail when the runtime value matches no member or the selected member is not JSON-serializable.

## Validation

- `cargo nextest run -p baml_tests --test errorcontext --test json_typed_union` (19 passed)
- `cargo nextest run -p bex_vm` (61 passed)
- `cargo clippy -p bex_vm --all-targets --all-features -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

---

> [!NOTE]
> **Medium Risk**
> Touches core VM error rendering and typed JSON serialization paths used in diagnostics and `baml.json.to_string<T>()`, with behavior changes for union members and non-class thrown values.
> 
> **Overview**
> Improves how thrown values appear in **ErrorContext** cause chains and stack traces, and fixes **typed union** JSON serialization so it follows declared member types instead of a loose runtime-shape walk.
> 
> **ErrorContext / diagnostics:** Non-string errors no longer collapse to a class name or `<error>`. Native rendering uses a new bounded structural printer (`render_value_structural`) with qualified class names, recursive fields, depth/node limits, and cycle truncation—without calling user `ToString` overrides (native paths cannot yield). Panic snapshots now include payload fields (e.g. `DivisionByZero { dividend: 42 }`).
> 
> **JSON:** `baml.json.to_string<T>()` for unions picks the **first declared member** that contains the value via `value_matches_template` (same as `is` / match), then serializes through that member’s typed path. Class, enum, and media values are preserved; values outside the union or non-JSON members (e.g. `uint8array`) raise `JsonSerializationError` consistently inside and outside unions.
> 
> Regression tests cover ErrorContext rendering, truncation budgets, and typed union JSON (including equivalence with direct class serialization).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94b78bd84527140f196057ca580d980ecc304c18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thrown-error rendering so class-identity and structural fields are preserved in error cause chains, including safe handling for non-string throws and suppression of user `ToString` overrides for class-thrown values.
  * Structural string rendering now safely truncates cyclic and deeply nested structures with clear ellipsis/limit markers.
  * Panic output now includes the panic payload, not just the variant name.
  * Typed JSON union serialization now selects the first compatible union member; invalid values fail with a `JsonSerializationError`.
* **Tests**
  * Added regression coverage for error context rendering, truncation behavior (including cyclic/deep and large arrays), and typed union JSON serialization (success and failure). Updated the panic snapshot.

### DIFF
--- a/baml_language/crates/bex_vm/src/package_baml/error_context.rs
+++ b/baml_language/crates/bex_vm/src/package_baml/error_context.rs
@@ -7,7 +7,7 @@
 
 use std::fmt::Write;
 
-use bex_vm_types::types::{Object, Value};
+use bex_vm_types::types::Value;
 
 use super::{BamlClassErrorsErrorContext, BamlClassErrorsStackTrace, PackageBamlImpl, view};
 use crate::BexVm;
@@ -18,27 +18,6 @@ use crate::BexVm;
 fn render_error_value(vm: &BexVm, value: Value) -> String {
     if let Ok(message) = vm.as_string(&value) {
         return message.to_string();
-    }
-    if let Some(ptr) = value.as_object_ptr()
-        && let Object::Instance(instance) = vm.get_object(ptr)
-        && let Object::Class(class) = vm.get_object(instance.class)
-    {
-        let fields = class
-            .fields
-            .iter()
-            .zip(instance.fields.iter())
-            .map(|(field, value)| {
-                format!(
-                    "{}: {}",
-                    field.name,
-                    super::root::render_value_structural(vm, value.load(), true)
-                )
-            })
-            .collect::<Vec<_>>();
-        if fields.is_empty() {
-            return class.name.to_string();
-        }
-        return format!("{} {{{}}}", class.name, fields.join(", "));
     }
     super::root::render_value_structural(vm, value, false)
 }
--- a/baml_language/crates/bex_vm/src/package_baml/error_context.rs
+++ b/baml_language/crates/bex_vm/src/package_baml/error_context.rs
@@ -12,9 +12,9 @@ use bex_vm_types::types::{Object, Value};
 use super::{BamlClassErrorsErrorContext, BamlClassErrorsStackTrace, PackageBamlImpl, view};
 use crate::BexVm;
 
-/// Render a thrown error value for the chain trace: a string is its own
-/// message; an error instance shows its class name (Python's `Type`); anything
-/// else falls back to a placeholder.
+/// Render a thrown error value for the chain trace. Error instances retain
+/// their qualified class name and fields so a caught-and-reported test failure
+/// is as actionable as an error escaping directly from `baml run`.
 fn render_error_value(vm: &BexVm, value: Value) -> String {
     if let Ok(message) = vm.as_string(&value) {
         return message.to_string();
@@ -23,9 +23,24 @@ fn render_error_value(vm: &BexVm, value: Value) -> String {
         && let Object::Instance(instance) = vm.get_object(ptr)
         && let Object::Class(class) = vm.get_object(instance.class)
     {
-        return class.name.to_string();
+        let fields = class
+            .fields
+            .iter()
+            .zip(instance.fields.iter())
+            .map(|(field, value)| {
+                format!(
+                    "{}: {}",
+                    field.name,
+                    super::root::render_value_structural(vm, value.load(), true)
+                )
+            })
+            .collect::<Vec<_>>();
+        if fields.is_empty() {
+            return class.name.to_string();
+        }
+        return format!("{} {{{}}}", class.name, fields.join(", "));
     }
-    "<error>".to_string()
+    super::root::render_value_structural(vm, value, false)
 }
 
 const CHAIN_SEPARATOR: &str = "\n\nDuring handling of the above error, another error occurred:\n\n";
--- a/baml_language/crates/bex_vm/src/package_baml/json.rs
+++ b/baml_language/crates/bex_vm/src/package_baml/json.rs
@@ -16,7 +16,7 @@
 
 use std::sync::Arc;
 
-use baml_type::{MediaKind, RealizedTy, TypeName};
+use baml_type::{MediaKind, RealizedTy, TyTemplate, TypeName};
 
 /// FQN of the recursive `json` type alias declared in `baml.json`.
 /// Mirrors `baml_base::qualified_name::BAML_JSON_JSON`; inlined here to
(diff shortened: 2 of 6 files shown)


baml_language/crates/bex_vm/src/package_baml/json.rs
```diff
@@ -861,8 +861,14 @@ fn ty_value_to_serde(
...
-            // than trying each member. Matches json-alias union semantics.
-            Ok(value_to_serde(vm, value))
+            // than trying each member. `value_to_serde` is intentionally only
+            // an untyped-json converter; although it already handles enum
+            // variants, class and media instances become null. Use the complete
+            // structural walker so every representable runtime shape survives.
+            // Empty override tables preserve typed `to_string`'s canonical
+            // behavior: custom `ToJson` implementations are not invoked.
+            let mut counter = 0;
+            render_to_serde(vm, value, &[], &[], &mut counter, path)
```
aaronvg: This is the right patch, but I think the fixpoint is **type-directed union serialization via membership testing**, not a shape-directed structural walk:

```rust
RealizedTy::Union(members, _) =>
    // first member m where runtime value ∈ m (same ordered, decidable
    // check BEP-015 match uses — not speculative try-and-backtrack)
    match first_matching_member(vm, value, members) {
        Some(m) => ty_value_to_serde(vm, value, m, path),
        None => Err(raise_serialize(vm, "value is not a member of the union", path, "union")),
    }
```

Why this is more correct than routing through `render_to_serde`:

1. **Coherence by construction**: `to_string<Record>(r) == to_string<Record | Phase>(r)` always, because both go through the same `RealizedTy::Class` arm. As written, the typed-class path (`serialize_class_instance`, declared-type-driven) and the union path (runtime-shape-driven) can diverge for the same value wherever the class path applies type-directed treatment to fields.
2. **Type-directed behavior composes**: media tagging, uint8array encoding rules, alias handling, and any future per-field attributes automatically apply inside unions; the structural walker has to re-implement or silently miss each of these forever.
3. **Honest errors**: a value outside every member is a type-contract violation — raising beats serializing it as *something*.

The existing comment's objection ("dispatch on runtime shape rather than trying each member") conflates speculative try-and-backtrack (bad: error-swallowing, non-deterministic) with membership testing, which is the same ordered, decidable check `match` lowering already uses. Overlapping members resolve first-match, exactly like match arms — a semantics users already hold. TYPE_SYSTEM.md's subset-based unions give you this primitive for free.

The `json` alias correctly stays on `value_to_serde` — it's *defined* as untyped. Fine as a follow-up; this patch is strictly better than what it replaces.

baml_language/crates/bex_vm/src/package_baml/error_context.rs
```diff
@@ -23,9 +23,24 @@ fn render_error_value(vm: &BexVm, value: Value) -> String {
...
+                )
+            })
+            .collect::<Vec<_>>();
+        if fields.is_empty() {
+            return class.name.to_string();
+        }
+        return format!("{} {{{}}}", class.name, fields.join(", "));
     }
-    "<error>".to_string()
+    super::root::render_value_structural(vm, value, false)
```
aaronvg, line 22: For this diagnostic path the governing principle should be **totality over fidelity**: it must never throw, never dispatch user code, and never diverge. The no-user-dispatch invariant is handled (empty override tables — and the test asserting the override is *not* invoked is exactly right), but `render_to_string` has no depth/cycle guard, and class fields are mutable (`AtomicValueSlot`), so a self-referential thrown value would stack-overflow *while rendering an error context*. The old class-name-only rendering was total by construction.

Suggested follow-up: a depth/node budget on `render_value_structural` that degrades to `…` truncation. (This weakness pre-exists in user-facing `to_string`, but ErrorContext is exactly the path that should be crash-proof.)

Two cosmetic notes while here: the top level renders as `user.AppError {message: ...}` (qualified, unpadded braces) while nested instances render as `Detail { label: ... }` (short name, padded braces) — both spellings now appear in one output line and are asserted verbatim in the tests. A `qualified: bool` knob on the walker would remove both the duplication of the instance-rendering loop and the spacing mismatch.

aaronvg (commented):
Both fixes are real fidelity improvements and strictly better than the old behavior (`null` union members, `<error>` placeholders). Leaving a few comments on where the fixpoint is — none of these block the patch.

**Behavior change worth documenting: the union arm is now fallible where it was total.** `value_to_serde` never raised (unrepresentable → `null`); `render_to_serde` raises `JsonSerializationError` for uint8arrays without encoding, functions, futures, etc. So a union member value that previously serialized (badly, as `null`) can now throw from `baml.json.to_string<T>()`. That's arguably the right semantics, but it deserves a line in the PR description and a regression test (e.g. a union member with a `uint8array` field).

**Test suggestion:** an assertion that `to_string<Record>(r) == to_string<Record | Phase>(r)` for a class with a nested typed field would lock in the intended equivalence between the typed-class path and the union path (see inline comment on the union arm for why they can diverge).

Issue #4154, opened 2026-07-23
openai-responses strips prompt_cache_breakpoint content-block metadata (GPT-5.6 prefix caching doesn't work)


conrad-scherb:
## Summary

OpenAI GPT-5.6 models support **explicit prompt cache breakpoints** (`prompt_cache_breakpoint` on an `input_text` content block, plus the request-level `prompt_cache_key` and `prompt_cache_options`). This is the only reliable way to get prefix caching on GPT-5.6 when the prompt suffix varies — without an explicit breakpoint, GPT-5.6 *writes* the cache on every call but never *reads* it on a varying suffix (so you pay the 1.25× cache-write fee and get nothing back).

BAML's `openai-responses` provider strips content-block metadata, so a `prompt_cache_breakpoint` set via `allowed_role_metadata` + `_.role("user", prompt_cache_breakpoint=...)` never reaches OpenAI. This is the exact problem **PR #4014** targeted; it is still reproducible in **0.217.0** and the latest **0.223.0** (PR #4014 was abandoned and never merged).

## Reproducible example

`baml_src/clients.baml`:
```baml
client<llm> LunaResponses {
  provider openai-responses
  retry_policy Constant
  options {
    model "gpt-5.6-luna"
    api_key env.OPENAI_API_KEY
    reasoning { effort "low" }
    allowed_role_metadata ["prompt_cache_breakpoint"]
  }
}
retry_policy Constant { max_retries 1 }
```

`baml_src/test.baml`:
```baml
function EditSentence(stable: string, suffix: string) -> string {
  client LunaResponses
  prompt #"
    {{ _.role("system") }}
    You are a copy editor. Reply with the edited sentence only.

    {{ _.role("user", prompt_cache_breakpoint={"mode": "explicit"}) }}
    {{ stable }}

    {{ _.role("user") }}
    Now edit this sentence: {{ suffix }}
  "#
}
```

Caller: 4 iterations. `stable` is a ~4,000-token block (well above the 1024-token minimum), identical across iterations. `suffix` varies per iteration.

## Expected vs actual

| arm | iter 0 | iter 1–3 |
|---|---|---|
| control — identical full prompt every iter | 0% (cold write) | **100% cached** ✅ |
| varying suffix + `prompt_cache_breakpoint` on stable block | warm | **0% cached** ❌ |

[...]

PR #4155, opened 2026-07-23
fix(openai): preserve content-block metadata on openai-responses (GPT-5.6 prompt_cache_breakpoint)

conrad-scherb:
Revives abandoned PR #4014 (cc @aaronvg per #4154).

## Problem

The `openai-responses` request builder discarded metadata from `ChatMessagePart::WithMeta`, so content-block fields such as GPT-5.6's explicit `prompt_cache_breakpoint` (set via `allowed_role_metadata` + `_.role("user", prompt_cache_breakpoint=...)`) never reached OpenAI.

On GPT-5.6, without an explicit breakpoint the model **writes** the cache on every call but never **reads** it when the prompt suffix varies — so callers pay the 1.25× cache-write fee and get no savings. This makes prefix caching unusable through BAML's `openai-responses` provider today.

## Fix

Extract the inline content-part mapping into a `responses_content_part(part, role, allowed_metadata)` helper and, for `WithMeta` parts, forward **allowlisted** metadata keys onto the content object (dropping non-allowlisted keys, matching the Chat Completions path). Provider-agnostic — no special-casing of `prompt_cache_breakpoint`, GPT-5.6, or OpenRouter.

## Tests

- `test_responses_api_adds_allowed_metadata_to_content_block` (from #4014): helper forwards `prompt_cache_breakpoint` and drops disallowed keys.
- `test_responses_api_build_body_forwards_allowed_metadata_to_content_block` (new): end-to-end `build_body` for the `ResponsesApi` strategy asserts the breakpoint lands on the serialized `input` content block and disallowed keys do not.
- All 13 `openai` tests pass; `cargo fmt --check` and `cargo clippy -p baml-runtime --lib --tests` clean.

## Empirical verification

Verified against the live OpenAI Responses API with a generated BAML client (`provider openai-responses` + `allowed_role_metadata: ["prompt_cache_breakpoint"]` + `_.role("user", prompt_cache_breakpoint={"mode":"explicit"})` on a stable prefix, varying suffix per iteration):

| arm | iter 0 | iter 1–3 |
|---|---|---|
| control — identical full prompt | 0% (cold write) | 100% cached |
| varying suffix + breakpoint | warm | **~99% cached** |

Before this change (BAML 0.217 and 0.223): the varying-suffix arm was **0% cached** on iter 1–3 (identical to implicit-only), because the breakpoint was stripped before reaching OpenAI. Issue #4154 has the full repro.

Refs: #4014, #4154.

---

> **Disclaimer:** This PR was drafted by an AI coding agent and reviewed by a human before submission. The fix is a revival of #4014 (originally AI-agent-drafted by @Yon-Fandorin); the added end-to-end test, clippy/fmt checks, and live-API verification were performed by the agent. The human reviewed the change before opening the PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with the OpenAI Responses API by correctly formatting text, images, audio, and PDF content.
  * Prevented unsupported media and invalid assistant content from being submitted.
  * Restricted forwarded metadata to approved values, preventing unintended metadata from appearing in requests.
* **Tests**
  * Added coverage for metadata filtering and content serialization across message segments.

### DIFF
--- a/engine/baml-runtime/src/internal/llm_client/primitive/openai/openai_client.rs
+++ b/engine/baml-runtime/src/internal/llm_client/primitive/openai/openai_client.rs
@@ -106,6 +106,119 @@ enum ProviderStrategy {
     StandardOpenAI { provider: String },
 }
 
+fn responses_content_part(
+    part: &ChatMessagePart,
+    role: &str,
+    allowed_metadata: &AllowedRoleMetadata,
+) -> Result<serde_json::Value> {
+    match part {
+        ChatMessagePart::Text(text) => {
+            let content_type = if role == "assistant" {
+                "output_text"
+            } else {
+                "input_text"
+            };
+            Ok(json!({
+                "type": content_type,
+                "text": text
+            }))
+        }
+        ChatMessagePart::Media(media) => {
+            // For assistant role, we only support text outputs in Responses API.
+            if role == "assistant" {
+                anyhow::bail!(
+                    "BAML internal error (openai-responses): assistant messages must be text; media not supported for assistant in Responses API"
+                );
+            }
+            match media.media_type {
+                baml_types::BamlMediaType::Image => {
+                    let image_url = match &media.content {
+                        baml_types::BamlMediaContent::Url(url_content) => url_content.url.clone(),
+                        baml_types::BamlMediaContent::Base64(b64_media) => {
+                            format!(
+                                "data:{};base64,{}",
+                                media.mime_type_as_ok()?,
+                                b64_media.base64
+                            )
+                        }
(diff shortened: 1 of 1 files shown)


sxlijin:
Hi there! Thanks for the fix. Are you an existing BAML user blocked by this bug, or can I ask what the motivation for this PR is?

conrad-scherb:
Hi @sxlijin yes we use BAML heavily in our workflows at the moment and we're looking to take advantage of cached input token pricing for GPT 5.6 models, which currently are costing us much more than they need to due to prefix caching not currently working

sxlijin:
Got it! We'll get a release out for you today :)

PR #4156, opened 2026-07-23
fix(release): satisfy the C# product-slice verify contract (#4120 follow-up)

ATX24:
## Problem

Since #4120 merged, **every BAML Language Release run fails** at `Verify C# SDK product package / Compute product consumer matrix`, and because the manifest-publish job is downstream of that gate, **the pkg.boundaryml.com channel manifests have been frozen at `0.15.1-nightly.20260716.b`** — `baml toolchain install nightly` resolves a week-stale toolchain everywhere (and no per-version manifest exists for anything newer, so pinning doesn't work either).

Two gaps between #4120's verifier and what the repo/pipeline actually provide:

1. `release/platforms.json` has **no `artifacts.csharp` blocks**, but the matrix derivation asserts 8 targets with `csharp` + `cffi`, 8 unique RIDs, 3 canonical native assets, and 8 string runners → `jq -e` prints `false` and the job exits 1.
2. The `bridge-cffi-*` producer uploads only the cdylib + `.sha256`, but the packager requires **exactly 3 files** per artifact including a 6-line `artifact-identity.txt` (source_sha / workflow_run_id / workflow_run_attempt / release_version / target / asset) it verifies line-by-line.

## Fix

- **`release/platforms.json`**: add `csharp: { rid, native_asset, consumer_runner }` to all 8 targets. RIDs are the standard .NET RIDs (8 unique); canonical assets are `libbridge_cffi.dylib` / `libbridge_cffi.so` / `bridge_cffi.dll` (3 unique, matching the pre-rename base names in the cffi build and the hard-coded `libbridge_cffi.so` in the verify scripts); consumer runners mirror each target's existing toolchain runner.
- **`build2-bridge-cffi.reusable.yaml`**: emit `artifact-identity.txt` next to the cdylib in the rename/checksum step and add it to the uploaded artifact.

@rossirpaulo — please sanity-check the RID/runner choices against what `publish2-csharp-sdk` expects; happy to adjust.

## Summary by CodeRabbit

* **New Features**
  * Added C# artifact configurations across supported macOS, Linux, and Windows targets.
  * Published native C# assets with platform-specific .NET runtime identifiers.

* **Improvements**
  * Build artifacts now include provenance details such as source revision, workflow run, release version, target, and asset.
  * Artifact uploads include an identity record alongside the library and checksum.

### DIFF
--- a/.github/workflows/build2-bridge-cffi.reusable.yaml
+++ b/.github/workflows/build2-bridge-cffi.reusable.yaml
@@ -191,6 +191,8 @@ jobs:
       - name: Rename and checksum artifact
         id: artifact
         working-directory: baml_language
+        env:
+          SOURCE_SHA: ${{ inputs.source_sha != '' && inputs.source_sha || github.sha }}
         run: |
           set -euo pipefail
           target="${{ matrix._.target }}"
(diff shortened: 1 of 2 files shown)


rossirpaulo:
Do not merge Claude. Bad boy.

PR #4206, opened 2026-07-26
feat(cache): authenticate remote bytecode-cache entries

antoniosarosi:
## Summary

- wrap uploaded remote bytecode-cache entries in an HMAC-SHA256 authenticated envelope
- keep signing and verification secrets local and separate from the bearer credential sent to the blob server
- require authentication before entry validation or deserialization when a verification key is configured, while retaining the existing integrity-only no-key mode

## Root cause

The existing payload SHA-256 was stored alongside attacker-controlled bytes, so a malicious remote could replace both the executable bytecode and its checksum. A keyed MAC known only to writers and readers prevents the blob-serving origin from forging a replacement.

## Validation

- `cargo fmt --all`
- `cargo test -p bex_cache`
- `cargo clippy -p bex_cache --all-targets -- -D warnings`

Fixes B-747

## Summary by CodeRabbit

* **New Features**
  * Added optional HMAC-SHA256 authentication for remote bytecode cache uploads and downloads.
  * Added support for configuring separate signing and verification keys through environment variables.
  * Remote cache entries can now be verified for authenticity before use.

* **Bug Fixes**
  * Tampered or improperly signed remote entries are rejected when verification is enabled.
  * Existing unsigned-cache behavior remains supported when authentication is not configured.

### DIFF
--- a/baml_language/Cargo.lock
+++ b/baml_language/Cargo.lock
@@ -1269,6 +1269,7 @@ dependencies = [
  "baml_version",
  "bex_vm_types",
  "borsh",
+ "hmac",
  "reqwest",
  "rustls",
  "sha2",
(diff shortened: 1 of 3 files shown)


antoniosarosi:
@coderabbitai review

antoniosarosi:

PR #4207, opened 2026-07-26
fix(compiler): support float literal patterns in match

antoniosarosi:
## Summary

- support positive and negative float literals in match patterns, including nested class-field patterns
- keep float literal types outside pattern position intentionally rejected
- document and test the runtime equality and exhaustiveness semantics

## Decision

This PR takes the float-literal pattern support path. Float patterns lower to the existing `==` operation, so they use IEEE 754 equality: `0.0` and `-0.0` match each other, while NaN falls through every float literal arm. Raw `float` remains an infinite domain for exhaustiveness, so a finite set of float literals still requires `_` or a `float` catch-all arm.

## Test plan

- `cargo build -p baml_cli`
- focused parser, TIR exhaustiveness, VM runtime, LSP diagnostic, and BAML bytecode tests
- targeted Clippy with warnings denied
- `cargo fmt --all`
- `baml run -e` repros for `0.0`, `-0.0`, NaN, class-field matching, and non-exhaustiveness

Fixes B-620

## Summary by CodeRabbit

* **New Features**
  * Added support for float literal patterns in match expressions and field patterns.
  * Float matching follows IEEE 754 equality, including `-0.0` behavior; `NaN` does not match literals.

* **Bug Fixes**
  * Corrected exhaustiveness handling so finite float literal patterns still require a catch-all case.

* **Documentation**
  * Clarified float pattern-matching behavior and catch-all requirements.


antoniosarosi:

PR #4209, opened 2026-07-26
fix(compiler): widen element types in Array.concat like array literals

antoniosarosi:
## Summary

- infer Array.concat argument elements independently from the receiver element type
- return the union of receiver and argument element types, matching array-literal widening
- cover primitive unions and spawned futures with distinct error types

## Root cause and impact

Array.concat reused its class-level T for both arrays, so method resolution substituted the receiver element type before checking the argument. A method-level U now lets normal generic inference collect the argument type and produces (T | U)[], allowing idiomatic future collection composition.

Array.push and Array.insert remain invariant because they mutate the receiver; widening an aliased mutable array would be unsound.

## Validation

- cargo build -p baml_cli
- baml run -e primitive concat repro
- baml run -e future error-union concat repro
- cargo test -p baml_tests --lib compiler2_tir::inference::array_concat
- cargo insta test --check -p baml_tests --lib -- compiles::__baml_std__
- cargo fmt --all

Fixes B-617

## Summary by CodeRabbit

* **New Features**
  * Array concatenation now supports combining arrays with different element types.
  * The resulting array automatically reflects the combined element types, including compatible future and error types.

* **Tests**
  * Added coverage for mixed-type arrays and future values with different error outcomes.


antoniosarosi:

PR #4210, opened 2026-07-26
fix(compiler): unify empty-array catch arms with try-block element type

antoniosarosi:
Fixes B-616

Use the try-block type as a soft expectation for synthesized catch arms so compatible empty and nested container literals adopt it without rejecting genuinely different union members.

Tests cover for-in, direct and let-bound use, nested and non-empty array literals, empty maps, and heterogeneous catch results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type inference for `catch` expressions, including empty and nested arrays, maps, and object literals.
  * Catch results now correctly support downstream operations such as iteration and length checks.
  * Incompatible catch-arm types now produce an appropriate union type instead of an erroneous mismatch.

* **Tests**
  * Added coverage for catch-all result type synthesis and nested container inference.

### DIFF
--- a/baml_language/crates/baml_compiler2_tir/src/builder.rs
+++ b/baml_language/crates/baml_compiler2_tir/src/builder.rs
@@ -8338,7 +8338,7 @@ impl<'db> TypeInferenceBuilder<'db> {
         } else {
             self.infer_expr(base_expr_id, body)
         };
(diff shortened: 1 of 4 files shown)


antoniosarosi:

PR #4211, opened 2026-07-26
fix(compiler): align catch effects for to_string fallback

antoniosarosi:
## Summary

- align catch-domain throw inference with the existing total `to_string` fallback semantics
- make plain and match-wrapped `to_string` forms report identical wildcard reachability
- add regressions proving genuinely `unknown`-throwing calls still propagate through match arms

## Root cause

On current canary, both throw walkers already recurse through match arms. The inconsistency came from the catch-specific walker treating an unresolved `recv.to_string()` fallback as an unaccounted call and adding `unknown`, even though that fallback lowers to `string.from(recv)` and is declared `throws never`.

The fix removes that spurious top-level `unknown` instead of injecting one into the match form. As a result, a redundant wildcard now reports E0063 in both ticket forms, removing it is exhaustive when the two declared errors are covered, and genuinely unknown-throwing callbacks still require a wildcard with E0062 in both plain and match-wrapped forms.

## Validation

- `cargo fmt --all`
- `cargo test -p baml_compiler2_tir --lib`
- `cargo test -p baml_lsp2_actions_tests --lib`
- `cargo clippy -p baml_compiler2_tir --lib -- -D warnings`

Fixes B-624

### DIFF
--- a/baml_language/crates/baml_compiler2_tir/src/builder.rs
+++ b/baml_language/crates/baml_compiler2_tir/src/builder.rs
@@ -9365,6 +9365,19 @@ impl<'db> TypeInferenceBuilder<'db> {
                 for arg in args {
                     self.collect_throw_facts_from_expr(arg.expr, body, out);
                 }
+                // Keep catch-domain inference aligned with the whole-body
+                // throws walker: an unresolved `recv.to_string()` call on a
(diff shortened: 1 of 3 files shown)


antoniosarosi:

PR #4212, opened 2026-07-26
fix(compiler): allow reassigning narrowed inferred locals

antoniosarosi:
## Summary

- Preserve the declaration-site inferred type as the stable assignment bound for unannotated locals.
- Keep flow narrowing read-only and update the flow type after assignment.
- Add regressions for valid union reassignment, narrowed and re-widened reads, and invalid assignments.

## Root cause

Unannotated locals only retained their flow-sensitive current type. An is check therefore replaced the only type available to assignment checking, so the narrowed branch rejected values allowed by the original inferred declaration type.

## Validation

- cargo fmt --all
- cargo clippy -p baml_compiler2_tir -p baml_tests --lib -- -D warnings
- cargo test -p baml_tests compiler2_tir::phase7 -- --nocapture
- cargo test -p baml_tests compiler2_tir::inference -- --nocapture
- cargo test -p baml_tests compiler2_tir::phase6 -- --nocapture
- standalone baml-cli repro

Fixes B-618

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type checking for reassignment after type narrowing by preserving an inferred declaration-site type and using it for compatibility validation.
  * Kept narrowed reads from overwriting the stable inferred declaration type, while still allowing narrowing to be reevaluated after assignment.
  * Added more precise `type mismatch` diagnostics when reassigned values fall outside the inferred union type.

* **Tests**
  * Added new Phase 7 tests for reassignment behavior with inferred union types after `is` narrowing.
  * Updated loop syntax test expected diagnostics to include the new mismatch span.

### DIFF
--- a/baml_language/crates/baml_compiler2_tir/src/builder.rs
+++ b/baml_language/crates/baml_compiler2_tir/src/builder.rs
@@ -464,6 +464,14 @@ struct ScopedLocalsSnapshot {
 #[derive(Clone)]
 pub(crate) struct LocalBinding {
     pub(crate) current_ty: Ty,
+    /// Stable assignment bound inferred at an unannotated declaration site.
+    ///
+    /// Unlike `current_ty`, narrowing and assignment never replace this type.
+    /// It is separate from `declared_ty` because explicitly annotated
+    /// containers use their annotation as an expected type when checking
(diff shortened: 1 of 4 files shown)


antoniosarosi:

PR #4217, opened 2026-07-27
Stop caching pnpm in npm publishers

rossirpaulo:
## Summary

- use `actions/setup-node@v6` directly in the Node.js and Web npm-only publishers
- preserve the npm registry configuration required for trusted publishing
- add a release-contract regression test that prevents either publisher from re-enabling the unused pnpm cache

## Why

The `0.15.1-nightly.20260727.a` Node and Web packages were successfully uploaded with OIDC provenance, but both release jobs failed afterward in the shared setup action's pnpm cache post-step. These publisher jobs intentionally install no pnpm dependency tree, so the cache path does not exist and should not be configured.

## Validation

- `python3 -m unittest scripts.tests.test_release_pipeline_contract` (20 tests)
- YAML parsing for both changed workflows
- `actionlint` for both changed workflows (excluding pre-existing SC2055 in untouched Web publish code)
- applicable pre-commit hooks
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Node.js and Web SDK publishing workflows to use a standardized Node.js setup and to explicitly configure npm registry access.
  * Ensured npm publishing no longer relies on pnpm-specific caching behavior.

* **Tests**
  * Added contract tests that verify both npm publishing workflows keep the expected Node.js setup and registry configuration, and do not enable pnpm caching.

### DIFF
--- a/.github/workflows/publish2-nodejs-sdk.yaml
+++ b/.github/workflows/publish2-nodejs-sdk.yaml
@@ -61,14 +61,13 @@ jobs:
           name: nodejs-umbrella-dist
           path: baml_language/sdks/typescript/bridge_typescript/dist
       - name: Setup Node.js
-        # Mirror legacy release.yml publish-to-npm: the composite sets registry-url
-        # (writes the .npmrc that npm/pnpm OIDC trusted publishing keys off of) and
-        # provides pnpm. latest ships an npm new enough for OIDC (checked below).
-        uses: ./.github/actions/setup-node
+        # The publisher only uses npm. Avoid the repository composite here because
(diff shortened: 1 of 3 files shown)


rossirpaulo:
Addressed CodeRabbit’s valid contract-test nit in `3dce767bf`: the test now scopes itself to each workflow’s `publish-npm` job, isolates the actual `Setup Node.js` step, parses that step’s `with:` inputs (including quoted scalars), and asserts the exact registry plus absence of `cache: pnpm`. The 20 release-contract tests and applicable pre-commit hooks pass.

PR #4221, opened 2026-07-27
wip: support keyed BAML runtime instances across bridge_cffi

sxlijin:
## Summary

- replace the process-wide bridge_cffi runtime singleton with a keyed runtime registry on native and WASM targets
- return runtime keys from initialization and route calls through the selected runtime across Python, TypeScript, Web, and the C ABI
- reserve runtime key `0` for generated SDK initialization and generated function calls while dynamically allocating independent nonzero keys
- update C++, C#, Go, Java, Rust, and Swift CFFI consumers for the expanded ABI
- add Python and TypeScript SDK regression coverage proving multiple runtime instances remain independent

## Why

The singleton bridge made dynamically constructed BAML runtimes overwrite one another. Runtime handles now carry a stable `u32` registry key, allowing calls to remain bound to the runtime instance that created them while generated SDKs continue using the well-known key `0`.

## Validation

- `cargo check --workspace`
- `cargo test -p bridge_cffi`
- `cargo test -p sdkgen_python_pydantic2 -p sdkgen_typescript_shared`
- `cargo check -p bridge_cffi -p bridge_typescript_web --target wasm32-unknown-unknown`
- Python SDK `function_calls::pytest`
- TypeScript Node SDK `function_calls::vitest_node`
- TypeScript Web browser and Workers function-call suites
- `pnpm build:debug` for `bridge_typescript`
- `go test ./...` for `sdks/go/baml_go`
- C ABI fixture compilation with warnings treated as errors

C# validation was unavailable because `dotnet` is not installed in the development environment. Swift tests require the local `BamlBridgeFFI.xcframework` binary artifact, which is not present.


sxlijin:
Superseded by #4754, which implements uint64 registrations for distinct generated SDKs and independent dynamic runtimes, with Python and TypeScript coexistence regressions.

PR #4232, opened 2026-07-27
Fix closure resolution in loop headers

codeshaunted:
## Issue Reference

B-532

## Changes

- Record arena-qualified local BindingIds during HIR lexical traversal.
- Consume stable local identities in TIR, MIR, and find-references instead of resolving from source offsets.
- Preserve the canonical binding identity across or-pattern alternatives.
- Add for and while loop-header closure regressions.

## Testing

- [x] Unit tests added
- [x] mise exec -- prek run --all-files
- [x] Focused loop-header closure regressions
- [x] Compiler, LSP, and baml_tests library suites: 1,453 passed, 6 ignored

## PR Checklist

- [x] Read and followed the contributing guidelines
- [x] Code follows the project style guidelines
- [x] Performed a self-review
- [x] No documentation changes are required
- [x] No new warnings

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved closure, lambda, and parameter-default name/path resolution across nested scopes, including correct handling of shadowing and captured locals.
  * Enhanced assignment and call/lvalue lowering so locals are resolved more reliably for both single- and multi-segment paths.
  * Improved local reference finding for usage queries to better reflect actual capture and nesting behavior.

* **Tests**
  * Added lexical scoping regression coverage for closure parameters used in `for` iterables and `while` conditions.
  * Added cursor-based `usages_at` coverage for references inside nested lambda closures.
  * Added prompt runtime coverage for `${ctx.output_format}` interpolation companions.

### DIFF
--- a/baml_language/crates/baml_compiler2_hir/src/builder.rs
+++ b/baml_language/crates/baml_compiler2_hir/src/builder.rs
@@ -26,8 +26,9 @@ use crate::{
     },
     scope::{FileScopeId, ItemScopeOwner, Scope, ScopeId, ScopeKind},
     semantic_index::{
-        BindingId, DefinitionSite, FileSemanticIndex, LocalBinding, PathResolution, ScopeBindings,
-        SemanticIndexExtra, visible_binding_at_in_scopes,
+        BindingId, DefinitionSite, ExprMetadataKey, ExprMetadataScope, FileSemanticIndex,
+        LocalBinding, PathResolution, ScopeBindings, SemanticIndexExtra,
+        visible_binding_at_in_scopes,
     },
 };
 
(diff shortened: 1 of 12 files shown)


baml_language/crates/baml_tests/tests/loop_header_closures.rs
```diff
@@ -0,0 +1,38 @@
+//! Regression tests for closure parameter resolution in loop headers.
+
+use baml_tests::baml_test;
+use bex_engine::BexExternalValue;
+
+#[tokio::test]
+async fn closure_parameter_resolves_in_for_iterable() {
```
aaronvg: we can move these to native baml tests in baml_src

PR #4240, opened 2026-07-28
Fix graph display for message-only errors

sxlijin:
## Summary

- Preserve root failures in graph projections even when the runtime error has no captured value.
- Render a diagnostic-only error preview on the workflow root instead of silently dropping the failure.
- Add regression coverage for an unattached root call with a message-only runtime error.

## Root cause

The graph projection only created a root error value when `RunError.valueRef` existed. Ordinary runtime failures can carry only a message, and root calls are often unattached to CFG overlay entries, so neither the value projection nor runtime decoration surfaced the failure.

## Validation

- `pnpm --filter @b/pkg-playground exec vitest run src/run-store-projections.test.ts`
- `pnpm --filter @b/pkg-playground test` (126 tests)

Fixes B-813.

## Summary by CodeRabbit

* **Bug Fixes**
  * Root-level failed runs without a value reference now display correctly on the graph root.
  * Error previews show an error state, a null value, and the associated runtime diagnostic message.

### Before

Message-only root failures were omitted from the playground graph.

Before: message-only root failure is omitted from the playground graph

### After

The playground graph renders a diagnostic error card on the failed root.

After: playground root renders a diagnostic error card for the plain runtime failure

### DIFF
--- a/typescript2/pkg-playground/src/run-store-projections.test.ts
+++ b/typescript2/pkg-playground/src/run-store-projections.test.ts
@@ -861,6 +861,49 @@ describe('run-store-projections', () => {
     ]);
   });
 
+  it('projects root errors without captured values onto the root graph node', () => {
+    const run = runFixture({
+      status: 'failed',
+      completedAtMs: 150,
+      rootCallNodeId: 'root-main',
+      error: {
+        class: 'Runtime',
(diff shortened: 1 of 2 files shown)


sxlijin:
@coderabbitai review

PR #4248, opened 2026-07-28
Polish playground graph runtime context

sxlijin:
## Summary

- distinguish direct graph runs from newer runs that merely contain the displayed function
- keep trace/profile context for containing workflow runs while refusing to paint their target-function overlay onto a different CFG
- show a compact “last ran inside” banner with navigation back to the containing run target
- surface unattached-call counts and server diagnostics in an expandable graph-canvas notice

## Why

The playground previously selected workflow runs for child-function graphs without preserving how they matched. Because CFG node IDs are function-local, that could paint runtime state and values onto unrelated child nodes. Overlay diagnostics were also invisible, leaving unexplained gaps in graph decoration.

## Validation

- `pnpm --dir typescript2/pkg-playground typecheck`
- `pnpm --dir typescript2/pkg-playground test` (126 tests)
- visual verification in `app-promptfiddle` for the compact and expanded diagnostics notice

Linear: B-822

## Summary by CodeRabbit

* **New Features**
  * Added contextual messaging when the latest graph run was reached “via calls”, including an option to jump to that run.
  * Added an in-graph “notices” panel that surfaces runtime diagnostics and unattached calls, with more details available in the trace view.

* **Bug Fixes**
  * Improved graph run selection so the overlay, calls, status, and errors shown match the correct run context and relationship.

## Before / after

Before — with the same unmatched runtime-call overlay, the exact PR base gives no visible indication that runtime context could not be placed on the graph.

Before: graph with no runtime diagnostics notice

After — the exact PR head surfaces the unmatched call as a compact graph notice and expands it to show the server diagnostic.

After: expanded graph runtime diagnostics notice

### DIFF
--- a/typescript2/pkg-playground/src/ExecutionPanel.tsx
+++ b/typescript2/pkg-playground/src/ExecutionPanel.tsx
@@ -52,6 +52,7 @@ import {
   type ProjectUpdate,
   type TestInfo,
   type Run,
+  type RunTarget,
   type BoundaryId,
   type RunStatus,
   type SourceNavigationTarget,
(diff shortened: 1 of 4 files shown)


sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

PR #4257, opened 2026-07-28
Keep playground form args marker-free

sxlijin:
## Summary

- Keep schema-driven form and raw editor state as ordinary JSON.
- Cast uniquely matched class and enum values only in transient execution payloads.
- Preserve unannotated runtime fallback for ambiguous, invalid, and unsupported values.

## Why

The dynamic args form currently writes `$baml` transport markers into user-visible JSON. The parameter and type schemas already identify most values, so marker generation belongs at the execution boundary rather than editor state.

## Validation

- `pnpm --filter @b/pkg-playground test` (132 tests)
- `pnpm --filter app-vscode-webview typecheck`
- `pnpm --filter app-vscode-webview test:unit:run` (21 tests)
- `pnpm run build:wasm`
- `pnpm --filter app-vscode-webview build`

Linear: B-814

### Before — `$baml` markers in editor state

At the exact PR base (`2960e7a`), the TypeScript2 playground's raw editor persists class and enum transport markers.

Before: raw playground arguments include class and enum $baml markers

### After — marker-free editor with execution-boundary casting

At the exact PR head (`c4ec9b2`), the same fixture and viewport keep ordinary JSON in the raw editor; the successful graph input renders `user.Person`, demonstrating that the known type is applied only at the execution boundary.

After: raw playground arguments are plain JSON while the successful graph input is typed as user.Person

### DIFF
--- a/typescript2/app-vscode-webview/src/ExecutionPanel.strict-mode.test.tsx
+++ b/typescript2/app-vscode-webview/src/ExecutionPanel.strict-mode.test.tsx
@@ -335,14 +335,14 @@ describe('ExecutionPanel args form', () => {
     fireEvent.change(nameInput, { target: { value: 'Ada' } });
     fireEvent.click(await screen.findByRole('button', { name: 'Green' }));
 
-    // Raw view shows the same argsJson the form writes — including the enum
-    // wire marker — proving the form and raw input share one state.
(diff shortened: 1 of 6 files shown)


sxlijin:
@coderabbitai review

PR #4262, opened 2026-07-28
fix(playground): render workflows when selecting tests

sxlijin:
## Summary

- make collected new-style test rows select and render that test body's control-flow graph
- navigate the editor to the selected test declaration
- keep test graphs isolated from the function workflow-prefetch cache

## Validation

- `cargo test -p baml_project`
- `cargo fmt -p baml_project -- --check`
- `pnpm --filter @b/pkg-playground test`
- browser QA in promptfiddle with two collected tests; verified graph and source selection switched independently for each row

Linear: [B-1048](https://linear.app/boundaryml2/issue/B-1048/clicking-on-a-test-should-show-the-workflow-rendered-for-the-test)

## Before / after

**Before:** on the exact PR base, clicking the collected `root::renders workflow` test row does not select it or navigate to its declaration; the playground remains on `Workflow()` and renders only the function graph.

Before: clicking the test row leaves the Workflow function graph and source context unchanged

**After:** on the replacement head, clicking the same row selects the test, navigates the editor to `"renders workflow"`, and renders the test-body graph from `root::renders workflow` through `Workflow(41)` to `Choose result`.

After: selected test row, highlighted test declaration, and rendered test-body workflow graph

### DIFF
--- a/baml_language/crates/baml_project/src/db.rs
+++ b/baml_language/crates/baml_project/src/db.rs
@@ -666,6 +666,119 @@ impl ProjectDatabase {
     ) -> Option<baml_compiler2_visualization::control_flow::ControlFlowGraph> {
         let mut ctx = CfgExpansionCtx::default();
         self.ast_control_flow_graph_impl(function_name, &mut ctx)
+            .or_else(|| self.ast_test_control_flow_graph_impl(function_name, &mut ctx))
+    }
+
+    /// Build a graph for a statically named top-level `test "..." { ... }`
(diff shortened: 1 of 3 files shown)


sxlijin:
Image assets used in the PR description.

**Before**

Before: SimpleWorkflowTest selected while ConditionalWorkflow context remains active

**After**

After: SimpleWorkflow and SimpleWorkflowTest active together

sxlijin:
Image assets used in the PR description for the TypeScript2 replacement implementation.

**Before**

clipboard

**After**

clipboard

PR #4275, opened 2026-07-29
fix(playground): hide default placeholders for explicit args

sxlijin:
## Summary

- hide a declared default expression once a default-capable argument is checked for an explicit value
- keep the declared default visible while the argument is unchecked and using its default
- preserve schema-specific placeholders for required arguments

## Root cause

`ArgsForm` always forwarded `param.defaultExpression` as the native input placeholder, regardless of whether the argument was using its default or had been checked for an explicit value. An empty explicit string therefore looked like the declared default would be sent.

### Before

Before: checked explicit argument still shows the declared default placeholder

### After

After: checked explicit argument has an empty input without the declared default placeholder

## Validation

- `pnpm exec biome check pkg-playground/src/ArgsForm.tsx`
- `pnpm --filter @b/pkg-playground test` (145 tests)
- `pnpm --filter app-vscode-webview typecheck`
- `pnpm --filter app-promptfiddle typecheck`
- `pnpm --filter app-vscode-webview test:unit:run -- ExecutionPanel.strict-mode.test.tsx` (22 tests)
- `git diff --check`

Linear: https://linear.app/boundaryml2/issue/B-1061/default-arg-ui-when-default-args-are-checked-the-placeholder-in-the

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parameter fields showing default-value placeholders even after an explicit value was entered.
  * Placeholders now correctly reflect whether a parameter is using its default or an explicitly provided value.
* **Documentation**
  * Updated the issue status record with reproduction details, validation results, and review progress.


sxlijin:
@coderabbitai review

PR #4277, opened 2026-07-29
feat(cli): feedback pipeline: title/description, zero-prompt sends, auth login, local report store

ATX24:
## Summary

Redesigns `baml feedback` around agents and zero interaction, and adds a gh-issues-style management surface.

- **Payload is `{title, description}`** with hard limits: `--title` one line, max 120 characters, in the form `Issue (feature): description`; `--description` max 500 words (about a page), with help text teaching what to cover (what I was doing, what went wrong, a minimum repro, what I want to happen, syntax ideas). Unknown JSON fields, including the old `issue`/`repro` names, are rejected.
- **Sends never prompt**: anonymous from the get-go under the persisted PostHog distinct id, with a preview of exactly what goes out; after `baml auth login`, reports carry the verified email. Anonymous sends print a notify hint pointing at `baml auth login`. The interactive mode prompt, sticky-anonymous flag, and decline cooldown are removed.
- **`baml auth login`**: login moves under `baml auth`; the top-level `baml login` is removed.
- **Local report store** (`<BAML_HOME>/feedback.json`, 0600): records start `open` and flip to `anonymous`/`reported` on delivery, so offline sends save locally and sync on any later run. Explicit `--anonymous` is stored on the record and honored at sync time with a one-shot id, even if a login exists by then. Management commands: `status`, `list` (`--status`/`--limit`/`--json`), `view` (`--json`), `disable`/`enable` (telemetry-style opt-out). Every event carries a `report_id` UUID for future server-side joins.
- **Bounded storage**: prune to 200 delivered records (never `open` ones), delivered descriptions truncated at 10KB (open records keep full text so deferred sends ship exactly what was written), and a 50-open-report cap where new sends refuse with guidance instead of growing the store unboundedly.
- Works around clap derive's `update_from_arg_matches` erroring on an absent optional subcommand via a manual `FromArgMatches` wrapper.

## Review history

- Two adversarial security passes: the only finding (feedback.json written with default umask while storing the verified email) is fixed with the same 0600 pattern as creds.json.
- Code-quality review found and fixed three sync-path bugs: forced anonymity being re-attributed at sync time, deferred sends shipping truncated descriptions, and delivery status flips matching by collidable 8-char id instead of `event_uuid`.

## Test plan

- 10 e2e tests against an in-process WorkOS/PostHog mock: anonymous-by-default with no prompt, `$identify` merge on login, one-shot ids post-login, forced-anonymity honored across offline sync after login, offline open-then-sync lifecycle, 50-open backlog refusal and drain, disable/enable gating, status/list/view incl. `--json` and `--status` filters, stdin JSON, unknown-field rejection, `baml login` removal.
- Unit tests for payload limits (exact boundaries), truncation char-safety, and prune-never-drops-open.
- Full `baml_cli` suite green; rustfmt clean. Manually exercised end to end against WorkOS staging and real PostHog via a local dev build.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated `baml auth login` (device-code) and updated `baml auth` help/examples for `whoami` and `logout`.
  * Reworked `baml feedback` into a non-interactive command with required `--title`, optional `--description`, repeatable `--files`, local history (`status`, `list`, `view`), and `enable`/`disable`, plus base64 attachment handling and offline queueing.
* **Bug Fixes**
  * Improved auth guidance and messaging across `whoami`, `logout`, token/session-expired flows; tightened login error handling.
  * Secured credentials file writes to owner-only permissions.
* **Tests**
  * Expanded end-to-end coverage for identity merging, offline queueing/sync, attachments, JSON validation, and local feedback history inspection.

### DIFF
--- a/baml_language/Cargo.toml
+++ b/baml_language/Cargo.toml
@@ -231,6 +231,7 @@ regex = { version = "1.10.2" }
 # gitignore-aware directory walking (workspace project discovery in the LSP)
 ignore = { version = "0.4" }
 base64 = "0.22"
+mime_guess = "2"
 bytecount = "0.6"
 reqwest = { version = "0.13.1", default-features = false, features = [
   "stream",
@@ -279,7 +280,7 @@ tracing-subscriber = { version = "0.3.18", default-features = false, features = 
 unicode-normalization = "0.1.25"
 ureq = { version = "3", default-features = false, features = [ "rustls", "gzip" ] }
 url = "2.5.2"
-uuid = { version = "1", features = [ "v4" ] }
+uuid = { version = "1", features = [ "serde", "v4" ] }
 walkdir = { version = "2.4" }
 tsify = { version = "0.5", features = [ "js" ] }
 wasm-bindgen = "0.2"
--- a/baml_language/crates/baml_cli/src/auth.rs
+++ b/baml_language/crates/baml_cli/src/auth.rs
@@ -1,9 +1,9 @@
-// `baml login` / `baml auth` — a human email login via WorkOS CLI Auth.
+// `baml auth` — a human email login via WorkOS CLI Auth (`baml auth login`).
 //
 // This replaces the earlier agent-registration (auth.md) design: identity
 // for feedback is now PostHog-first (see `feedback_command.rs`) — anonymous
 // reporters get a locally generated PostHog distinct id, and logging in
-// exists to attach a verified email to that id. `baml login` runs the
+// exists to attach a verified email to that id. `baml auth login` runs the
 // OAuth 2.0 device authorization grant (RFC 8628): show a one-time code,
 // the user confirms it in a browser, the CLI polls for tokens. On success,
 // if an anonymous distinct id exists, a PostHog `$identify` event merges the
@@ -86,6 +86,8 @@ fn client_id() -> Result<String> {
 
 #[derive(Subcommand, Debug)]
 pub(crate) enum AuthCommands {
+    #[command(about = "Log in to Boundary with your email")]
+    Login(LoginArgs),
     #[command(about = "Show the current identity")]
     Whoami(WhoamiArgs),
     #[command(about = "Log out (keeps your anonymous feedback id)")]
@@ -97,6 +99,7 @@ pub(crate) enum AuthCommands {
 impl AuthCommands {
     pub fn run(&self) -> Result<crate::ExitCode> {
         match self {
+            AuthCommands::Login(args) => args.run(),
             AuthCommands::Whoami(args) => args.run(),
             AuthCommands::Logout(args) => args.run(),
             AuthCommands::Token(args) => args.run(),
@@ -104,22 +107,18 @@ impl AuthCommands {
     }
 }
 
-/// Log in to Boundary with a device-code flow.
-///
-/// Opens a browser to complete authentication, then stores the resulting
-/// session locally. In headless environments, use `--no-open` to print the
-/// verification URL instead.
+/// `baml auth login` — device-code email login.
 #[derive(Args, Debug)]
 #[command(after_long_help = "\
 Examples:
   Log in using a browser:
-    baml login
+    baml auth login
 
   Print the verification URL instead:
-    baml login --no-open")]
+    baml auth login --no-open")]
 pub(crate) struct LoginArgs {
     /// Print the verification URL instead of opening a browser.
-    #[arg(long, help_heading = "Login options")]
+    #[arg(long)]
     pub no_open: bool,
 }
 
@@ -135,7 +134,7 @@ pub(crate) struct LogoutArgs {}
 pub(crate) struct TokenArgs {}
 
 impl LoginArgs {
-    /// Runs `baml login`.
+    /// Runs `baml auth login`.
     ///
     /// Device-code login via WorkOS CLI Auth, then — when an anonymous
     /// PostHog distinct id exists — sends a `$identify` event so all
@@ -148,8 +147,8 @@ impl LoginArgs {
     /// Errors:
     /// - When authorization can't be started, the user denies the login,
     ///   the code expires, or [`LOGIN_TIMEOUT`] elapses.
-    #[allow(clippy::print_stdout)]
     pub fn run(&self) -> Result<crate::ExitCode> {
+        let reporter = crate::reporter::Reporter::new();
         let mut existing = Credentials::read()?.unwrap_or_default();
         // Only skip the flow when the stored session can still produce a
         // token; a session that can't refresh falls through to a fresh
@@ -157,15 +156,18 @@ impl LoginArgs {
         if let Some(email) = existing.user_email.clone() {
             if existing.access_token().is_ok() {
                 existing.write()?;
-                println!("Already logged in as {email}.");
+                reporter.status("Login", format!("already logged in as {email}"));
                 return Ok(crate::ExitCode::Success);
             }
-            println!("The session for {email} has expired; logging in again.");
+            reporter.status(
+                "Login",
+                format!("the session for {email} has expired; logging in again"),
+            );
         }
         let creds = device_login(self.no_open, existing)?;
         match creds.user_email.as_deref() {
-            Some(email) => println!("Logged in as {email}."),
-            None => println!("Logged in."),
+            Some(email) => reporter.finish("Login", format!("logged in as {email}")),
+            None => reporter.finish("Login", "logged in"),
         }
         Ok(crate::ExitCode::Success)
     }
@@ -186,8 +188,8 @@ impl LoginArgs {
 /// Errors:
 /// - When authorization can't be started or the poll ends in denial,
 ///   expiry, or timeout.
-#[allow(clippy::print_stdout)]
 pub(crate) fn device_login(no_open: bool, existing: Credentials) -> Result<Credentials> {
+    let reporter = crate::reporter::Reporter::new();
     let client_id = client_id()?;
     let device: DeviceAuthResponse = post_form(
         &format!("{}/user_management/authorize/device", api_domain()),
@@ -195,25 +197,37 @@ pub(crate) fn device_login(no_open: bool, existing: Credentials) -> Result<Crede
     )
     .context("Failed to start device authorization")?;
 
-    println!("First, copy your one-time code: {}", device.user_code);
+    reporter.status(
+        "Login",
+        format!("first, copy your one-time code: {}", device.user_code),
+    );
     let open_uri = device
         .verification_uri_complete
         .as_deref()
         .unwrap_or(&device.verification_uri);
     if no_open {
-        println!("Then log in at: {}", device.verification_uri);
+        reporter.status(
+            "Login",
+            format!("then log in at: {}", device.verification_uri),
+        );
     } else {
-        println!("Then confirm the login in your browser...");
+        reporter.status("Login", "then confirm the login in your browser...");
         if webbrowser::open(open_uri).is_err() {
-            println!(
-                "Could not open a browser. Log in at: {}",
-                device.verification_uri
+            reporter.status(
+                "Login",
+                format!(
+                    "could not open a browser; log in at: {}",
+                    device.verification_uri
+                ),
             );
         }
     }
-    println!(
-        "Waiting for confirmation... (times out in {} minutes, Ctrl-C to cancel)",
-        LOGIN_TIMEOUT.as_secs() / 60
+    reporter.status(
+        "Waiting",
+        format!(
+            "for confirmation (times out in {} minutes, ctrl-c to cancel)",
+            LOGIN_TIMEOUT.as_secs() / 60
+        ),
     );
 
     let tokens = poll_token_endpoint(
@@ -235,8 +249,6 @@ pub(crate) fn device_login(no_open: bool, existing: Credentials) -> Result<Crede
             .posthog_distinct_id
             .clone()
             .or_else(|| Some(uuid::Uuid::new_v4().to_string())),
-        feedback_anonymous: false,
-        feedback_declined_at: None,
         user_id: user.and_then(|u| u.id.clone()),
         user_email: user.and_then(|u| u.email.clone()),
         access_token: Some(tokens.access_token),
@@ -259,20 +271,20 @@ impl WhoamiArgs {
         match Credentials::read()? {
             Some(creds) if creds.user_email.is_some() => {
                 println!(
-                    "Logged in as {}.",
+                    "logged in as {}",
                     creds.user_email.as_deref().unwrap_or("<unknown>")
                 );
                 Ok(crate::ExitCode::Success)
             }
             Some(creds) if creds.posthog_distinct_id.is_some() => {
                 println!(
-                    "Anonymous (feedback id {}). Run `baml login` to attach your email.",
+                    "anonymous (feedback id {}); run `baml auth login` to attach your email",
                     creds.posthog_distinct_id.as_deref().unwrap_or("<unknown>")
                 );
                 Ok(crate::ExitCode::Success)
             }
             _ => {
-                println!("Not logged in.");
+                println!("not logged in");
                 Ok(crate::ExitCode::Other)
             }
         }
@@ -283,12 +295,12 @@ impl LogoutArgs {
     /// Removes the login session. The PostHog distinct id is preserved so
     /// anonymous feedback continuity — and a future re-login's retroactive
     /// attribution — still work.
-    #[allow(clippy::print_stdout)]
     pub fn run(&self) -> Result<crate::ExitCode> {
+        let reporter = crate::reporter::Reporter::new();
         match Credentials::read()? {
-            None => println!("Not logged in."),
+            None => reporter.status("Logout", "not logged in"),
             Some(creds) if creds.user_email.is_none() && creds.access_token.is_none() => {
-                println!("Not logged in.");
+                reporter.status("Logout", "not logged in");
             }
             Some(creds) => {
                 let anonymous = Credentials {
@@ -296,7 +308,7 @@ impl LogoutArgs {
                     ..Credentials::default()
                 };
                 anonymous.write()?;
-                println!("Logged out.");
+                reporter.status("Logout", "logged out");
             }
         }
         Ok(crate::ExitCode::Success)
@@ -304,16 +316,19 @@ impl LogoutArgs {
 }
 
 impl TokenArgs {
-    #[allow(clippy::print_stdout, clippy::print_stderr)]
+    #[allow(clippy::print_stdout)]
     pub fn run(&self) -> Result<crate::ExitCode> {
+        let reporter = crate::reporter::Reporter::new();
         let mut creds = match Credentials::read()? {
             Some(creds) => creds,
-            None => {
-                eprintln!("Not logged in. Run `baml login`.");
-                return Ok(crate::ExitCode::Other);
-            }
+            None => return Ok(reporter.fatal("not logged in; run `baml auth login`")),
         };
-        println!("{}", creds.access_token()?);
+        // The token itself is this command's data output (stdout); auth
+        // failures are fatal diagnostics.
+        match creds.access_token() {
+            Ok(token) => println!("{token}"),
+            Err(e) => return Ok(reporter.fatal(e)),
+        }
         creds.write()?;
         Ok(crate::ExitCode::Success)
     }
@@ -383,7 +398,7 @@ fn poll_token_endpoint(
             }
             "access_denied" => anyhow::bail!("Login was denied in the browser."),
             "expired_token" => anyhow::bail!(
-                "The confirmation code expired before it was used. Run `baml login` again."
+                "the confirmation code expired before it was used; run `baml auth login` again"
             ),
             _ => anyhow::bail!("Auth server returned {status}: {value}"),
         }
@@ -455,29 +470,6 @@ fn form_urlencode(s: &str) -> String {
     out
 }
 
-/// Prints a prompt and reads one trimmed line from stdin.
-///
-/// Errors:
-/// - When stdin is closed (EOF — e.g. a piped or non-interactive session),
-///   so retry loops fail cleanly instead of spinning.
-#[allow(clippy::print_stdout)]
-pub(crate) fn prompt(message: &str) -> Result<String> {
-    use std::io::Write as _;
-    print!("{message}");
-    std::io::stdout().flush().ok();
-    let mut line = String::new();
-    let bytes_read = std::io::stdin()
-        .read_line(&mut line)
-        .context("Failed to read input")?;
-    if bytes_read == 0 {
-        anyhow::bail!(
-            "Standard input closed before a response was entered. \
-             This command needs an interactive terminal."
-        );
-    }
-    Ok(line.trim().to_string())
-}
-
 // ---------------------------------------------------------------------------
 // Credential storage: ~/.baml/creds.json
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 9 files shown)


baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -222,50 +387,64 @@ impl FeedbackArgs {
...
+        if let Some(description) = obj.get("description").and_then(Value::as_str) {
+            let words = description.split_whitespace().count();
+            if words > MAX_DESCRIPTION_WORDS {
+                anyhow::bail!(
+                    "Description is {words} words; the limit is \
+                     {MAX_DESCRIPTION_WORDS}. Trim it, or link to a gist for \
+                     the long version."
+                );
+            }
+        }
```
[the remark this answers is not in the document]
ATX24: Fixed in 00c99117f: payload() now validates that title/description are strings before the whitelist size checks, with a unit test covering both the object-description and numeric-title cases.

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +514,463 @@ impl FeedbackArgs {
...
 }
 
-enum ReportMode {
-    Anonymous,
-    Email,
-    /// Declined at the prompt: nothing is sent, nothing is persisted.
-    DontSend,
+/// A minimal extension-based mime guess; unknown extensions fall back to
+/// octet-stream (the Boundary side mostly cares about image-vs-text).
+fn mime_for(name: &str) -> &'static str {
```
ATX24: seems like poor code

baml_language/crates/baml_cli/src/auth.rs
```diff
@@ -97,29 +99,23 @@ pub(crate) enum AuthCommands {
...
 #[derive(Args, Debug)]
 #[command(after_long_help = "\
 Examples:
-  Log in using a browser:
-    baml login
-
-  Print the verification URL instead:
-    baml login --no-open")]
+  baml auth login
+  baml auth login --no-open")]
```
codeshaunted: examples formatting should match what the other commands look like

with bolded headings and then normal weight commands

i dont think this will look the same

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -1,31 +1,63 @@
...
+use clap::{Args, Subcommand, ValueEnum};
+use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
 
 use crate::auth::{self, Credentials};
 
 /// Feedback events must stay comfortably inside PostHog's per-event limits.
 const MAX_PAYLOAD_BYTES: usize = 256 * 1024;
 
+/// Cap on the description, in words: about one single-spaced page (a page
```
codeshaunted: seems wrong to enforce these client side

having been on the other side of this, client side enforcement is one of the first things i would try to abuse lol

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -35,119 +67,235 @@ fn posthog_host() -> String {
...
+
 #[derive(Args, Debug)]
+#[command(args_conflicts_with_subcommands = true)]
 #[command(after_long_help = "\
 Examples:
-  Report an issue:
-    baml feedback --issue \"parser panics on nested unions\"
-
-  Submit an anonymous report with a reproduction:
-    baml feedback --anonymous --issue \"...\" --repro \"class A { ... }\"
```
codeshaunted, line 49: just make sure the formatting of the help command looks good in both short form and long form modes and matches my PR from yesterday

ideally maybe we have an LLM lint like sam's bridge one that lets us enforce this, because it will be hard to do manually

codeshaunted: or we could make a custom derive macro or something that does this for us

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -35,119 +67,235 @@ fn posthog_host() -> String {
...
+            }) => return run_list(&store, *status, *limit, *json),
+            Some(FeedbackAction::View { id, json }) => return run_view(&store, id, *json),
+            Some(FeedbackAction::Disable) => return run_set_enabled(&mut store, false),
+            Some(FeedbackAction::Enable) => return run_set_enabled(&mut store, true),
+            None => {}
+        }
+
+        if !store.enabled {
+            eprintln!(
+                "Feedback is disabled on this machine. Run `baml feedback enable` \
```
codeshaunted: the convention i've been going with is that most all output should be lowercased and be a single sentence, just for consistency and brevity

also we shouldn't use `eprintln!` directly, we should be using reporter from `baml_exec`, but claude keeps forgetting that exists

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -35,119 +67,235 @@ fn posthog_host() -> String {
...
+
+        // The sync above is the only thing that shrinks the open backlog;
+        // if it couldn't, refuse to queue more rather than grow the store
+        // without bound on a machine that can never reach PostHog.
+        let open_count = store
+            .reports
+            .iter()
+            .filter(|r| r.status == ReportStatus::Open)
+            .count();
+        if open_count >= MAX_OPEN_REPORTS {
```
codeshaunted: dont think this should be enforced client side

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -162,36 +310,62 @@ impl FeedbackArgs {
...
+                } else {
+                    ReportStatus::Anonymous
+                };
+                let delivered_email = identified.then(|| creds.user_email.clone()).flatten();
+                if let Some(r) = store.reports.iter_mut().find(|r| r.event_uuid == report_id) {
+                    r.mark_delivered(delivered_status, delivered_email.as_deref());
+                }
+                store.save()?;
+                if identified {
+                    println!(
```
codeshaunted, line 321: again should probably use our reporter struct, but we also need to decide on what ends up in stdout and what ends up in stderr

we're inconsistent with this right now and we haven't really codified a standard

baml_language/crates/baml_cli/src/feedback_command.rs
codeshaunted: we have a crate to handle this i think

codeshaunted (dismissed):

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -231,54 +390,93 @@ impl FeedbackArgs {
...
-            .map(|s| s.len())
-            .unwrap_or(0);
+        let size = serde_json::to_string(&from_json)?.len();
         if size > MAX_PAYLOAD_BYTES {
             anyhow::bail!(
                 "Feedback payload is {size} bytes; the limit is {MAX_PAYLOAD_BYTES}. \
-                 Trim the repro/context (a link to a gist works well)."
+                 Trim the description (a link to a gist works well)."
             );
         }
```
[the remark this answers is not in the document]
ATX24: Superseded by the review sweep and fixed in 3afb9870d: the per-file/count caps are removed entirely (client-side enforcement dropped per codeshaunted's review), the single remaining guard is the event ceiling, and its message now says "trim the description or attachments".

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -231,54 +369,95 @@ impl FeedbackArgs {
...
+                let path = entry
+                    .as_str()
+                    .ok_or_else(|| anyhow::anyhow!("\"files\" entries must be path strings"))?;
+                file_paths.push(path.into());
+            }
+        }
+        file_paths.extend(self.files.iter().cloned());
+        if !file_paths.is_empty() {
+            obj.insert("files".into(), encode_attachments(&file_paths)?);
+        }
```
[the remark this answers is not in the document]
ATX24: Fixed in 3afb9870d: `files` is no longer accepted from JSON payloads at all; attachment paths come only from the trusted `--files` flag. A piped payload naming paths now fails the whitelist with guidance, and a unit test pins the behavior.

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +466,433 @@ impl FeedbackArgs {
...
+        }));
     }
-}
-
-enum ReportMode {
-    Anonymous,
-    Email,
-    /// Declined at the prompt: nothing is sent, nothing is persisted.
-    DontSend,
+    Ok(Value::Array(encoded))
```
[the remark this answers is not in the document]
ATX24: Fixed in 3afb9870d: attachment size is checked from fs::metadata before the file is read, so an oversized file errors without being buffered or encoded.

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +466,433 @@ impl FeedbackArgs {
...
+            &record_distinct_id,
+            &record.payload(),
+            record_identified,
+            &record.event_uuid,
+        )
+        .is_err()
+        {
+            // Still unreachable; keep it open and stop hammering.
+            break;
+        }
```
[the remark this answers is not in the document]
ATX24: Fixed in 3afb9870d: deferred deliveries are capped at MAX_SYNC_PER_RUN=25 per invocation; the remainder syncs on later runs.

baml_language/crates/baml_cli/tests/feedback_auth_e2e.rs
```diff
@@ -159,35 +173,53 @@ fn feedback_events(state: &MockState) -> Vec<Value> {
         .collect()
 }
 
+/// A dead endpoint: reserved port on localhost with nothing listening.
+fn dead_posthog() -> String {
+    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+    let addr = listener.local_addr().unwrap();
+    drop(listener);
+    format!("http://{addr}")
+}
```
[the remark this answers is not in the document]
ATX24: Fixed in 3afb9870d: dead_posthog now keeps the listener alive and accept-then-drops every connection, so requests fail fast without the port ever being reusable by a concurrent test.

baml_language/crates/baml_cli/src/auth.rs
ATX24: done in 3afb9870d: auth login and feedback examples now use the labeled heading + indented command format from #4267, and I verified them through `baml help auth login` / `baml help feedback` rendering.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: agreed and removed in 3afb9870d: the per-file/count caps and the open-report send refusal are gone. The one check kept is the ~256KB event ceiling, since PostHog rejects oversized events server-side and we'd rather fail with a clear error than have the event silently dropped; that one's error handling, not enforcement. Worth saying out loud: with a public write-only ingest key there is no real enforcement surface at all; anything abuse-resistant needs a Boundary ingestion API in front (each event carries a report_id to make that migration possible later).

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: checked both: -h shows the concise form and `baml help feedback` the detailed one with the labeled examples. An LLM lint (or derive macro) for help conventions sounds worth a separate issue; happy to draft one.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: done in 3afb9870d: all output in feedback_command.rs and auth.rs is now lowercase single sentences, and error/warning paths go through reporter (print_error/print_warning) instead of raw eprintln.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: removed in 3afb9870d (see thread above); with one adjustment from CodeRabbit's follow-up: sync now batches at 25 deferred deliveries per invocation so a long-offline backlog can't turn one command into hundreds of blocking posts.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: done for the error/warning paths (reporter); data output (preview, status/list/view, --json) stays on stdout so it pipes cleanly, which seems like the natural line for the stdout/stderr standard: data on stdout, diagnostics on stderr. Happy to write that up if we want to codify it.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: yep; swapped to mime_guess (already in the tree via reqwest), with a UTF-8-content fallback to text/plain for extensions it doesn't know (.baml et al).

baml_language/crates/baml_cli/src/auth.rs
```diff
@@ -157,15 +156,15 @@ impl LoginArgs {
         if let Some(email) = existing.user_email.clone() {
             if existing.access_token().is_ok() {
                 existing.write()?;
-                println!("Already logged in as {email}.");
+                println!("already logged in as {email}");
                 return Ok(crate::ExitCode::Success);
             }
-            println!("The session for {email} has expired; logging in again.");
+            println!("the session for {email} has expired; logging in again");
```
codeshaunted: ideally we use reporter, not raw `println!`

baml_language/crates/baml_cli/src/auth.rs
```diff
@@ -590,7 +554,7 @@ impl Credentials {
     ///   refreshed.
     pub fn access_token(&mut self) -> Result<&str> {
         if self.access_token.is_none() {
-            anyhow::bail!("Not logged in. Run `baml login`.");
+            anyhow::bail!("not logged in; run `baml auth login`");
```
codeshaunted, line 580: i cant remember if i made a reporter thing for fatal errors, but if i didnt you should add one and use reporter here

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -24,8 +29,19 @@ const MAX_PAYLOAD_BYTES: usize = 256 * 1024;
 
 const FEEDBACK_EVENT: &str = "baml_feedback";
 
-/// How long a "Don't send" choice suppresses the prompt.
-const DECLINE_COOLDOWN_SECS: u64 = 24 * 60 * 60;
+/// Cap on the local store: newest records win, but never prune undelivered
+/// (`open`) reports.
+const MAX_STORED_REPORTS: usize = 200;
```
codeshaunted: should we even cap it tbh? this is a skill issue on their agents' part lol

codeshaunted: i think it adds more complexity to cap this stuff locally

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -215,70 +362,95 @@ impl FeedbackArgs {
...
-            .map(|s| s.len())
-            .unwrap_or(0);
+        let size = serde_json::to_string(&from_json)?.len();
         if size > MAX_PAYLOAD_BYTES {
             anyhow::bail!(
-                "Feedback payload is {size} bytes; the limit is {MAX_PAYLOAD_BYTES}. \
-                 Trim the repro/context (a link to a gist works well)."
+                "feedback payload is {size} bytes and the event ceiling is \
+                 {MAX_PAYLOAD_BYTES}; trim the description or attachments (a \
+                 link to a gist works well)"
```
codeshaunted: yeah see this is just too much stuff for us to care about

just dont cap it, we can add server-side caps later like vbv said

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -215,70 +362,95 @@ impl FeedbackArgs {
...
+        // files. Attachments come only from the `--files` flag (trusted CLI
+        // input).
         let unknown: Vec<&str> = obj
             .keys()
             .map(String::as_str)
-            .filter(|k| !matches!(*k, "issue" | "repro" | "context"))
+            .filter(|k| !matches!(*k, "title" | "description"))
             .collect();
         if !unknown.is_empty() {
             anyhow::bail!(
```
codeshaunted: tbh not a fan of anyhow bail since it's less explicit than just returning an err variant but that's a nit

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+struct FeedbackRecord {
+    /// Short id shown in `list`/`view` (a prefix of `event_uuid`).
+    id: String,
+    /// Full uuid, sent as the `report_id` event property so server-side
+    /// state can be joined to this record later.
+    event_uuid: String,
```
codeshaunted: uuid probably shouldnt just be a string and we should use a library like the uuid crate

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+    /// delivery must honor it even if a login exists by sync time.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    forced_anonymous: bool,
+    /// The verified email the report was (or will be) attributed to.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    email: Option<String>,
+    /// Attachments. `content_base64` is kept while `open` (a deferred
+    /// delivery must ship it) and dropped once delivered, leaving only the
+    /// name/mime/size metadata for `view`.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
```
codeshaunted: do we really need to skip this if it's empty? nit

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+    created_at: u64,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+struct StoredFile {
+    name: String,
+    mime: String,
+    size_bytes: u64,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    content_base64: Option<String>,
```
codeshaunted: instead of doing this you could have this literally just be a byte stream or file that gets serialized base64 with some serde thing; nit

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+            let truncated = truncate_for_store(description);
+            self.description = Some(truncated);
+        }
+        for file in &mut self.files {
+            file.content_base64 = None;
+        }
+    }
+}
+
+const TRUNCATION_MARKER: &str = "… [truncated locally; the full text was sent]";
```
codeshaunted: too much effort to do this, adds complexity

just dont truncate it and let them send whatever

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+    }
+
+    fn save(&mut self) -> Result<()> {
+        self.prune();
+        let path = Self::path();
+        if let Some(dir) = path.parent() {
+            std::fs::create_dir_all(dir)
+                .with_context(|| format!("Failed to create {}", dir.display()))?;
+        }
+        let content = serde_json::to_string_pretty(self)?;
```
codeshaunted: does it need to be pretty printed?

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+                .with_context(|| format!("Failed to write {}", path.display()))?;
+            std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o600))?;
+            file.write_all(content.as_bytes())
+                .with_context(|| format!("Failed to write {}", path.display()))?;
+        }
+        #[cfg(not(unix))]
+        std::fs::write(&path, content)
+            .with_context(|| format!("Failed to write {}", path.display()))?;
+        Ok(())
+    }
```
codeshaunted: what's up with the unix specific handling here? seems hacky

baml_language/crates/baml_cli/src/feedback_command.rs
```diff
@@ -287,43 +459,450 @@ impl FeedbackArgs {
...
+    } else {
+        println!("Reports:");
+        for r in &store.reports {
+            println!("  [{}] {}: {}", r.status, r.id, r.title);
+        }
+    }
+    Ok(crate::ExitCode::Success)
+}
+
+#[allow(clippy::print_stdout)]
```
codeshaunted: you should never have to annotate a function with this

baml_language/Cargo.toml
```diff
@@ -235,6 +235,7 @@ regex = { version = "1.10.2" }
 # gitignore-aware directory walking (workspace project discovery in the LSP)
 ignore = { version = "0.4" }
 base64 = "0.22"
+mime_guess = "2"
```
codeshaunted, line 234: there was some other crate we used for this

ideally we dont add more crates

codeshaunted: if im wrong and we removed that crate, disregard

make sure any crate you add is actively maintained... this one last had an update 10 months ago

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: removed in 42a861328: no local store cap at all anymore (prune deleted).

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: done in 42a861328: the payload ceiling is gone; the only remaining size behavior is PostHog's own server-side rejection, and we can add real server caps later per vbv.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: removed in 42a861328: no truncation; delivered records keep the full description. Attachment bytes are still dropped after delivery since re-sending is the only thing that needed them.

baml_language/crates/baml_cli/src/auth.rs
ATX24: done in 42a861328: login/logout/device-flow output goes through Reporter (status/finish); whoami keeps stdout since it's data output.

baml_language/crates/baml_cli/src/auth.rs
ATX24: there wasn't one, so added Reporter::fatal(msg) -> ExitCode in 42a861328 and used it for the token command's not-logged-in/expired paths.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: fair; switched the payload-validation bails to explicit return Err(anyhow!(...)) in 42a861328.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: done in 42a861328: event_uuid is uuid::Uuid (serde feature enabled at the workspace).

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: dropped the skip in 42a861328; empty files arrays serialize as [].

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: done in 42a861328: StoredFile.content is Option<Vec<u8>> with a small serde base64 adapter (field still serializes as content_base64), and the event json encodes from the same bytes.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: no; switched to compact to_string in 42a861328.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: deduped in 42a861328: feedback.json now uses the same write_owner_only helper as creds.json (0600 on unix, default perms elsewhere) instead of an inline cfg block. If we'd rather not restrict it at all, that's a one-line change, but it stores the verified email so owner-only seemed right.

baml_language/crates/baml_cli/src/feedback_command.rs
ATX24: agreed; the per-function allows are gone in 42a861328; one module-level allow with a justification comment, matching telemetry_command.rs.

baml_language/Cargo.toml
ATX24: checked the tree: mime_guess (and mime) are already transitive deps of reqwest, pinned in Cargo.lock at 2.0.5, so this adds no new crate to the build graph, just a direct reference. On maintenance: it's a thin wrapper over a static extension registry, so it being quiet for 10 months reads as finished rather than abandoned. If we'd still rather not take the direct dep, the UTF-8-content fallback alone could replace it entirely; say the word.

PR #4324, opened 2026-08-05
feat(typescript): compiler-backed TypeScript tooling for direct BAML imports (opt-in v1)

roninjin10:
## What this adds

Opt-in v1 TypeScript tooling for importing BAML directly — `import { b, Resume } from './baml_src/resume.baml'` and `import { b } from 'baml:client'` — with cross-language editor support (definition, references, rename, diagnostics, completion, hover) and build integration for Vite, Rollup, Webpack, esbuild, and Bun. The compiler owns all BAML semantics; TypeScript hosts only snapshot sources and translate compiler-owned spans. `baml generate` and every v0 surface are unchanged.

## Series narrative (restructured)

The series is rebuilt documentation-first so each layer can be reviewed against the contract above it:

1. **chore scaffolding** — temporary automation, removed by the last commit (final tree has zero `.smithers/` paths).
2. **📝 docs** — the architecture: compiler-owned semantics, package boundaries, project ownership, caching/HMR, sidecars.
3. **✨ spec** — the `baml.tooling.v1` protobuf, generated bindings, sync script: the wire contract before any behavior.
4. **✨ API stubs (Rust + TS)** — the entire public surface with `todo!()`/throwing bodies; compiles standalone (`cargo check`, `tsc`), no tests by design.
5. **Area implementations, each atomic with its tests** — SDK emission → rename actions → compiler project + bridge sessions → TS project + sidecar → unplugin adapters → language-service features. Tests are never a separate commit from the code they prove.
6. **✅ acceptance** — packed/tsserver/backend-parity suites that are genuinely cross-cutting.
7. **👷 ci**, then scaffolding removal.

## Review findings resolved (adversarial review, both seats Not-LGTM → fixed)

| Finding | Resolution |
|---|---|
| **B1** watch mode poisons compiler DB | Filtered to `.baml`/`baml.toml` in `build-core.watchChange` + Vite `handleHotUpdate` early-return (765dbc5cc), defensive `NonBamlPath` rejection in `ProjectSession::update_file` (77cafde14), tests at both layers |
| **M1** tsserver virtual-file identity | Hash the importer-resolved physical source, not the raw specifier (f9602af3b); unit + real-language-service e2e |
| **M2** overlay lifecycle frozen at open | Add/remove rediscovery on every sync, `updateFile(path, null)` for deletions, live `layout()` refresh (f9602af3b + 64d91283e); e2e add/delete scenarios |
| **M3** WASM same-root session clobber | Per-instance `BamlToolingBridge` on WASM, matching napi parity (77cafde14); Rust protocol isolation test + e2e on both backends |
| tooling.rs bare-name symbol join + zip truncation | Join by stable symbol identity; drifting signatures skipped, never truncated (1647b8159) |
| m1–m8, m11 | Fixed in owning commits (tombstone versions, toTextSpan guard, lexical path normalization, JS-safe string escapes, no production dev-flip, reference flags preserved, member hover, session dispose, env-var doc note) |
| m9–m10 | Closed with justification in commit-review Resolution notes (cache eviction deferral documented; perf-only, math verified) |

The verified-sound list from the reconciled review was not churned. Per-commit reviews with per-issue resolutions live in the review deck (link below); every commit reaches LGTM by the code being clean.

## Round-2 review findings resolved (both seats Not-LGTM → fixed)

Head: `cfb9ac80b`. Every fix is folded into its owning commit; each behavior change lands atomically with a test reproducing the reviewer's scenario.

| Finding | Resolution |
|---|---|
| **R2-MAJOR-1** tsserver cannot serve dependency/workspace `.baml` imports | Bare specifiers now go through Node package resolution host-side (`resolveBamlSpecifier`, respecting pnpm symlinks via realpath) in `9d75e5de1`; the plugin keeps one compiler session per owning root — the app's project plus a lazily opened, single-flighted lane per dependency/sibling — and routes each virtual declaration to its owner in `ad9e3c35e`. The bridge fails loudly instead of emitting an empty filtered module: bare specifiers get `unresolved_specifier`, unknown entry files get `unknown_file` in `71df28c06`, and an unresolvable import surfaces as an error diagnostic on the import span (never a silent empty declaration). Ownership semantics documented in `a0ae7b0b8`. Test gap closed at both editor levels: in-process language-service e2e for the node_modules dependency and the pnpm workspace sibling (`ad9e3c35e`), and real tsserver protocol coverage driving definition/diagnostics across both (`fa463cd7f`). |
| **R2-MAJOR-2** M2 fix was a tsserver-wide performance regression | Rediscovery is no longer a per-request recursive walk: per-file version counters catch edits and existence checks catch deletions for known sources, while the full tree walk runs only when a directory mtime stamp moved (`BamlFileDiscovery`, `9d75e5de1` + `ad9e3c35e`). Existing add/delete e2e stays green; a new test proves repeated requests do not re-walk. |
| **R2-MINOR-1** `baml.toml` suffix match | Watch events match the project's exact config path (basename pre-filter + canonical equality against `layout().configPath`), so `mybaml.toml`/`sub/baml.toml` never reach the Rust session; watch hooks are wrapped so a rejected update reports through `watchError` and can never throw out of a bundler hook (`abc43b3f9`). |
| **R2-MINOR-2** m8 fix leaks on raced reopen | The superseded load now disposes its freshly loaded native session on the stale-ticket path (`ad9e3c35e`); raced-reopen e2e proves the session keeps serving. |
| **R2-MINOR-3** m9 disk cache unbounded | Decision: bound it. The disk artifact cache caps at 512 entries (configurable), prunes oldest-write-first every 32 puts down to 75% of the cap, single-flighted and miss-safe under races (`9d75e5de1`); eviction test + documented in the architecture doc. |
| **R2-MINOR-4** m10 offset-index performance | `buildLineIndex` is a single allocation-free pass (code-point arithmetic, no per-character `TextEncoder`) and indexes are cached per (file, snapshot) via `LineIndexCache`; diagnostics compute each span once (`9d75e5de1` + `ad9e3c35e`). |

Regression caught by the parity gate while landing R2-MAJOR-1: `Path::is_absolute` returns false for `/…` under the wasm32 host, so the absolute-specifier check is a string test — folded into `71df28c06` with a regression test, native/WASM byte parity green.

## New coverage: dependency & monorepo imports

Verified **broken** before this change (imports from a dependency silently emitted an *empty* module), now fixed and proven:

- **(a) Dependency in node_modules**: a real packed tarball shipping its own `baml_src` + `baml.toml`, installed via `npm install`, imported by a consumer — the dependency's project is discovered by walking up to its own `baml.toml` and compiled in its own session.
- **(b) pnpm-style monorepo**: an app package importing a sibling shared package's `.baml` through a symlinked `node_modules` entry — realpath canonicalization + walk-up ownership routes it to the sibling's project.

Both bundle two live projects in one build (two runtime modules, per-project virtual ids and cache namespaces). Bare specifiers go through Node resolution; concurrent `resolveId` calls open a dependency exactly once (single-flighted); `.baml` files created after open are registered eagerly.

## Round-4 review findings resolved (Sol seat Not-LGTM → fixed)

Head: `2a4643805`. Every fix is folded into its owning commit — no fixup commits; the docs-first shape, atomic implementation+tests per area commit, and the zero-`.smithers/` final tree are unchanged.

| Finding | Resolution |
|---|---|
| **R4-MAJOR** host-only new `.baml` files undiscoverable | The reviewed head already had `BamlFileDiscovery.poll()` host-authoritative (the host listing is read and compared *before* the disk-stamp check, so a stamp can never skip it), but the failure still reproduced one layer up: the tsserver plugin fed that listing from `LanguageServiceHost.readDirectory`, which is backed by the process filesystem and cannot see a buffer the editor never wrote. The plugin's host view now unions the disk listing with the editor's own open-file registry (`ProjectService`), and reads existence, text, **and** script versions from that same union in `2804bc231` — without the existence half, the per-request deletion sweep would have deleted the buffer discovery had just added, and without the version half, edits to a buffer the TypeScript project does not own would never reach the compiler. Discovery changes now also bump the project version, so `baml:client` declarations, diagnostics, and completions actually refresh instead of being served from the already-built program. Absent or renamed tsserver internals degrade to the previous disk-only listing, never to a thrown request. Proven at both editor levels with **no disk write and no import**: in-process language-service e2e (layout, `externalFiles`, `baml:client` member completions, specifier completions, buffer close withdrawing it again, and stability across repeated requests) in `2804bc231`, and real-tsserver protocol coverage in `a6ec890b7`. The round-3 performance fix is intact — the no-re-walk test stays green. Invariant restated in `d985b2de2`. |
| **R4-MINOR-1** cache prune threshold instance-local | The bound is now derived from on-disk state: every process reconciles against the directory once, on its first write, so a short-lived CLI/build that never reaches the in-process threshold still prunes a directory earlier runs grew to the cap (`8d26563b8`). Counting is readdir-only — the stat-per-entry that oldest-write-first eviction needs is paid only when the directory is genuinely over cap, so a process start costs no stats. The bound is amortized, and the doc now says so precisely: at most `PRUNE_EVERY_PUTS` entries of overshoot between prunes, and the overshoot does not grow with the number of writes, processes, or the age of the directory. New test simulates 40 short-lived processes × 2 writes and asserts the on-disk bound holds (it fails with 12 > 10 without the fix). |
| **R4-MINOR-2** `dispose()` disposed nothing | Real disposal, end to end, keeping native/WASM parity. `close` is now a protocol request (`ProjectRequest close = 14` → `ProjectClosed { released }`) in `eb0a3a44b`, generated bindings regenerated in the same commit, so the checked-in `generated/tooling.ts` never drifts from its source `.proto`; the Rust dispatcher drops the session *and* its id→root entry in `47fd05532`, so the compiler database is released rather than merely unreachable; `BamlProject.dispose()` sends it and clears the host-side source/module mirrors in `8d26563b8`. Idempotent and non-throwing on teardown paths; a released project id is dead (`unknown_project`) instead of silently reopening. Tests: Rust asserts the session leaves the workspace, a closed id errors, closing twice is a no-op, and the root reopens cleanly (`47fd05532`); TS asserts the close request is dispatched, is not re-sent, and survives a dead bridge (`8d26563b8`); the backend-parity corpus dispatches `close` to native and WASM and compares response bytes, so neither host can drift into leaking what the other frees (`a6ec890b7`). Documented in the architecture doc and the package README (`d985b2de2`).

### Gates (round 4)

`cargo fmt --all --check` clean · `cargo clippy -p baml_tooling_api -p bridge_ctypes -p bridge_wasm --all-targets` clean · `cargo test -p baml_tooling_api` 26 passed · `pnpm build:wasm` (required — the parity test compares native against WASM bytes) · `pnpm typecheck` 12/13 packages green · `pnpm test` for `@boundaryml/baml-tooling` + `@b/pkg-ts-tooling-e2e` green · `pnpm build` green. Pre-existing and unrelated to this PR: `app-website` lint (no ESLint config is tracked, so `next lint` prompts and fails non-interactively) and `app-website` typecheck (zod/react-hook-form version drift) — this PR touches zero `app-website` files.

## Old → new commit mapping

`cde9f3bb1→fdece85de` scaffolding · `0d1a92e28→43943ab66` docs (moved to front, claims corrected, ownership documented) · `eefd05a22→0dae56b34` spec · *new* `05da43656` Rust stubs · *new* `bab7bfc52` TS stubs · `379237d40→1647b8159` SDK emission (+identity fix) · `3655bd506→d8f3af3da` rename (+reserved/scoped fixes) · `4478f17c5→77cafde14` bridge sessions (+B1-Rust, M3, m1/m3/m7, config embed) · `1f1aff078→64d91283e` TS project/sidecar · `cc3fbdf78→765dbc5cc` unplugin (+B1-TS, m5, multi-project routing) · `2091352cd→f9602af3b` language-service (+M1, M2, m2/m6/m8) · `1805300cf→267b3032a` acceptance (split atomically + cross-cutting remainder) · `8245941e5→196470986` CI (+wasm/napi/bun/publish fixes) · `eff838806→431c54e5b` cleanup

Final tree diverges from the original head only by the deliberate fixes above and the new coverage — enumerated per commit in the review deck.

## Gates run

`cargo test` for all touched crates (baml_tooling_api 22, sdkgen_typescript_shared 50, baml_lsp2_actions 136, baml_lsp2_actions_tests 470) · `cargo fmt`/`clippy -D warnings` clean · `tsc --noEmit` for both TS packages · biome clean · full `pkg-ts-tooling-e2e` suite green (24 tests): builds (Vite/Rollup/Webpack/esbuild/Bun), dependency-import (packed tarball + pnpm workspace), language-service (incl. M1/M2 scenarios, editor-side dependency/sibling imports, no-rewalk, raced reopens), native-compiler, backend-parity (+M3 isolation, absolute-specifier wasm regression), packed-install, tsserver-protocol (incl. real-tsserver dependency and workspace-sibling imports).

Review deck: https://baml-unplugin-pr.smithers.sh

## Final Sol review findings resolved (one major + two minors)

Head: `6b4e7b64f`. Every change is folded into its owning commit; the series remains 14 commits, `fdece85de` is unchanged, and the final tree has zero `.smithers/` paths.

| Finding | Resolution |
|---|---|
| **R5-MAJOR** deleted `.baml` files survive as phantom compiler sources under real tsserver | `hasOpenBuffer` now reads `ScriptInfo.isScriptOpen()` instead of treating retained `ScriptInfo` existence as an open editor buffer (`7e6076b9a`). The language-service harness faithfully retains a closed ScriptInfo, records the actual bridge update, and asserts `updateFile(path, null)`, removal from specifier/member completions, and refreshed diagnostics; the genuinely open unsaved-buffer case remains green. |
| **R5-MINOR-1** 32-bit virtual-file hash can collide | Virtual identity now uses the full SHA-256 digest of root plus resolved physical source (`7e6076b9a`). A concrete regression uses two same-basename paths that collide under the former FNV-style key and proves distinct virtual declarations. |
| **R5-MINOR-2** non-BAML watch regression stopped below the adapter boundary | The regression now drives the actual Vite `handleHotUpdate` and Rollup `watchChange` hooks and asserts no compiler update, BAML invalidation, or full reload (`fc691164c`). |

### Gates (final head)

`@boundaryml/baml-tooling` build + unit tests: 37 passed · tooling and e2e TypeScript typechecks green · Biome clean on all four changed files · language-service e2e: 9 passed, including retained-ScriptInfo deletion and unsaved-buffer coverage. Rust, WASM, and packed/tarball gates were intentionally not rerun because this round is TypeScript-plugin/build-adapter local.


sxlijin:
Hi @roninjin10 - love this! Is it possible to do this through just `baml-bridge` itself? I'm poking around to understand how this works right now, did not even realize this was possible, so just working through what this changes for users.

The easiest way to talk about this would be on the https://boundaryml.com/discord, but barring that we can talk here!

aaronvg:
this is awesome, love the `import { b, Resume } from './baml_src/resume.baml` if you have a chance we definitely want to see if you have 20min to demo to us!

roninjin10:
Hey ya'll sorry about dragging my feet on this. I will join discord. I'm doing a release for smithers.sh soon so that has eaten up all my time I plan on circling back to this hopefully next week

Happy to schedule some time next week to demo this and explain how this works. I generated this code with smithers script that you can see in the history and a lot of this is taken from my previous work [evmts/tevm](https://github.com/evmts/tevm) which supports importing solidity into typescript. The patterns are pretty similar to how other things work such as 

- Effect.ts has a special typescript fork that adds effect specific linting
- Frameworks like astro vue and svelte do this to support their custom file formats
- CSS modules requires similar patterns

There are some negative tradeoffs that are worth talking about but in general it's pretty magical to have the build system and typechecker replace codegen in terms of devx

PR #4326, opened 2026-08-05
feat(tir): enrich PackageInterface export schema for source-less deps (BEP-066 slice 6a, PR 1)

antoniosarosi:
**BEP-066 slice-6a stack, PR 1 of 5** (track B: interface-centric dependency consumption). Derivation only — **zero behavior change by construction**; consumers land in the follow-up PRs of this stack.

## What
Grows the B-694 `PackageInterface` export schema so a *user package* can eventually be consumed by the compiler as a serialized blob with no source present:

- **`ExportedType::Interface`** — generic params + symbolic `Self`, per-param bounds, **pre-flattened transitive `requires` closure**, associated types (bound + default pre-lowered with symbolic `Self`), fields with attrs, required+default methods.
- **`ExportedFunction`**: `generic_param_bounds` (parallel to `generic_params`), `interface_target`, stable `callable_fqn`.
- **`ExportedType::Class`**: field attrs, per-method `interface_target`, per-param bounds.
- **`PackageInterface.namespaces`**: the full namespace-path set (interface-only namespaces included).

## How the high-risk parts are derived (no parallel lowering)
- `requires` closure: a thin wrapper (`interface_requires_closure_symbolic`) over the checker's own `interface_closure_locs_with_args_and_assoc` walk — identity args, `fill_associated_defaults=false` per the rigid-`Self` rule; consumers fill defaults from the exported rows post-substitution.
- Assoc defaults: `interface_associated_type_default` output stored verbatim.
- Default-method signatures: a twin query sharing a verbatim-extracted spine with `resolve_interface_required_methods` (whose output stays bit-identical), because `function_signature_ty` lowers default methods without a `Self` binding.

## Behavior-neutrality proof
- Every reader of `PackageInterface.types` audited; the two paths where new Interface rows could newly resolve (`resolve_type` / `resolve_type_in_own_then_deps`) explicitly skip them pending the dualization PR.
- Full `cargo test -p baml_tests -- --skip parser_stress`: green, **no snapshot churn**. Also green: tir (324), project (82), cli bytecode_cache incl. B-694 determinism (43), LSP (468).
- New suite (8 tests): fixture assertions incl. `Swapped<A,B> requires Pair<B,A>` substitution, borsh round-trip + two-DB byte determinism (user + stdlib), stdlib derivation with the `baml.iter.Iterator` requires/assoc spot-check.

## Notes
- Discovered gap, deliberately not fixed here: interface-field `@alias`/`@description` aren't hoisted into `FieldData.attributes` by AST lowering (class-side hoist doesn't run for interfaces); the export reads both homes. Parser fix + snapshot check deferred.
- Design of record: `thoughts/antonio/bep066-architecture-and-slices.md` §2.8a; research: `bep066-salsa-research.md` R1/R2.

## Summary by CodeRabbit

* **New Features**
  * Added support for compiling and running source-less mounted packages.
  * Mounted packages now support classes, interfaces, methods, functions, generics, associated types, aliases, operators, callbacks, streaming, and interface dispatch.
  * Added dependency linking and package metadata for interfaces, implementations, namespaces, schemas, and callable names.
  * Added diagnostics for unsupported mounted-package calls.

* **Bug Fixes**
  * Improved interface resolution, inheritance, default methods, coherence checks, and deterministic serialization.

* **Tests**
  * Added comprehensive mounted-package integration and parity coverage.

### DIFF
--- a/baml_language/crates/baml_cli/src/bytecode_cache.rs
+++ b/baml_language/crates/baml_cli/src/bytecode_cache.rs
@@ -3749,6 +3749,7 @@ mod tests {
                 transitive: Default::default(),
             },
             namespaces: Default::default(),
+            impls: Default::default(),
         };
         let mut seed = std::collections::BTreeMap::new();
         seed.insert(
--- a/baml_language/crates/baml_cli/src/bytecode_cache.rs
+++ b/baml_language/crates/baml_cli/src/bytecode_cache.rs
@@ -3748,6 +3748,7 @@ mod tests {
                 direct: Default::default(),
                 transitive: Default::default(),
             },
+            namespaces: Default::default(),
         };
         let mut seed = std::collections::BTreeMap::new();
         seed.insert(
--- a/baml_language/crates/baml_compiler2_emit/src/lib.rs
+++ b/baml_language/crates/baml_compiler2_emit/src/lib.rs
@@ -497,6 +497,74 @@ fn build_packages(
             }
         }
     }
+    // Mounted interfaces contribute the same declaration-side runtime facts as
+    // source interfaces. Their compiled objects/functions already live in the
+    // linked base image; these loc-free rows let a USER impl inherit a mounted
+    // default method, bake an associated default, and construct field links.
+    // Without this prepass a source `implements app.Named {}` got
+    // `label=app.Named.label`, while the blob path emitted an empty method
+    // table — a check-clean program whose virtual dispatch differed at runtime.
+    for package in baml_compiler2_hir::package::mounted_package_names(db) {
+        let Some(package_interface) =
+            baml_compiler2_tir::package_interface::mounted_interface(db, &package)
+        else {
+            continue;
+        };
+        let mut interfaces: Vec<_> = package_interface
+            .types
+            .values()
+            .flat_map(|namespace| namespace.values())
+            .filter_map(|exported| match exported {
+                baml_compiler2_tir::package_interface::ExportedType::Interface {
+                    qtn,
+                    self_param,
+                    generic_params,
+                    associated_types,
+                    fields,
+                    default_methods,
+                    ..
+                } => Some((
+                    qtn,
+                    self_param,
+                    generic_params,
+                    associated_types,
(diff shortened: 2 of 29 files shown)


antoniosarosi:
Review round addressed in 0e461fdd4:

- **Symbolic `Self` (major)** — half confirmed. The *requirement generic-args* half was real, and worse than reported: the shared closure walker lowered `requires Parent<Self>` / `requires Parent<Self.Item>` args with `self_ty: None`, producing `Ty::Error` **silently** (the walker discards its local diags) — a pre-existing checker bug, not export-only. Fixed in the walker itself (checker and export share the one derivation, so parity holds by construction): args lowering now mirrors the adjacent assoc-binding lowering (`Self` param + `self_bound` + symbolic `self_ty`). Realized walks collapse pinned members to witnesses; the export's identity root keeps symbolic projections — both asserted in a new test. The two existing corpus tests using this shape (E0132 coherence canaries) still pass; zero snapshot churn.
- The *param/method-bounds* half is not an issue: interface-level bounds lower through `env_concrete_lowering_scope`, which carries a real symbolic `self_ty` + `Self → constraint` bound; method-level bounds resolve `Self` through `method_generics` (whose first entry is the `Self` param) via the deliberate fall-through at `lower_type_expr.rs:761`. Probe output showed `Parent<TypeVar(Self#0)>` exported correctly before any change — now pinned by test so it can't regress.
- **Guard-coverage nitpick** — added `dep_interface_rows_stay_invisible_to_resolution`: `baml.iter.Iterator` is exported as an Interface row yet resolves to `None` on both guard paths, while `baml.iter.Range` (class, same namespace) still resolves — proving the guards are interface-specific.

antoniosarosi:
Closing per plan change: stacked PRs aren't available in this repo, so the whole of BEP-066 ships as a single PR — #4325 is now the long-lived feature branch. This PR's content (enriched export schema, plus the rest of the slice-6a track: #4327 #4329 #4332 #4333) is fully contained in #4325 via merge commit b84086b6a. Review happens there.

PR #4332, opened 2026-08-06
feat(compiler): execute calls into mounted packages (BEP-066 slice 6a, PR 4)

antoniosarosi:
**BEP-066 slice-6a stack, PR 4 of 5** — chained on #4329. Mounted-blob packages become *callable*: TIR's loc-free `MemberResolution::External` flows through MIR to B-693 unit symbols, and a two-DB compile-then-link harness proves it end to end **at runtime**.

## What
- Loc-free call resolution for mounted packages: free functions, methods on mounted classes, interface dispatch through blob impls (incl. out-of-body and blanket impl methods), UFCS. PR 3's reserved E0158 shrinks to the genuine builtin-only residue (enumerated + tested).
- MIR lowers `External` callees to fq-name unit symbols; impl-method `{iface}$for${target}` symbols reconstructed structurally with source/blob agreement.
- Callback-effect framing (synthetic effect params) verified against exported signatures.
- `$stream` policy landed + tested.
- **The two-DB link harness** — compile library → capture interface blob → fresh consumer DB mounting it → emit units from both → link → execute — is the working prototype of slice 6's dynamic linking. 16/16 execution tests: runtime call results asserted, mounted-class construction + field reads, blob-impl dispatch, generic bounds via blob impls, cross-package throws caught in user code.

## Gates
Focused suite 16/16 · tir + project + bytecode_cache green · full baml_tests green · strict all-features clippy clean on every touched crate · fmt clean · **zero snapshot churn**. (`--no-verify` only for the known environmental mise/swift-protobuf hook failure; gates run directly.)

Implemented by a Codex (gpt-5.6-sol) worker under stack-manager review.

## Summary by CodeRabbit

* **New Features**
  * Added support for calling functions and methods from mounted, precompiled packages.
  * Enabled cross-package interfaces, generics, implementations, streaming types, callbacks, and error handling.
  * Added interface dispatch, class construction, UFCS calls, and higher-order calls across package boundaries.
  * Improved editor symbol classification and diagnostics for external package symbols.

* **Bug Fixes**
  * Restored mounted implementations and recursive type aliases when rebuilding from a precompiled standard library.
  * Improved diagnostics for ambiguous interface field access.

* **Tests**
  * Added comprehensive end-to-end coverage for mounted package compilation, linking, and runtime execution.

PR #4333, opened 2026-08-06
feat(compiler): guarantee mounted package parity (BEP-066 slice 6a, PR 5 — capstone)

antoniosarosi:
**BEP-066 slice-6a stack, PR 5 of 5 — the capstone.** Chained on #4332. With this green, the slice-6a stack is complete: a package consumed as an interface blob is provably indistinguishable from the same package consumed as source.

## The parity oracle (`mounted_package_parity.rs`)
- Source-vs-blob **runtime parity** across the full surface: interfaces + requires + assoc defaults + default methods, attrs, generics/bounds, enums, aliases, in-body/out-of-body/blanket impls, user impls of mounted interfaces, caught-throw behavior.
- **Byte-identical diagnostic fingerprints** (code/message/severity/phase/user range) for the misuse variants: wrong arity, unsatisfied bound, non-exhaustive match, missing impl members.
- **Byte-identical artifacts**: dependency units, package metadata, consumer import tables raw-identical; consumer units + whole programs identical after normalizing database-local debug FileIds — the sole legitimate difference (the known B-693 §9b wart), with paths/ranges/symbols/executable bytes identical.

## What the oracle caught (fixed in the implementation, never weakened in the oracle)
1. Source auxiliary packages didn't participate in user dependency resolution the way mounted ones do.
2. Mounted interface defaults/assoc-defaults/fields/method tables seeded emission differently from source interfaces.
3. Impl-method exports mishandled synthetic callback-effect parameters vs enclosing impl generics.

## Also in this PR
- **Coherence-fails-open hole closed and proven**: user-vs-blob impl overlap yields exactly one primary-only E0132 attributed at the user's impl, dep side rendered structurally.
- **Public link seam**: the two-DB compile-and-link harness promoted from test scaffolding to a documented API — slice 6's `Package.compile` consumes this.
- **Residue ledger** documented adjacent to the public API (E0158 builtin residue, span-dependent LSP ops, re-aliasing/transitive re-export, live `$stream` loading, artifact versioning, dispatcher synthesis) — slice 6 starts from a ledger, not archaeology.

## Gates
Full baml_tests green · all-features mounted call/parity suites 26/26 · HIR/TIR/project · bytecode_cache 43/43 · fmt/clippy/rustdoc all-features `-D warnings` · **zero snapshot churn**.

Implemented by a Codex (gpt-5.6-sol) worker under stack-manager review.

## Summary by CodeRabbit

- **New Features**
  - Added support for compiling applications against mounted, source-less packages, including their default methods, type defaults, and field definitions.
  - Added dependency linking for precompiled package artifacts before consumer compilation.

- **Bug Fixes**
  - Improved handling of generic callback-effect metadata in exported functions and implementation methods.
  - Improved package discovery and dependency resolution for auxiliary and mounted source packages.

- **Tests**
  - Added comprehensive parity coverage for source-based and serialized mounted packages, including runtime behavior, diagnostics, imports, and linking.

### DIFF
--- a/baml_language/crates/baml_compiler2_emit/src/lib.rs
+++ b/baml_language/crates/baml_compiler2_emit/src/lib.rs
@@ -497,6 +497,74 @@ fn build_packages(
             }
         }
     }
+    // Mounted interfaces contribute the same declaration-side runtime facts as
+    // source interfaces. Their compiled objects/functions already live in the
+    // linked base image; these loc-free rows let a USER impl inherit a mounted
+    // default method, bake an associated default, and construct field links.
+    // Without this prepass a source `implements app.Named {}` got
+    // `label=app.Named.label`, while the blob path emitted an empty method
+    // table — a check-clean program whose virtual dispatch differed at runtime.
+    for package in baml_compiler2_hir::package::mounted_package_names(db) {
+        let Some(package_interface) =
+            baml_compiler2_tir::package_interface::mounted_interface(db, &package)
+        else {
+            continue;
+        };
+        let mut interfaces: Vec<_> = package_interface
+            .types
+            .values()
+            .flat_map(|namespace| namespace.values())
+            .filter_map(|exported| match exported {
+                baml_compiler2_tir::package_interface::ExportedType::Interface {
+                    qtn,
+                    self_param,
+                    generic_params,
+                    associated_types,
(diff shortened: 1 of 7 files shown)


antoniosarosi:
@coderabbitai review

PR #4342, opened 2026-08-07
fix(fmt): keep dotted member chains atomic, break only at call boundaries

aaronvg:
## Problem

`baml fmt` exploded dotted namespace paths one segment per line whenever the full expression overflowed the line width:

```baml
let result = root
    .ai
    .Agent<Itinerary>
    .new()
    .run(
        plan_trip_spec(
            "...",
            root.anthropic.AnthropicClient.new(),
        ),
    );
```

## Fix

The chain printer (`PrintChain` in `baml_fmt`) now follows the standard prettier/rustfmt member-chain rule: plain accesses glue to their receiver; call sites are the break points.

1. **Plain-access paths are atomic.** Namespace segments, field accesses, and generic type segments (`root.ai.Agent<Itinerary>`, `StopReason.Complete`) never break internally, regardless of length.
2. **Tail-broken layout (new).** When everything up to the final call fits on the line, only the final call's arguments wrap:

```baml
let result = root.ai.Agent<Itinerary>.new().run(
    plan_trip_spec(
        "...",
        root.anthropic.AnthropicClient.new(),
    ),
);
```

3. **Break at call boundaries only.** When the chain itself is too long, it breaks before the `.name` of each call group, keeping the leading path — plus the first call when it fits — glued to the receiver line:

```baml
let result = root.ai.Agent<Itinerary>.new()
    .with_client(root.anthropic.AnthropicClient.new())
    .with_options(the_default_options)
    .run(the_spec_value);
```

## Verification

- All 102 pre-existing `baml_fmt` tests pass unchanged (including the 1024-case scenario matrix with idempotency checks) — zero churn in existing formatter behavior outside chains.
- 8 new regression tests in `member_chain_layout_tests` (glued namespace chain, exploded-input collapse, long array arg, chain that must break, atomic long path, `StopReason.Complete` in match arms, short chain single-line, trailing plain accesses); each asserts idempotency.
- `cargo clippy -p baml_fmt --all-targets` clean; `baml_tests` formatter-dependent suites (`backtick_fmt_value_preservation`, `interfaces_associated_types`) pass.
- Ran the patched `baml-cli fmt` against a copy of the `_planv2` BAML sources and diffed against a canary-baseline CLI run: all 12 changed files show only the intended chain-layout improvements (e.g. `live/anthropic.baml` collapses to exactly the desired form above).

Note: kept out of scope per the in-flight `KW_CLIENT` formatter fix — no changes to keyword-as-identifier handling.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the formatter’s layout logic and snapshots; they do not affect parsing, type-checking, or runtime behavior.
> 
> **Overview**
> **`baml fmt` no longer splits long dotted paths one segment per line.** `PrintChain` now uses a prettier/rustfmt-style rule: field accesses, namespace segments, and generic type args stay glued to their receiver; line breaks happen only at method-call/index groups (with width-based “glue” for the first group when it fits).
> 
> A new **`try_print_tail_call_broken`** layout runs before the multi-line fallback so when the prefix fits on one line, only the **final** call or index wraps its arguments (including optional `?.(...)` chains).
> 
> Multi-line printing was rewritten around **call groups** (method name + following calls/indexes), leading optional calls on the receiver, and accurate **`multi_lined`** reporting. **`GenericArgs`** are formatted as plain chain segments instead of separate breakable items.
> 
> Regression coverage adds **`member_chain_layout_tests`** and updates the **`chain_expr`** formatter snapshot to match the new output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63da78ca89f5261dba6ff70a272ca9a221c69fe8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting for chained expressions, keeping related member and namespace paths together.
  * Added more consistent line breaks around method calls, indexing, and generic arguments.
  * Preserved compact layouts for short chains while wrapping long call or index arguments cleanly.
  * Fixed formatting for enum-member paths and trailing field accesses after wrapped calls.

* **Tests**
  * Added regression coverage for chained-expression formatting and argument wrapping.

### DIFF
--- a/baml_language/crates/baml_fmt/src/ast/expressions.rs
+++ b/baml_language/crates/baml_fmt/src/ast/expressions.rs
@@ -5669,13 +5669,15 @@ impl PrintChain<'_> {
     /// Applies when the whole prefix up to the final call fits the line.
     /// Should be passed a sub-printer to avoid printing partial output in
     /// the event that the layout does not apply.
-    fn try_print_tail_call_broken(&self, shape: &Shape, printer: &mut Printer) -> Option<PrintInfo> {
+    fn try_print_tail_call_broken(
+        &self,
+        shape: &Shape,
+        printer: &mut Printer,
+    ) -> Option<PrintInfo> {
         let (last, prefix) = self.chain_members.split_last()?;
         let question_dot = match last {
             PrintChainItem::Call(_) | PrintChainItem::Index(_) => None,
-            PrintChainItem::OptionalCall(qd, _) | PrintChainItem::OptionalIndex(qd, _) => {
-                Some(*qd)
-            }
+            PrintChainItem::OptionalCall(qd, _) | PrintChainItem::OptionalIndex(qd, _) => Some(*qd),
             PrintChainItem::FieldAccess(..)
             | PrintChainItem::OptionalFieldAccess(..)
             | PrintChainItem::GenericArgs(..) => return None,
@@ -5687,6 +5689,10 @@ impl PrintChain<'_> {
         if printer.output.len() > shape.width {
             return None;
         }
+        // `shape.width` is the remaining line budget measured from the
+        // chain's start column (`width + indent + first_line_offset ==
+        // line_width`), and this sub-printer's output also starts at that
+        // column, so the args' budget is what the prefix left over.
         let args_shape = Shape {
             width: shape.width.saturating_sub(printer.output.len()),
             indent: shape.indent,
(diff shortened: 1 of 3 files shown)


aaronvg:
@coderabbitai review

aaronvg:
@coderabbitai resolve

PR #4415, opened 2026-08-13
test(fmt): cover baml.spawn paths in type positions

sxlijin:
## Summary

- add B-1509 regression coverage for `baml.spawn.CancelToken` in parameter, return, class-field, and generic-argument type positions
- document that formatter contextual identifiers cover both expression and type path segments
- keep the existing B-258 expression-position behavior covered alongside the new cases

## Root cause

`spawn` is lexed as `KW_SPAWN` even when it appears as a namespace segment. The formatter previously expected a plain `WORD` in type paths. The shared contextual-keyword handling now accepts this valid parser output, but type-position coverage was missing.

## Impact

Files with signatures such as `tok: baml.spawn.CancelToken` remain formatable and formatter behavior stays aligned with the parser across all reported type positions.

## Validation

- `cargo test -p baml_fmt --lib` (125 passed)
- `cargo fmt --all -- --check`
- `git diff --check`

Linear: B-1509

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting support for qualified type paths containing keyword-like segments, such as `baml.spawn.CancelToken`.
  * Ensured these paths remain intact across parameters, return types, class fields, and generic types.

* **Tests**
  * Added regression coverage verifying formatting succeeds and produces consistent, repeatable output.

### DIFF
--- a/baml_language/crates/baml_fmt/src/ast/tokens.rs
+++ b/baml_language/crates/baml_fmt/src/ast/tokens.rs
@@ -597,11 +597,11 @@ impl Word {
 /// True for token kinds the parser accepts as identifiers in name positions.
 ///
 /// The lexer emits dedicated keyword kinds for these words, but the parser
-/// keeps them valid as field, parameter, method, and member-access names
-/// (e.g. a class field or parameter named `client`, or `x.implements(y)` on
-/// the reflection `type` value). The CST therefore contains the keyword kind
-/// where the strong AST expects a name, and [`Word::from_cst`] must accept it.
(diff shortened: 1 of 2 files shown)


sxlijin:
Closing as redundant: the reported formatter behavior was already fixed by #4341 (`fix(fmt): accept contextual keywords as identifiers in name positions`), merged as `061b224ea4`. B-1509 now links that original fix; this PR only added explicit regression coverage for behavior already present on `canary`.

PR #4446, opened 2026-08-15
fix(csharp): align release verifier with generated surface

sxlijin:
## Summary

- Fixes the deterministic C# verification failure in the baml_language 0.17.0 production release run: https://github.com/BoundaryML/baml/actions/runs/31857280879
- Aligns the phase-12 package consumer with the generated C# surface established by #4366, which intentionally omits methods inherited through BAML interface implementations.
- Keeps runtime coverage for the C# CSV methods and resource fields that are generated.

## Root cause

The phase-12 fixture is normally ignored because of B-1059, but the production full-surface package verifier compiles and runs it. Its stale calls to CsvReader.Iter/Next and CsvRows.Iter/Next no longer compile after #4366 removed interface-implementation methods from generated class method collections.

## Validation

- cargo nextest run -p sdk_test_csharp --all-features --run-ignored only -E test(test_phase12_executes_native_typed_resource_apis_lifetimes_and_state)
- dotnet build sdk_tests/crates/csharp/phase12_resources/Phase12Resources.csproj --configuration Release --no-restore
- git diff --check

## Release follow-up

After this merges and post-merge canary CI succeeds, 0.17.0 must be dispatched from the new immutable baml-language source tag. This PR does not publish or dispatch the release.

## Summary by CodeRabbit

* **Tests**
  * Updated C# CSV resource validation to verify headers, reader availability, and the initial reading position.
  * Simplified resource cleanup validation to ensure the active CSV reader is properly closed.
  * Removed checks for iterator identity, typed row decoding, and raw record access.

### DIFF
--- a/baml_language/sdk_tests/crates/csharp/phase12_resources/Program.cs
+++ b/baml_language/sdk_tests/crates/csharp/phase12_resources/Program.cs
@@ -400,55 +400,21 @@ finally
 
 using Baml.Csv.CsvReader csvReader = Baml.Csv.Functions.Reader(
     "name,count\nalpha,7\n");
-Require(ReferenceEquals(csvReader, csvReader.Iter()), "CsvReader.Iter did not preserve iterator identity");
 IReadOnlyList<string>? headers = await csvReader.HeadersAsync();
 Require(
     headers is not null && headers.SequenceEqual(new[] { "name", "count" }),
     "native CsvReader.headers changed");
 
 using Baml.Csv.CsvRows<CsvRow> csvRows = csvReader.Rows<CsvRow>();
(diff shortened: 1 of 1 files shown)


sxlijin:
@coderabbitai review

sxlijin:
Attempt 2 of the 0.17.0 release run confirmed the same deterministic failure: https://github.com/BoundaryML/baml/actions/runs/31857280879/attempts/2?pr=4446

The phase-12 consumer fails to compile because generated CsvReader has no Iter/Next and generated CsvRows<T> has no Iter/Next/NextAsync. This PR removes those stale interface-implementation calls and has already passed the exact phase-12 native integration test locally.

PR #4450, opened 2026-08-15
perf: eliminate remaining BEP-066 hot-path overhead

antoniosarosi:
## Summary

- Removes the remaining BEP-066 compiler hot-path cost by memoizing ground impl resolution through Salsa, keeping mounted candidates lazy, and restoring the ordinary no-runtime-slot inference path.
- Removes the remaining VM hot-path cost by omitting exact runtime-type metadata from ordinary frames, bypassing empty type-argument work, and compacting immutable static virtual-dispatch cache keys.
- Preserves dynamic builder/package behavior, mint identity, runtime generic checks, and the live resolver for runtime-owned rules.

## Root cause

Compiler phase timing isolated the residual to MIR lowering and body inference. Repeated operator/interface lookups rebuilt an eager package impl-candidate vector and rescanned every source and mounted impl for the same ground primitive/container types. The fix makes candidate production lazy and memoizes the dependency-tracked resolved impl set by interned ground type.

Runtime bytecode for the class-construction residual was instruction-identical to `d8ad5e58c`, confirming VM overhead. Ordinary non-generic allocations and calls still created/moved empty rich type-argument containers; every frame carried dynamic-definition and exact-value lanes; static virtual-dispatch hits still reconstructed owned method/interface/receiver keys. The fix keeps those data structures on the BEP-066 path and leaves ordinary calls/allocations lean.

## Benchmarks

Five samples, `BAML_PROFILE=0`, same quiet machine and paired measurement window. Runtime used `DIVAN_MAX_TIME=20` so heavy adaptive workloads also received five samples. Baseline is pre-BEP-066 canary `d8ad5e58c`; PR is `5024874a4`.

### Compiler

| Benchmark | d8ad5e58c median | PR median | Delta |
|---|---:|---:|---:|
| compile_baml_tests_project | 2.098 s | 1.537 s | -26.7% |
| compile_empty_project | 319.8 ms | 259.1 ms | -19.0% |

The empty-project improvement is consistent with removing fixed BEP-066 query/metadata overhead; no stdlib-content exception is needed.

### Runtime

| Workload | d8ad5e58c median (ms) | PR median (ms) | Delta |
|---|---:|---:|---:|
| classes/method_call_100k | 32.21 | 30.89 | -4.1% |
| classes/method_chain_100k | 34.93 | 34.96 | +0.1% |
| compute/array_build_sum_100k | 187.9 | 101.8 | -45.8% |
| compute/binary_tree_depth_20 | 338.2 | 298.1 | -11.9% |
| compute/bubble_sort_5k | 1710 | 1762 | +3.0% |
| compute/call_chain_100x10k | 117.4 | 101.7 | -13.4% |
| compute/class_instances_100k | 14.36 | 10.74 | -25.2% |
| compute/closure_apply_1m | 506.6 | 516.4 | +1.9% |
| compute/collatz_100k | 729.2 | 673.0 | -7.7% |
| compute/divide_guard_1m | 188.2 | 174.2 | -7.4% |
| compute/fib32_recursive | 1122 | 975.9 | -13.0% |
| compute/fib_iterative_100k | 256.3 | 246.3 | -3.9% |
| compute/generic_apply_with_inferred_type_args_1m | 653.1 | 573.0 | -12.3% |
| compute/grid_alloc_1000x100 | 153.3 | 107.6 | -29.8% |
| compute/guard_clauses_1m | 223.9 | 203.8 | -9.0% |
| compute/if_else_dispatch_1m | 225.3 | 202.0 | -10.3% |
| compute/mixed_ops_5k | 3.497 | 2.445 | -30.1% |
| compute/nested_field_access_500k | 145.1 | 135.3 | -6.8% |
| compute/nested_loops_500x500 | 11.13 | 10.55 | -5.2% |
| compute/pure_call_1m | 153.2 | 136.5 | -10.9% |
| compute/wide_nested_class_create_50k | 60.43 | 56.86 | -5.9% |
| concurrency/parallel_sum_4x250k | 18.64 | 14.83 | -20.4% |
| concurrency/parallel_sum_sequential_1m | 38.22 | 36.13 | -5.5% |
| concurrency/shared_array_1w_4r | 25.18 | 22.04 | -12.5% |
| concurrency/shared_array_push_4x25k | 19.42 | 17.48 | -10.0% |
| concurrency/spawn_await_x10k | 110.5 | 108.1 | -2.2% |
| concurrency/spawn_fan_out_100 | 0.7856 | 0.7856 | 0.0% |
| interfaces/default_method_100k | 277.5 | 173.7 | -37.4% |
| interfaces/match_dispatch_100k | 99.33 | 60.63 | -39.0% |
| interfaces/polymorphic_dispatch_100k | 101.4 | 82.18 | -19.0% |
| string/concat_loop_10k | 1.754 | 1.100 | -37.3% |
| string/contains_medium_literal_100k | 17.36 | 15.44 | -11.1% |
| string/split_long_literal_1k | 2.667 | 2.214 | -17.0% |
| string/split_medium_literal_10k | 13.58 | 13.24 | -2.5% |
| string/split_short_literal_100k | 84.11 | 80.05 | -4.8% |
| string/string_split_100k | 83.48 | 46.80 | -43.9% |
| string/substring_slice_long_100k | 40.70 | 34.85 | -14.4% |
| string/trim_padded_literal_100k | 34.76 | 28.12 | -19.1% |

Every measured regression is within +5%; workloads outside the negative side of the literal band are improvements and were not deliberately slowed.

## Validation

- Focused compiler/VM checks passed.
- Focused `bex_vm` + `baml_compiler2_hir_ty` nextest: 226/226 passed twice, including after the final Salsa cache.
- Pinned gate: 3,724/3,724 passed, 24 skipped; doctests passed; no changed or unreferenced snapshots.
- Pinned gate package set includes `baml_tests`, `baml_cli`, `baml_lsp2_actions`, `baml_lsp2_actions_tests`, and `baml_surface` with all features.
- `cargo fmt --all -- --check`, `git diff --check`, and affected-package Clippy with all targets/features and warnings denied passed.
- Full workspace Clippy with all targets/features and warnings denied passed in pre-commit hooks.
- Temporary phase timing and bytecode-dump instrumentation was removed before commit; the detached `d8ad5e58c` baseline worktree is clean.

## Summary by CodeRabbit

* **Performance**
  * Improved generic type handling and virtual-call caching, especially when runtime metadata is unnecessary.
  * Reduced unnecessary type resolution, validation, and allocation during compilation and execution.
  * Accelerated implementation candidate lookup and calls without runtime type arguments.

* **Reliability**
  * Strengthened runtime validation of generic arguments and bounds when enabled.
  * Improved consistency of type information across virtual methods and instance creation.

* **Bug Fixes**
  * Fixed virtual dispatch caching for distinct generic interface instantiations.

## Review follow-up

- Routed VM frame marking through `Frame::collect_roots`, so boxed exact-type metadata and its owner/definition pointers survive and are forwarded by GC; removed the duplicated class-definition traversal.
- Preserved the sparse exact runtime `TypeValue` lane for virtual generic methods after resolver-provided owner/impl slots.
- Added focused unit regressions for metadata GC survival/forwarding and runtime-mint slot alignment.
- Review-amended validation: `bex_vm` 71/71; dynamic reflection/builder/identity/binding cluster 35/35 (1 skipped); repeated pinned gate 3,724/3,724, 24 skipped; doctests and snapshot-reference rejection clean; full workspace Clippy hook green.

## Adversarial review follow-up

- Fixed F1 in `88653e17c`: static virtual-dispatch keys now retain realized positional interface arguments, so full-context mint equivalence cannot alias instantiations that fact-poor `StructuralEquivCtx` selection distinguishes.
- Added a heterogeneous single-call-site regression using `Tagged<Animal>` and `Tagged<Animal | Dog>`; it reproduced the old wrong result (`animal|animal`) before the fix and now returns `animal|animal-or-dog`.
- Fixed F2 with Salsa `cycle_result` recovery returning the inductive empty impl set. Also folded in F4/F5/F6 dependency/invariant documentation and the exact compile-time heap-pointer receiver predicate.
- Focused validation: compiler+VM 228/228; full interface suite 490/490 (2 skipped); strict focused Clippy green.
- Quiet paired non-generic dispatch spot checks versus `d8ad5e58c`: default 153.1 vs 273.8 ms, match 53.81 vs 102.9 ms, polymorphic 69.17 vs 104.4 ms; no richer-key performance loss.
- Repeated widened pinned gate: 3,725/3,725, 24 skipped; doctests and snapshot-reference rejection clean; full workspace Clippy hook green.

### DIFF
--- a/baml_language/crates/baml_compiler2_hir_ty/src/impls.rs
+++ b/baml_language/crates/baml_compiler2_hir_ty/src/impls.rs
@@ -326,7 +326,7 @@ pub(crate) fn is_concrete_receiver(ty: &Ty) -> bool {
 
 /// Resolution-relevant impl facts in a location-free shape shared by source
 /// blocks and mounted export rows.
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct MountedImplFacts {
     pub interface: InterfaceRef,
     pub for_ty_pattern: Ty,
@@ -334,6 +334,7 @@ pub struct MountedImplFacts {
     pub associated_types: Vec<(Name, Ty)>,
 }
 
+#[derive(Clone, PartialEq)]
 pub enum ResolvedImplFacts<'db> {
     Source(&'db ImplFacts<'db>),
     Mounted(MountedImplFacts),
@@ -370,6 +371,7 @@ impl ResolvedImplFacts<'_> {
 }
 
 /// The dispatch identity retained after matching an impl.
+#[derive(Clone, PartialEq)]
 pub enum ResolvedImplOrigin<'db> {
     Source {
         block: ImplLoc<'db>,
@@ -381,10 +383,46 @@ pub enum ResolvedImplOrigin<'db> {
 }
 
 /// One resolved impl plus the generic instantiation the match pinned.
+#[derive(Clone, PartialEq)]
 pub struct ResolvedImpl<'db> {
     pub origin: ResolvedImplOrigin<'db>,
     pub facts: ResolvedImplFacts<'db>,
     pub bindings: FxHashMap<ParamTy, Ty>,
+}
+
+#[derive(Clone, PartialEq)]
+enum CachedResolvedImplOrigin<'db> {
+    Source {
+        block: ImplLoc<'db>,
+    },
+    Mounted {
+        methods: Vec<crate::package_interface::ExportedFunction>,
+        facts: MountedImplFacts,
+    },
+}
+
+#[derive(Clone, PartialEq)]
+struct CachedResolvedImpl<'db> {
+    origin: CachedResolvedImplOrigin<'db>,
+    bindings: FxHashMap<ParamTy, Ty>,
+}
+
+// SAFETY: the cached rows contain Salsa-interned locations tied to the query
+// database. This is the same PartialEq-driven overwrite contract as ImplFacts.
+#[allow(unsafe_code)]
+unsafe impl salsa::Update for CachedResolvedImpl<'_> {
+    #[allow(unsafe_code)]
+    unsafe fn maybe_update(old_pointer: *mut Self, new_value: Self) -> bool {
+        #[allow(unsafe_code)]
+        unsafe {
+            let changed = *old_pointer != new_value;
+            if changed {
+                std::ptr::drop_in_place(old_pointer);
+                std::ptr::write(old_pointer, new_value);
+            }
+            changed
+        }
+    }
 }
 
 impl ResolvedImpl<'_> {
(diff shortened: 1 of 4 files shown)


antoniosarosi:
# Adversarial review — round-2 perf commits

Deep review of the optimization claims (same process as the round-1 review). **Verdict: SHIP-WITH-FIXES — two blockers, both small and local.** Note: this analyzed head `5024874a4`; commit `563f7f013` landed mid-review and already addresses what the report calls F3 (the CodeRabbit findings), so F3 is superseded.

## Blocking

**F1 — HIGH. Dispatch-cache key collapses interface instantiations the resolver distinguishes** (`bex_vm/src/vm.rs:7052-7060`). The old key compared `interface + interface_args` structurally; the new key is `canonical_digest` minted under the full `BexVm` fact context — but impl selection compares interface args under the deliberately fact-poor `StructuralEquivCtx` (`package_baml/resolve.rs:672-677`). Canonicalization consults exactly the facts that differ (`normalize.rs:2040/2049/2356/1312`), so the key is strictly coarser than the selection predicate. Two generic-interface instantiations `I<A>`/`I<B>` at one call site can alias in the cache and silently dispatch the wrong impl. Tests miss it because every dispatch site is exercised with a homogeneous interface. Fix: include the realized `iface_args` in the key (empty for non-generic interfaces, so the perf win is untouched), or mint the key digest under `StructuralEquivCtx`.

**F2 — MEDIUM. `impls_for_type_cached` lacks salsa cycle recovery** (`baml_compiler2_hir_ty/src/impls.rs:721-731`). Promoting the recursive fn to `#[salsa::tracked]` turns re-entry (reachable via `bounds_hold → resolve_within_depth → realized_assoc_default → … → impls_for_type`) into a default-strategy salsa cycle panic. Sibling precedent exists (`interfaces/impl_rules.rs:220` uses `cycle_result`). Add `cycle_result` returning an empty set, or document why re-entry is impossible.

## Hardening follow-ups (non-blocking)

- **F4**: the lazy candidate iterator records dependency edges only for yielded blocks; sound today (the memoized consumer never `break`s), but a future `break` becomes silent staleness — add a comment or a debug assertion.
- **F5**: `CachedResolvedImplOrigin::Source` must stay fact-free (facts re-hydrate live at `impls.rs:696-707`); the `Mounted` arm follows the opposite rule. Document the invariant; fix the `SAFETY:` comment to state the real justification (no `db` borrows, only owned/Copy data).
- **F6**: the receiver-key gate uses `runtime_type.is_none()` + a media-kind string match as a proxy; `heap.is_compile_time_ptr(instance.class)` is the exact structural predicate and available.

## Verified clean (equivalence-proved, not just tested)

Zero-count `TakenTypeArgs` bypass (bit-identical to old path), marker-branch hoist (order preserved, both validators disabled identically), all three infer early-gates (case-by-case substitution equivalence through `InterfaceRef::new`), name-based gates (resolved-definition re-verification downstream; no import aliases exist), pattern hoist (guard implies the hoisted condition), the parts of the cache key that hold (`(function_addr, call_pc)` pins baked operands; type tags are emit-time-verified bijections; runtime mints carry `$dyn/<mint>` so definitionally-equal mints never alias; static-before-dynamic rule order closes the Session.eval invalidation attack), salsa dependency tracking (`MountedPackages` is a tracked input; `ImplTypeKey` is finer than the internal equivalence — dilution, never aliasing), and diff completeness (three source files, zero snapshot churn — pointedly unlike round 1).

Also worth crediting: the `VirtualCall` fast-path guard correctly includes `pending_call_type_values.is_empty()` — precisely the subtle condition a careless version of this optimization would drop.

PR #4453, opened 2026-08-16
perf(runtime): precompile stdlib prefix for Package.compile

antoniosarosi:
## Summary

`reflect.Package.compile` rebuilt the embedded ~519 KiB/90-file stdlib in every fresh transient compiler database. Diagnostics walked all builtin sources and emission compiled the full program before discarding stdlib units, making a small runtime package pay a ~4–5 second fixed cost.

This PR builds a compiler-coupled stdlib artifact in `bex_project`'s Cargo build unit and embeds:

- Borsh `PackageInterface` bytes for each stdlib package;
- the stable optimized stdlib `Program` prefix;
- an artifact key containing schema version, canonical BAML version/channel, `OptLevel`, and test-emission mode. Cargo's dependency fingerprint and `OUT_DIR` couple producer and consumer, so a compiler or configuration change rebuilds the artifact.

Each `Package.compile`/session invocation deserializes an owned artifact, creates a fresh DB without stdlib source files, mounts stdlib interfaces as a new immutable-precompiled classification, and emits only runtime units against the prefix. There is no process `Lazy`, retained compiler database, or first-call bootstrap.

## Semantics and perf-preserving design

- The ordinary `set_project_root` and `emit_units` compiler paths are unchanged.
- Precompiled stdlib impl rows use a dedicated fact-free `Precompiled` origin, not the dynamic/fact-bearing `Mounted` shape, so the #4448 impl cache remains eligible.
- Static VM callers retain the original image-function-address cache keys and interpreter branches from #4450. Runtime-grafted callers resolve Function, Closure, and BoundMethod callees to the shared heap `Object::Function`, then key caches with `function.as_ptr() | 1`; major GC forwards live tagged keys and evicts dead ones. Runtime-only key construction is out of line, and `Function` layout is unchanged.
- Prefix decomposition links runtime units to host-image stdlib objects while preserving package contents, imports, diagnostics, and created-once `Package.compile` mints.
- Session declaration types retain created-once generative mints and package provenance. Two sessions declaring an equal-shaped class remain type-identity-distinct, and host-side interface dispatch can recover the declaring session package. A live escaped type handle retains that provenance graph; dropping the handle releases it.
- Reserved stdlib names cannot be rebound through `Package.compile` dependencies, and VM type lookup treats the immutable stdlib set as authoritative before runtime aliases.
- Only Vm/Io precompiled builtins become linkable. Intrinsic/AwaitAny calls remain reserved, with the exception trusted only for immutable precompiled packages.

## Performance

Cold public path (five-sample median):

| benchmark | before | after |
|---|---:|---:|
| small `Package.compile` | ~4–5 s | **20.6 ms** |

Paired stdlib-interface dispatch:

| path | 0 iterations | 100k iterations | setup-subtracted loop |
|---|---:|---:|---:|
| static compile | 5.749 µs | 84.39 ms | 84.384 ms |
| runtime `Package.compile` | 28.84 ms | 115.3 ms | 86.46 ms |

This sample indicates a small slowdown, but setup subtraction and mixed sampling variance are at least as large as the delta, so it is directional rather than a precise +2.46% bound. A quiet-window review-fix spot-check subtracted to 82.565 ms static vs 85.42 ms runtime (about +3.5%), still within the requested ±5%/noise bar. The regressions cover shared cache identity through Closure/BoundMethod callees and dispatch before/after major GC.

Controlled exact-base/current compiler A/B on this host:

| compile bench | `origin/canary` | branch | delta |
|---|---:|---:|---:|
| empty project | 547.6 ms | 539.8 ms | -1.4% |
| full project | 2.366 s | 2.382 s | +0.7% |

This machine could not reproduce the historical ~1.54 s absolute full-project rate for either exact revision, so the paired result is the meaningful preservation check.

The full runtime speedtest suite was measured base/current. Every workload was within ±5% or resolved as noise on an immediate paired rerun. Representative deltas: class dispatch +3.7–3.9%, binary trees +2.0%, bubble sort +2.1%, interface default -2.4%, interface polymorphic -2.5%, divide loop +3.0%; noisy string concat re-paired at +1.1%.

The build script necessarily carries the compiler graph and the embedded compiler-produced artifact increases shipped native files by roughly 3.2 MB (compressed growth is materially smaller; wasm gzip grew by about 0.55 MB). Platform baselines and absolute ceilings are re-pegged to the first CI measurements with the repository's usual ~3% headroom.

## Verification

- Focused integration suites: 62/62 (`runtime_package_compile` 14, `mounted_package_calls` 16, `mounted_package_parity` 10, `runtime_session` 17, `emit_determinism` 5); the runtime cache identity unit regression and 11/11 size-gate tests also passed.
- Exact widened pinned gate: 3,742/3,742 tests passed, 24 skipped; doctests clean; no unreferenced snapshots.
- The public source-less prefix output is Borsh-byte-identical to normalized user units from a full-source compile. Link/relink byte-identity, mounted parity, cross-session mint inequality, host dispatch on session classes, major-GC, and 500-session-evaluation retention oracles passed.
- `cargo fmt --all -- --check` passed.
- Strict affected-package, all-target/all-feature Clippy with `-D warnings` passed.
- Canonical-version metadata check and `bex_project` wasm32 check passed.

The repository-wide pre-commit Clippy hook currently stops on an unrelated existing `clippy::question_mark` finding in `sys_auth/src/io_bridge.rs:105`; that file is byte-identical to `origin/canary`. Commits skipped only that hook after the scoped strict Clippy and full pinned gate passed.

## Follow-ups

None required for Wasm: the existing bridge consumes this mechanically, and the wasm32 build is green.

## Summary by CodeRabbit

* **New Features**
  * Runtime compilation now uses a precompiled standard library for faster, more consistent execution.
  * Improved support for standard-library functions, static methods, interfaces, generics, and externally provided packages.
  * Runtime type and virtual-dispatch caching improves repeated calls and type operations.
  * Fully qualified interface names can now be resolved at runtime.

* **Bug Fixes**
  * Reserved dependency aliases now produce clear compilation errors.
  * Improved session value handling, garbage-collection behavior, and runtime type identity.

* **Tests**
  * Added coverage for compilation determinism, package compilation performance, runtime dispatch, and session behavior.

### DIFF
--- a/baml_language/.cargo/size-gate.toml
+++ b/baml_language/.cargo/size-gate.toml
@@ -43,13 +43,13 @@ policy.max_delta_pct = 3.0
 # Reclaiming the `syn` cost would mean dropping prettyplease (raw,
 # unformatted generated SDK output).
 [artifacts.baml-cli.platform.aarch64-apple-darwin]
-max_file_bytes = "21.2 MiB"
+max_file_bytes = "25.0 MiB"
 
 [artifacts.baml-cli.platform.x86_64-unknown-linux-gnu]
-max_file_bytes = "27.2 MiB"
+max_file_bytes = "31.1 MiB"
 
 [artifacts.baml-cli.platform.x86_64-pc-windows-msvc]
-max_file_bytes = "23.5 MiB"
+max_file_bytes = "26.7 MiB"
 
 [artifacts.packed-program]
 kind = "pack"
@@ -64,13 +64,13 @@ policy.max_delta_pct = 3.0
 # bytecode ships in every packed program. Each ceiling is ~3% above the
 # re-recorded baseline in `.ci/size-gate`.
 [artifacts.packed-program.platform.aarch64-apple-darwin]
-max_file_bytes = "17.0 MiB"
+max_file_bytes = "20.2 MiB"
 
 [artifacts.packed-program.platform.x86_64-unknown-linux-gnu]
-max_file_bytes = "21.2 MiB"
+max_file_bytes = "24.4 MiB"
 
 [artifacts.packed-program.platform.x86_64-pc-windows-msvc]
-max_file_bytes = "18.2 MiB"
+max_file_bytes = "21.3 MiB"
 
 [artifacts.packed-program.pack]
 cli_package = "baml_cli"
@@ -93,4 +93,4 @@ wasm = true
 policy.max_delta_pct = 3.0
 
 [artifacts.bridge_wasm.platform.wasm32-unknown-unknown]
-max_gzip_bytes = "4.7 MiB"
+max_gzip_bytes = "5.3 MiB"
--- a/baml_language/.ci/size-gate/aarch64-apple-darwin.toml
+++ b/baml_language/.ci/size-gate/aarch64-apple-darwin.toml
@@ -1,15 +1,15 @@
 version = 1
-recorded_at = "2026-08-12T05:04:16Z"
-git_sha = "a51ffdf12b5caf288cc7028fa5fbd86449c0d518"
+recorded_at = "2026-08-16T04:00:02Z"
+git_sha = "2273e45dbf4e23231743f428bd2e53d978b91e9a"
 
 [artifacts.baml-cli]
-file_bytes = "20.6 MiB"
-stripped_bytes = "20.6 MiB"
-gzip_bytes = "9.8 MiB"
+file_bytes = "24.2 MiB"
+stripped_bytes = "24.2 MiB"
+gzip_bytes = "10.6 MiB"
 
-# Re-recorded from CI run 31906918242: BEP-066 (evaluation/type-construction/
-# reflection) VM+bytecode surface plus the native provider stdlib additions
-# since the 2026-08-12 record (EnvRef, ai.internal prompt plumbing).
+# Re-recorded from CI run 31925567696: Package.compile's compiler-built stdlib
+# prefix is embedded in bex_project so first-call runtime compilation does not
+# rebuild the full stdlib.
 [artifacts.packed-program]
-file_bytes = "16.6 MiB"
-gzip_bytes = "7.2 MiB"
+file_bytes = "19.6 MiB"
+gzip_bytes = "7.8 MiB"
--- a/baml_language/.ci/size-gate/wasm32-unknown-unknown.toml
+++ b/baml_language/.ci/size-gate/wasm32-unknown-unknown.toml
@@ -1,9 +1,10 @@
 version = 1
-recorded_at = "2026-08-12T05:04:16Z"
-git_sha = "a51ffdf12b5caf288cc7028fa5fbd86449c0d518"
+recorded_at = "2026-08-16T04:00:02Z"
+git_sha = "2273e45dbf4e23231743f428bd2e53d978b91e9a"
 
-# Re-recorded from CI run 31906918242: BEP-066 VM+bytecode surface plus the
-# native provider stdlib additions since the 2026-08-12 record.
+# Re-recorded from CI run 31925567696: Package.compile's compiler-built stdlib
+# prefix is embedded in bex_project so first-call runtime compilation does not
+# rebuild the full stdlib.
 [artifacts.bridge_wasm]
-file_bytes = "17.0 MiB"
-gzip_bytes = "4.6 MiB"
+file_bytes = "20.3 MiB"
+gzip_bytes = "5.1 MiB"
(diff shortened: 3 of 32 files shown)


antoniosarosi:
# Adversarial review — stdlib-prefix PR

Same deep-review process as #4450. **Verdict: SHIP-WITH-FIXES — three blockers.** The two hardest surfaces attacked — artifact staleness and tagged-pointer ABA on the cache keys — both hold up under inspection (proofs in the clean-areas section below). Note: the PR body/log describe a `runtime_dispatch_id`; no such identifier exists in the code — the implemented scheme is `function_object.as_ptr() | 1` (tagged, GC-forwarded heap address). The implemented scheme is actually sound; the description should be corrected to match it.

## Blockers

**H1 — Session declarations lose runtime identity AND package provenance** (`bex_vm/src/package_baml/reflect.rs:1632-1640` deletion). Removing `allocate_runtime_declaration_types` from `graft_session_submission` dropped not just the mint pre-allocation but the `RuntimeTypeProvenance` assignment. Consequences: (a) `type.of<C>()` for a session-declared class now yields a structural static mint — two sessions declaring identical `class C { x int }` produce **mint-equal** type values, violating generative identity (a ratified, non-negotiable semantic); (b) `value_runtime_package` returns null for session-class instances, so host-side interface dispatch on them misses the session's grafted impl rules → `UnresolvedVirtualCall`. Green suites miss it because no test declares the same class in two sessions or dispatches a session class's impl from host code. Restore the provenance path (or an equivalent) and add both regression tests.

**H2 — Runtime cache keys use the frame's callee object, not the underlying Function** (`vm.rs:2672`, `:2699` vs the static path at `:7207`/`:7907`). Closures/bound methods are fresh heap allocations per `MakeClosure`, so runtime-code lambdas doing `VirtualCall`/`LoadType` get a new key per closure instance: permanent cache misses plus per-closure pinned `Object::Type` entries (cache values are GC roots). Fix: resolve through Closure/BoundMethod to the shared underlying `Object::Function` pointer before tagging. The regression/benches missed it because they call from a named function where frame.function IS the Function.

**H3 — `Package.compile` dependency aliases skip the reserved-name check, and the new stdlib fallback consults them first** (`reflect.rs:639-647` has the check in `Session::_new`; `_finish` at `:1777` does not; `vm.rs:1976-1990` checks `dependency_names` before the stdlib set). `compile(files, {"baml": p})` makes runtime `baml.*` lookups resolve into `p` while the compiler type-checked against the real stdlib — check/run divergence, wrong-class resolution on name collision. Mirror the `Session::_new` reserved-name rejection in `_finish` and/or test the stdlib set first.

**Plus the missing oracle**: no test compares a `Package.compile` artifact via the prefix against the old full-compile path — the byte oracle that would have caught H1. `stdlib_splice_is_byte_identical_to_full_compile` still exercises the old branch (its db has stdlib source, so the `builtin_count == 0` branch never runs). Add the prefix-vs-full comparison, normalized like `mounted_package_parity` does.

## Follow-ups (non-blocking)
- M4: minor GCs evict most runtime cache entries (Gen2 objects aren't traced into the forwarding map), so the ±5% parity number represents allocation-free loops; add a minor-GC bench variant or keep-if-live semantics.
- M5: `mark_precompiled_callables_linkable` promotes `Intrinsic`/`AwaitAny` too — an unintercepted stdlib intrinsic becomes "runtime link failed" instead of E0158. Promote only `Vm`/`Io`.
- M6: gate `external_callee`'s builtin_kind trust on `is_precompiled_package` (defence-in-depth).
- M7: add `OptLevel` (and `emit_test_cases`) to `ARTIFACT_KEY`; note the build-dependency weight and artifact size in the PR body (Size Gate relevance).
- L8-L12 as listed in the full report (decompose_units ordinary-path edge, class_static_value precedence test, `.expect()`s in impls_for_type, odd-tag debug_assert, `is_external_package` scan on the hot path).
- PR-body honesty edits: the `runtime_dispatch_id` description, and the +2.46% delta is quoted from mixed sampling regimes with setup-subtraction variance ≥ the delta — state it directionally, not as a bound.

## Verified clean
Artifact staleness (include_str! → dep-info → build-script re-run chain is airtight; producer/consumer cannot desync), tagged-key ABA (every collection forwards roots; an address cannot be recycled while a stale entry survives), cross-process safety (no serialized identities), the ReservedBuiltin trust boundary (user mounts cannot claim the promotion; E0158 regression still binds), the ordinary compile path (prefix machinery degenerates when the precompiled set is empty), interface-row method parity, and the benchmark drives the real public path with correct pins.

PR #4430, opened 2026-08-14
Native BAML LLM provider clients: delete sys_llm, auth-only Rust

aaronvg:
## What

Converts the entire LLM provider layer to native BAML under `baml_std/`, deletes `crates/sys_llm` (20,453 lines) and `forks/aws-bedrock`, and fixes 8 compiler bugs the migration surfaced. Rust's remaining provider role is exactly auth: the new `crates/sys_auth` (GCP OAuth minting, AWS SigV4 + credential chain) exposed as `ai.internal._*` sysops.

### Providers (all typed wire classes end to end)

- **openai**: `ResponsesClient` (renamed from `OpenAiClient`; typed rewrite + streaming fixes + `image_generation` tool), new `ChatClient` / `GenericClient` / `AzureClient` / `OllamaClient` / `OpenRouterClient` over one shared `ChatCompat` core, new `ImageClient` (`/v1/images/generations`)
- **anthropic**: typed rewrite; 6 verified bugs fixed (streaming error events, stop reasons, cache tokens, reasoning-only turns…); `cache_control` via per-message metadata
- **google**: Gemini typed rewrite + new `VertexClient` (OAuth, express mode, project/location chains, **Claude-on-Vertex** via `publishers/anthropic:rawPredict` reusing the anthropic body builder)
- **aws**: new `BedrockClient` (Converse; SigV4-signed as the final request mutation; toolConfig + cachePoint net-new)
- **vercel**: `AiGatewayImageClient` (exact sys_llm port + mime bug fix)
- 8 new `"provider/model"` shorthand prefixes

### Platform work

- Media outputs: `ai.content.Media` + runner coercion (sys_llm contract, byte-identical errors) — `-> image` functions work end to end
- Structural prompt parts + per-message metadata; `ai.wire` seams (`merge_request_body` null-deletes, journal sanitization, `resolve_media`)
- Stream termination contract (`require_terminal`): truncated SSE is a typed error, never a silent `Complete`; `Retry-After` populates `RateLimited`
- Test infra: `ns_llm_mock` kit, `[test]` profiles (offline default; live via `infisical run … --profile live`), ~250 new provider/regression tests

### Compiler fixes (minimal repros: `_plan/sys_llm_native/compiler-bugs/min-repros/`)

[...]

PR #4458, opened 2026-08-16
perf(compiler): memoize canonical body inference facts

antoniosarosi:
## Summary

- memoize recursive alias/enum facts and canonical interned forms within one body-inference context
- reject distinct stable nominal heads before canonicalization and fuse external-package classification into one mounted-package lookup
- add cold/warm cache coverage against a fast-reject-free canonical oracle, including a load-bearing interface `requires` relation
- assert the infer-free cache invariant at the private canonicalization boundary

## Diagnosis

The supplied #4430 profile showed body inference scaling super-linearly: the empty compile moved from about 368 ms to 989 ms, the realistic project from 2.39 s to 4.61 s, and the worst OpenAI bodies reached 500–660 ms.

Temporary phase/counter instrumentation on the current canary workload inferred 2,160 bodies. It ruled out two suspected causes: the entire compile registered/attempted only 42–51 obligations, and package-resolution misses were zero. A single-thread profile removed parallel lock-wait attribution and isolated recursive JSON/provider walkers instead. For example, `baml/ns_toml/toml.baml|item_to_json` performed 1,602 canonical roots / 5,542 canonical nodes; OpenAI metadata/directive walkers repeatedly traversed roughly 175–382 roots and 1,945–3,055 nodes per body.

The hot path rebuilt canonical forms for every subtype/equivalence query and repeatedly materialized the same recursive alias definitions and enum rows from interned compiler data. This change scopes both caches to one inference body and one immutable fact set; inference-bearing types bypass the canonical cache because their meaning is table-relative.

All diagnostic instrumentation and temporary proof drivers were removed before commit.

## Compiler A/B

Rust 1.93.0, `CARGO_BUILD_JOBS=8`, `BAML_PROFILE=0`, same machine and command on `a3bc09ba7` canary and this branch:

```sh
cargo bench -p baml_tests --bench compiler_benchmark -- compile_
```

| Benchmark | canary median | branch median | delta |
|---|---:|---:|---:|
| `compile_empty_project` | 530.2 ms | 489.1 ms | -41.1 ms (-7.75%) |
| `compile_baml_tests_project` | 2.333 s | 2.120 s | -0.213 s (-9.13%) |

Ranges: empty canary 511.6–553.8 ms vs branch 465.6–501.1 ms (100 samples); realistic canary 2.319–2.515 s vs branch 2.103–2.269 s (5 samples). This clears the brief's <500 ms empty median and <=2.8 s realistic ceilings.

Controlled per-body profile, 2,160 bodies on each side:

| Body | canary | branch | delta |
|---|---:|---:|---:|
| `openai/chat.baml|invoke_stream` | 196.4 ms | 131.6 ms | -33.0% |
| `openai/ns_internal/chat.baml|chat_invoke_stream` | 192.9 ms | 128.7 ms | -33.3% |
| `openai/generic.baml|invoke_stream` | 182.4 ms | 125.9 ms | -31.0% |
| `openai/ns_internal/chat.baml|chat_build_request` | 148.9 ms | 106.0 ms | -28.8% |

## Peak RSS A/B

Required review remeasurement of `compile_baml_tests_project`: release benchmark executable run directly under `/usr/bin/time`, three fresh processes per side, five Divan samples per process, `BAML_PROFILE=0`, and identical default/full-machine conditions on canary `a3bc09ba7` and the final candidate. Table values are medians across the three processes.

| Measurement | canary | branch | delta |
|---|---:|---:|---:|
| Peak RSS | 714,960 KiB (698.2 MiB) | 710,832 KiB (694.2 MiB) | -4,128 KiB / -4.0 MiB (-0.58%) |
| Compile median | 2.362 s | 2.193 s | -0.169 s (-7.16%) |

Peak-RSS ranges were 711,464–716,612 KiB on canary and 707,680–715,836 KiB on the branch.

The suggested `Rc<NormalTy>` memo was evaluated in the same three-process setup and dropped because it did not measure as a win: compile median 2.193→2.200 s (+0.3%) and peak RSS 710,832→712,864 KiB (+0.3%). The final patch retains the simpler owned cache. `provable_subtype` threading is intentionally deferred.

## Runtime safety

The full generated runtime speedtest suite ran in a same-machine A-B-A bracket with five samples per workload (`BAML_PROFILE=0`, `DIVAN_MAX_TIME=20`). All 38 non-sleep workloads completed; stable representative medians stayed flat (for example method call +1.8%, bubble sort +0.7%, call chain -0.8%, pure call +2.8%, interface default dispatch +2.4%, interface match dispatch +1.7%). Concurrency rows showed the expected broad scheduler variance across the bracket.

As a stronger zero-delta proof, a temporary harness compiled and Borsh-serialized every measured workload on canary and the branch. Both revisions produced exactly 38 programs, 94,239,414 aggregate bytes, and corpus digest `838ab2f95eb1bc62`: the VM received byte-identical programs. The harness was removed before commit.

## Validation

- reject-free canonical-cache oracle covers cold/warm recursive aliases, unions/literals, enum collapse, distinct nominal heads, and `Readable <: Displayable` through populated `requires`
- `cargo nextest run -p baml_type -p baml_compiler2_hir -p baml_compiler2_hir_ty --all-features`: 384/384 passed
- strict all-target/all-feature Clippy for the focused packages passed with warnings denied
- all three commit hooks passed full-workspace Clippy
- repeated pinned gate passed: 3,742/3,742 tests (59 slow), 24 skipped, in 1,505.171 s; doctests passed/ignored as expected
- Insta: no unreferenced snapshots and no snapshots to review

Pinned command, rerun with default cargo/nextest parallelism per the corrected coordinator instruction:

```sh
rustup run 1.93.0 cargo insta test --test-runner nextest \
  -p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests \
  -p baml_surface --all-features --unreferenced=reject
```

## Summary by CodeRabbit

* **Performance**
  * Improved type-checking responsiveness by caching repeated type comparisons and resolved type information.
  * Added faster handling for clearly unrelated type categories.

* **Bug Fixes**
  * Improved classification of mounted, precompiled, reserved, and standard-library packages.
  * Added coverage for recursive aliases, literals, unions, enums, classes, and interface requirements to ensure cached type results remain accurate.

### DIFF
--- a/baml_language/crates/baml_compiler2_hir/src/package.rs
+++ b/baml_language/crates/baml_compiler2_hir/src/package.rs
@@ -258,10 +258,13 @@ mod tests {
     };
 
     use baml_base::{FileId, SourceFile};
-    use baml_workspace::{Compiler2ExtraFiles, Project};
+    use baml_workspace::{Compiler2ExtraFiles, MountedPackages, Project};
     use salsa::Setter;
 
-    use super::{PackageId, is_allowed_builtin_namespace_shadow, package_items};
+    use super::{
+        PackageId, is_allowed_builtin_namespace_shadow, is_external_package, is_mounted_package,
+        is_precompiled_package, package_items,
+    };
     use crate::Db;
 
     #[salsa::db]
@@ -270,6 +273,7 @@ mod tests {
         next_file_id: AtomicU32,
         project: Option<Project>,
         extra: Option<Compiler2ExtraFiles>,
+        mounted: Option<MountedPackages>,
     }
 
     impl Default for TestDb {
@@ -279,6 +283,7 @@ mod tests {
                 next_file_id: AtomicU32::new(0),
                 project: None,
                 extra: None,
+                mounted: None,
             }
         }
     }
@@ -301,6 +306,15 @@ mod tests {
             db.extra = Some(Compiler2ExtraFiles::new(&db, builtin_files));
             db
         }
+
+        fn with_mounts(
+            by_package: std::collections::BTreeMap<String, Vec<u8>>,
+            immutable_precompiled: std::collections::BTreeSet<String>,
+        ) -> Self {
+            let mut db = Self::default();
+            db.mounted = Some(MountedPackages::new(&db, by_package, immutable_precompiled));
+            db
+        }
     }
 
     #[salsa::db]
@@ -310,6 +324,10 @@ mod tests {
     impl baml_workspace::Db for TestDb {
         fn project(&self) -> Project {
             self.project.expect("test db initialized")
+        }
+
+        fn mounted_packages(&self) -> Option<MountedPackages> {
+            self.mounted
         }
     }
 
@@ -365,6 +383,31 @@ mod tests {
             &[baml_base::Name::new("other")],
             id_def
         ));
+    }
+
+    #[test]
+    fn external_package_fast_path_matches_composed_classification() {
+        let absent = TestDb::default();
+        assert!(!is_external_package(&absent, &baml_base::Name::new("app")));
+
+        let by_package = ["app", "baml", "log", "user"]
+            .into_iter()
+            .map(|name| (name.to_owned(), Vec::new()))
+            .collect();
+        let immutable_precompiled = ["baml", "user", "missing"]
+            .into_iter()
+            .map(str::to_owned)
+            .collect();
+        let db = TestDb::with_mounts(by_package, immutable_precompiled);
+
+        for raw_name in ["app", "baml", "log", "user", "missing", "env"] {
+            let name = baml_base::Name::new(raw_name);
+            assert_eq!(
+                is_external_package(&db, &name),
+                is_mounted_package(&db, &name) || is_precompiled_package(&db, &name),
+                "fused classification diverged for {raw_name}"
+            );
+        }
     }
 }
 
@@ -449,7 +492,17 @@ pub fn is_precompiled_package(db: &dyn crate::Db, name: &Name) -> bool {
 /// Whether `name` is any source-less dependency served from a serialized
 /// `PackageInterface`.
 pub fn is_external_package(db: &dyn crate::Db, name: &Name) -> bool {
-    is_mounted_package(db, name) || is_precompiled_package(db, name)
+    let Some(mounted) = db.mounted_packages() else {
+        return false;
+    };
+    if !mounted.by_package(db).contains_key(name.as_str()) {
+        return false;
+    }
+    if !is_reserved_package_name(name.as_str()) {
+        return true;
+    }
+    baml_builtins2::stdlib_package_names().contains(&name.as_str())
+        && mounted.immutable_precompiled(db).contains(name.as_str())
 }
 
 /// The *direct* dependencies of `package_id` (hardcoded for now).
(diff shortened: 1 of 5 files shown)


antoniosarosi:
# Adversarial review — canonicalization memoization

Deep-review (same process as #4450/#4453). **Verdict: SHIP-WITH-FIXES — no semantic defect found; blockers are test/assert hardening.**

## Held under attack (proofs in short)
- **Fact immutability per body**: one `Facts` per body, never reassigned; `&'db dyn Db` borrow makes mid-inference input mutation a compile error; salsa cycle iterations build fresh contexts. Scoped `type T = unreflect(...)` rigid params are statement/pattern-derived (`0x8000_0000|fnv(owner,stmt)` / `0xc000_0000|fnv(owner,pat)`), so sibling scopes get distinct handles — no handle-equal/meaning-different case; exit erasure produces new types, never mutates handles.
- **Infer exclusion**: both entry points guard on `has_infer()` (covers `Infer(Some)` and `_` holes via the intern-time flag); a miss panics (`unreachable!` in `from_interned`) before insertion — fail-loud, not poison. Skolemized vars verified non-aliasing.
- **Keys/values**: hash-cons ptr identity sound; ABA dismissed (map keys pin pool entries); fuel/assumption-sensitive recursive canonicalization is NOT memoized (only top-level canonical forms — the co-inductive threading is untouched); relations recomputed per query.
- **Fast-reject parity**: `interned_heads_definitely_differ` textually matches the plain rule; alias heads correctly undecided; Interface/Interface excluded from the subtype reject; audited every cross-name arm of `is_subtype_of_inner` incl. enum-collapse and sentinels.
- **Fused classification**: boolean-algebra-exact vs the original incl. reserved-name ordering; #4453's tests still bind. (Minor note: `env.X`-style paths now record a `MountedPackages` salsa dep they didn't before — net wash.)
- Instrumentation fully removed; diff has no snapshot changes; RefCell re-entrancy and `!Sync` sharing checked.

## Blocking (small)
1. **The new parity test is a tautology** (`normalize/tests.rs:1624-1663`): it compares the cache against `is_subtype_interned`/`equivalent_interned` — which now contain the SAME fast-reject; and the existing plain-entry cross-oracle has no two-distinct-heads pairs. A fast-reject bug would pass everything. Add a test whose oracle is the reject-free path (`canonical_interned` forms compared/subtyped directly), with a context whose `requires` is populated so the Interface/Interface exclusion is load-bearing.
2. **`debug_assert!(!ty.has_infer(), ..)` in `InternedCanonicalCache::canonical`** — the invariant is currently enforced only at the two hir_ty call sites; future callers (notably the §follow-up below) deserve a test failure, not a release `unreachable!`.

## Before merge (measurement ask)
- Peak-RSS before/after on `compile_baml_tests_project`: the per-body memo grows unbounded (keys fragmented by attrs/freshness) and each key pins its interned subtree against pool eviction for the body's life; the log reports wall time only.

## Follow-ups (not blocking; both likely widen the win)
- Store `Rc<NormalTy>` in the memo + `Rc::ptr_eq` fast path in `equivalent` — hits currently deep-clone ~5k-node trees, so a hit is still O(nodes).
- Thread the cache through `provable_subtype` (infer/pat.rs:1476-1489 + the three infer.rs call sites) — the recursive JSON/provider walker path the diagnosis actually named goes through UNCACHED `is_subtype_interned` today; it takes `&Facts` so it structurally can't reach the body cache. (This is also exactly the caller that makes blocker 2's assert earn its keep.)
- `AliasOnlyFacts` short-lived instances pay two RefCell allocs for memos that never hit — scope the memo to long-lived `Facts` or hoist those contexts.

PR #4460, opened 2026-08-16
fix: harden runtime compilation boundaries

antoniosarosi:
## Summary

- Reject runtime-checked arguments on indirect calls during inference with the shared `E0010` diagnostic. Indirect opcodes cannot encode the required runtime check, so this removes the former debug-only emitter panic and release-mode check omission.
- Sanitize synthetic mounted-package stubs only when a return or interface field type directly or recursively contains a runtime-minted qualified name. Ordinary source-spellable mounted ABI types remain intact instead of being broadly erased to `unknown`.
- Restore the HIR type sweep's file and typed-node coverage counts, together with the bytecode display fixture and freshly generated textual, expanded, and unoptimized snapshots.
- Treat mounted `external_callee.builtin_kind` metadata as compiler-owned only for precompiled packages, preventing ordinary mounted packages from selecting privileged builtin lowering.

## Impact

Runtime-dependent generic checks now fail consistently before emission in both debug and release builds. Runtime-created types no longer leak hidden names through phantom `runtime_mount_*` diagnostics, while normal mounted return and interface types retain their source ABI. The restored sweep and bytecode snapshots re-establish regression coverage for HIR typing and VM display output.

## Validation

Focused regressions passed:

- `indirect_runtime_checked_call_is_rejected_before_emission` in debug and release modes
- `mounted_await_any_kind_is_trusted_only_for_precompiled_packages`
- `s15_sweep_baml_src`
- `bytecode_display_formats`
- `runtime_minted_name_detection_is_recursive`
- `mounted_runtime_types_do_not_leak_into_phantom_stub_diagnostics`
- `pack_e2e_current_package_compiles_and_runs_skill`

Formatting and diff checks passed:

- `cargo fmt --all -- --check`
- `git diff --check`

Full pinned gate from `baml_language`:

```text
CARGO_BUILD_JOBS=8 rustup run 1.93.0 cargo insta test --test-runner nextest -p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface --all-features --unreferenced=reject
```

Exit 0. Summary: `[1513.387s] 3746 tests run: 3746 passed (72 slow), 24 skipped`; doctests clean, no unreferenced snapshots, and no snapshots to review.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added a clear `E0010` diagnostic when runtime type arguments are used with unsupported indirect calls.
  - Prevented external package calls and intrinsics from being misclassified as built-in operations.
  - Improved runtime package compilation for nested and generic types, preventing misleading stub-related errors.

- **Tests**
  - Expanded coverage for compiler diagnostics, package compilation, call handling, and bytecode display formatting.
  - Added bytecode snapshots covering optimized and unoptimized output.

## Review fix round

- Added E0010 validation for optional indirect calls and reworded E0010 around runtime-checked arguments.
- Re-pinned the mounted-builtin regression to the real `$await_any` lowering surface and gated external `log.*`/`type.of` intrinsic recognition on compiler-built precompiled packages.
- Added a changelog compatibility note and a debug assertion protecting mounted-package/precompiled-stdlib installation order.
- Focused post-review validation: 2/2 mounted MIR trust tests, 6/6 runtime diagnostic consistency tests, 15/15 runtime package compilation tests, the diagnostic constructor test, and the legitimate `reflect_type_of_class` MIR test.

## Deferred follow-ups

- **A-2:** restore a non-silent back-end guard for front-end/emitter oracle divergence, covering `GenericFunction` callees and globals-absent functions.
- **A-5:** broaden indirect-call shape coverage to bound-method values, field-held/IIFE/returned/container callables, negative cases, and `Package.compile` parity.
- **B residuals:** replace `$dyn`-spelling heuristics with a builtin-or-stubbed-here nominal-leaf rule, covering two-hop QTNs, `with_types` compiled classes, exported aliases without stubs, non-source future spelling, frame-generic binders, and the missing interface-arm `source_identifier` guard.


antoniosarosi:
# Adversarial review — runtime hardening batch

Same deep-review process as #4450/#4453/#4458. **Verdict: SHIP-WITH-FIXES.** Three of four items sound in substance; C (restored coverage) is fully clean with freshness proven (namespace/method/provider markers impossible in the old bytes; snapshot counts track today's corpus).

## Must-fix
1. **A-1 (HIGH): `f?.(…)` bypasses the new E0010** — `report_runtime_indirect_call` is called only from the plain-call arm; `OptionalCall` delegates to the same lowering (`lower_optional_call` → `lower_call`) so the runtime check is silently dropped in debug AND release, contradicting the PR body. The fix + test already sit UNCOMMITTED in the working tree — commit and push them.
2. **D-1/D-2 (MEDIUM): the M6 test does not test the gate** — it asserts no `SysOp`, but sys-op selection never reads `external.builtin_kind` (external resolutions return None there); the assertion held before the change. The gate's only real surface is `AwaitAny` via `check_await_any` — make the fixture a `$await_any` body and assert no `Terminator::AwaitAny`.
3. **D-3 (MEDIUM): two trust sites left ungated** — `check_intrinsic` and `check_type_of_intrinsic` still trust blob-supplied `builtin_kind == Intrinsic` (bounded only by name matches); the forged-blob escape the new test models for SysOp remains open there. Gate them too, or state the scoping rationale in the PR body.

## Should-fix (cheap, fold in)
- A-3: E0010's message says "runtime type arguments" but the only reachable trigger is runtime-checked *arguments* — reword.
- Changelog line: E0010 turns previously-compiling (silently unchecked) release programs into compile errors.
- D-4: assert/debug-check the set_mounted_packages → set_precompiled_stdlib_packages ordering dependency.

## Follow-up ticket (not this PR)
- A-2: restore a non-silent back-end guard for the front-end/emit oracle divergence (assert-if-body-error-free, or trap on runtime-checked indirect residue; `GenericFunction` callees and globals-absent functions are the known residues).
- A-5: broaden indirect-shape coverage (bound-method value, field-held callable, IIFE, returned callable, container element, negatives, Package.compile parity in the consistency oracle).
- B residuals: the interface/return-arm fix keys on `$dyn` spelling, but phantom-diagnostic classes remain (two-hop dependency QTNs, `with_types` of compiled classes, exported aliases with no stub, `future<V,E>` non-source spelling, frame-generic binders, and the missing `source_identifier` guard on the interface arm). Recommended rewrite: spell the real type only when every nominal leaf is builtin-or-stubbed-here, else `unknown` — subsumes the current rule.

## Verified clean
A's allowlist/ordering/false-positive analysis (allowlist mirrors MIR's own table; plan writes precede the report; scenario corpus untriggerable), B's masking risk unfounded (blob is the ABI, stub units discarded, `unknown` is top so over-widening only over-rejects, no diagnostics filtered), C fully (fresh snapshots proven, plumbing/auto-discovery verified), D's no-regression claim for legitimate stdlib builtins (ordinary compiles never reach the gate; runtime path orders correctly; Vm/Io never flowed through it).

PR #4461, opened 2026-08-17
perf(compiler): reuse inference caches

antoniosarosi:
## Summary

- Thread the existing canonical subtype cache through provable_subtype call sites while retaining equality and infer/error filtering before cache entry.
- Extract uncached definition and enum lookup helpers so short-lived AliasOnlyFacts avoids allocating memo maps.
- Keep memoization on long-lived Facts, with the existing debug assertions preserving cache consistency checks.

This is a pure performance refactor with no intended semantic delta.

## Benchmarks

| benchmark | canary baseline | candidate repeats | repeat-median delta |
| --- | ---: | ---: | ---: |
| compile_baml_tests_project | 2.114 s | 2.193 s / 2.194 s | +3.8% |
| compile_empty_project | 490.7 ms | 505.5 ms / 504.4 ms | +2.9% |

Both deltas are within the established +/-5% run-to-run band. The aggregate result is neutral/noisy, so this PR does not claim an end-to-end speedup; the win is the removal of redundant cache construction and canonical subtype work.

## Scope note

The optional #4453 M4 item was skipped because it belongs to separate VM/GC benchmark and retention work rather than this contained compiler follow-up.

## Validation

- Focused nextest: 384/384 passed for baml_type, baml_compiler2_hir, and baml_compiler2_hir_ty with all features.
- Strict Clippy passed for those packages across all targets and features.
- Cargo format and diff checks passed.
- Exact pinned full gate passed: 3748/3748 tests, 24 skipped, 51 slow, 1409.829 s; doctests clean; no unreferenced snapshots and no snapshots to review.

Full gate command: CARGO_BUILD_JOBS=8 rustup run 1.93.0 cargo insta test --test-runner nextest -p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface --all-features --unreferenced=reject

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type resolution for aliases and enum variants.
  - Improved type checking for pattern matching, assignment narrowing, and catch-arm handling.
  - Made subtype checks more consistent across interfaces, classes, arrays, unions, and pattern-based matching.
  - Improved handling of type narrowing and subtraction in flow analysis.

- **Performance**
  - Reduced repeated subtype calculations during compilation, which may improve compiler responsiveness for complex type definitions.

### DIFF
--- a/baml_language/crates/baml_compiler2_hir_ty/src/facts.rs
+++ b/baml_language/crates/baml_compiler2_hir_ty/src/facts.rs
@@ -59,9 +59,55 @@ impl<'db> Facts<'db> {
     /// Resolves a qualified name back to its definition through the owning
     /// package's canonical (ppir) items.
     pub fn definition_of(&self, name: &QualifiedTypeName) -> Option<Definition<'db>> {
-        let package = PackageId::new(self.db, name.package().clone());
-        baml_compiler2_ppir::package_items(self.db, package)
-            .lookup_type(name.namespace(), name.name())
+        definition_of(self.db, name)
+    }
+}
+
+fn definition_of<'db>(
+    db: &'db dyn baml_compiler2_ppir::Db,
+    name: &QualifiedTypeName,
+) -> Option<Definition<'db>> {
+    let package = PackageId::new(db, name.package().clone());
+    baml_compiler2_ppir::package_items(db, package).lookup_type(name.namespace(), name.name())
+}
+
+/// Resolves an alias without allocating a memo table. Short-lived fact-poor
+/// contexts use this directly; body-lifetime [`Facts`] caches its result.
+pub(crate) fn uncached_alias_def(
+    db: &dyn baml_compiler2_ppir::Db,
+    name: &QualifiedTypeName,
+) -> Option<Ty> {
+    if let Some(Definition::TypeAlias(alias)) = definition_of(db, name) {
+        return Some(crate::lower::type_alias_value(db, alias).to_plain());
+    }
+    match crate::package_interface::mounted_type_row(db, name) {
+        Some(crate::package_interface::ExportedType::TypeAlias { resolved, .. }) => {
+            Some(resolved.clone())
+        }
+        _ => None,
+    }
+}
+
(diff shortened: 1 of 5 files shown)


antoniosarosi:
# Post-merge audit — this PR merged before its adversarial review completed

Verdict had this been pre-merge: **SHIP-WITH-FIXES**. Part (a) — threading the canonical cache through `provable_subtype` — is clean on all four attack angles (infer-filtering precedes every cache entry across all 16 routes with a deep `has_infer` flag verified variant-by-variant; no cross-body sharing is constructible; scoped-rigid/skolem safety carries over; the cached oracle is a line-for-line mirror). Part (b) — removing the `AliasOnlyFacts` memos — has two MAJOR problems now live on canary:

**F1 (likely perf regression):** the memo removal hit `impls.rs:808` (`impls_for_type_cached`) and `impls.rs:1114` (`resolve_within_depth`) — single instances deliberately hoisted across all-packages × all-impl-candidate scans. Alias-headed types there now re-run `type_alias_value` (NOT salsa-tracked — full RHS re-lowering) per candidate instead of once per scan; a `List<SomeAlias>` receiver against ~136 stdlib implements blocks re-lowers the alias ~136×. The stated rationale ("avoids allocating memo maps") is false: empty `FxHashMap`s don't allocate. Keep the memo-free form only at the four genuinely per-call sites (`Facts::project`, obligations.rs:527, method_resolution.rs:753, impls.rs:1757); restore memoization at the two scan sites.

**F2 (measurement honesty):** the PR's bench baseline was NOT rerun in-session (the log says so), and both table rows are regressions (+3.8% / +2.9%) waved off as within-band — `compile_empty_project` has no user bodies so part (a) barely runs there, while F1's scans do; a uniform ~3% is more consistent with F1 than with noise. Rerun back-to-back, split (a)-only vs (a)+(b).

Minor: F4 — make `provable_subtype` an `InferenceContext` method so the facts/cache pairing is structural; F3 — fix the "short-lived contexts" doc claim; F6 — one-sentence doc caveat on the salsa-cycle window widening (inherited from #4458, not a blocker).

Fast-follow PR dispatched.

PR #4463, opened 2026-08-17
perf(compiler): restore alias memos for impl scans

antoniosarosi:
## Summary

- restore alias and enum memoization only in the project-scale candidate scans in impls_for_type_cached and resolve_within_depth
- keep the memo-free AliasOnlyFacts constructor at Facts::project, infer/obligations.rs, method_resolution.rs, and interface_requires_inner
- make provable_subtype an InferenceContext method so the body facts and canonical cache are structurally paired at every call
- correct the one-shot context documentation and document the Salsa re-entrancy-window caveat for scan-local memos

This is the focused correction for the #4461 post-merge audit. It restores repeated alias-head lookup reuse without widening memoization to the four one-shot sites.

## Fresh back-to-back A/B

Both refs were measured consecutively in the same shell and target cache with profiling disabled. Both workloads used exactly five one-compile samples:

cargo bench -p baml_tests --bench compiler_benchmark -- compile_ --sample-count 5 --sample-size 1

Environment: Rust 1.93.0, CARGO_BUILD_JOBS=8, BAML_PROFILE=0.

| benchmark | canary 07c118b68, 5 samples | candidate 0eb5f0915, 5 samples | median delta |
| --- | ---: | ---: | ---: |
| compile_baml_tests_project | 2.180 s (2.144-2.351 s) | 2.133 s (2.122-2.295 s) | -47 ms (-2.2%) |
| compile_empty_project | 497.5 ms (485.8-499.9 ms) | 485.6 ms (478.2-488.5 ms) | -11.9 ms (-2.4%) |

No earlier baseline values are reused. Canary 07c118b68 was the exact #4461 tip when this branch and A/B began; #4462 landed afterward and changes only nextest scheduling configuration.

## Validation

- cargo check -p baml_compiler2_hir_ty
- focused nextest: 384/384 passed for baml_type, baml_compiler2_hir, and baml_compiler2_hir_ty with all features
- strict all-target/all-feature Clippy passed for those packages; workspace commit-hook Clippy also passed
- cargo fmt --all -- --check and git diff --check
- exact pinned full gate: 3,748/3,748 passed, 24 skipped, 50 slow in 1,375.321 s
- doctests clean; no unreferenced snapshots and no snapshots to review

Full gate command: CARGO_BUILD_JOBS=8 rustup run 1.93.0 cargo insta test --test-runner nextest -p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface --all-features --unreferenced=reject

## Summary by CodeRabbit

* **Improvements**
  * Improved type inference consistency across assignments, pattern matching, flow narrowing, and catch-arm analysis.
  * Enhanced handling of aliases and enum variants during type resolution.
  * Repeated type checks are now more efficient, helping reduce unnecessary compiler work.
  * Existing subtype, union, coverage, and narrowing behavior remains consistent while benefiting from more centralized analysis.

### DIFF
--- a/baml_language/crates/baml_compiler2_hir_ty/src/facts.rs
+++ b/baml_language/crates/baml_compiler2_hir_ty/src/facts.rs
@@ -71,8 +71,8 @@ fn definition_of<'db>(
     baml_compiler2_ppir::package_items(db, package).lookup_type(name.namespace(), name.name())
 }
 
-/// Resolves an alias without allocating a memo table. Short-lived fact-poor
-/// contexts use this directly; body-lifetime [`Facts`] caches its result.
+/// Resolves an alias without retaining the result in a memo. One-shot
+/// fact-poor contexts use this directly; repeated scans use a cached context.
 pub(crate) fn uncached_alias_def(
     db: &dyn baml_compiler2_ppir::Db,
     name: &QualifiedTypeName,
@@ -88,8 +88,8 @@ pub(crate) fn uncached_alias_def(
     }
 }
 
-/// Resolves enum variants without allocating a memo table. Short-lived
-/// fact-poor contexts use this directly; body-lifetime [`Facts`] caches it.
+/// Resolves enum variants without retaining the result in a memo. One-shot
+/// fact-poor contexts use this directly; repeated scans use a cached context.
 pub(crate) fn uncached_enum_variants(
     db: &dyn baml_compiler2_ppir::Db,
     name: &QualifiedTypeName,
--- a/baml_language/crates/baml_compiler2_hir_ty/src/impls.rs
+++ b/baml_language/crates/baml_compiler2_hir_ty/src/impls.rs
@@ -256,20 +256,39 @@ pub fn package_impl_locs<'db>(
 /// re-enter the resolver that called it).
 pub(crate) struct AliasOnlyFacts<'db> {
     db: &'db dyn baml_compiler2_ppir::Db,
+    memoized: Option<crate::facts::Facts<'db>>,
 }
 
 impl<'db> AliasOnlyFacts<'db> {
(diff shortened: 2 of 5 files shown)


antoniosarosi:
# Adversarial review — memo-scan fix

**Verdict: SHIP-WITH-FIXES — one doc sentence pre-merge; all load-bearing content clean.**

Verified clean: memo scope is structurally per-scan (body-local in a salsa-tracked query — rebuilt on every cycle-fixpoint re-execution, so provisional-state rows can't leak; private visibility enforces non-escape); the per-call/per-scan site split independently re-derived and correct (the four memo-free sites bound by the requires-DAG, not the 136-impl scan); the facts/cache pairing is now unconstructible-to-mismatch (all 16 routes are InferenceContext methods; free fn deleted); semantics byte-identical vs pre-#4461 with dependency registration proven not lost on memo hits; no RefCell re-entrancy; diff contains exactly the mandated items. The F1 mechanism (non-tracked `type_alias_value` full re-lowering per candidate) is confirmed real and this fix removes it.

**Pre-merge fix (N1)**: `impls.rs:269-271` — the new comment says cached rows "widen the re-entrancy window", which is backwards (memoization *reduces* re-entrant lookups; salsa can't mutate mid-query). The real constraint: a memo must not outlive one query execution — it would serve rows from a stale revision and suppress the dependency reads a later execution needs. A future reader following the wrong rationale could delete the constraint and reintroduce F1.

Non-blocking notes: the A/B is directionally sound (−2.2%/−2.4%, arithmetic checks out, divan flags real) but 2-arm/fixed-order/n=5 with overlapping ranges — recovery is established by mechanism, not by measurement power; a third arm at the pre-#4461 base would have closed it. N3-N6 (delegation nit, redundant db field, no cheap perf guard exists for this class, whole-self borrow trade) recorded for the backlog.

PR #4464, opened 2026-08-17
Add get_field method for runtime values

antoniosarosi:
## Summary

- Add `unknown.get_field<T>(name)` as the method-call spelling for typed reads from runtime-shaped class values.
- Delegate the method to the existing `reflect.class.get_field<T>(value, name)` implementation, preserving lookup, type checking, and thrown diagnostics.
- Keep the existing free-function spelling, add completions and keyword documentation for the method form, and record the language-surface addition in the changelog.

## Feasibility and scope

The stdlib-only rung is not sufficient because builtin methods are reached through a receiver companion mapping, and `unknown` previously had no companion. This uses the contained compiler-assist rung: the two existing companion lookup paths map only `TyKind::Unknown` to `baml.Unknown`.

Declared members win by construction, not by tested candidate ordering: typed class receivers resolve against their own class and never consult the `baml.Unknown` companion. A typed class without `get_field` therefore remains an E0007 unresolved-member error.

`baml.Unknown` lives after `TaggedString` so adding it does not shift the source lines of existing core items. The remaining snapshot updates record the new companion, its final source location, and the cross-referenced free-function documentation.

## Validation

- Runtime class/composite suite: 10/10 passed, including missing-field and wrong-`T` diagnostics through both spellings plus the typed-class E0007 containment guard.
- Unknown-member completion fixture and rendered keyword-topic assertions passed.
- Stdlib snapshot suite: 5/5 passed.
- Full pinned gate: 3,751/3,751 passed; 24 skipped; doctests passed; no unreferenced or pending snapshots.
- Commit hooks: formatting, YAML validation, and workspace-wide Clippy passed.

## Summary by CodeRabbit

- **New Features**
  - Added checked `get_field<T>` access for reading named fields from runtime-shaped `unknown` values.
  - Added support for both method and function-style field access.
  - Added editor completions for `get_field` on `unknown` values.

- **Bug Fixes**
  - Improved diagnostics for missing fields, invalid types, and unsupported field access.

- **Documentation**
  - Documented checked field-read syntax and equivalent access forms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. See [conven
 
 ## [0.225.0](https://github.com/boundaryml/baml/compare/0.224.0..0.225.0) - 2026-07-31
 
+### Features
+
+- add `get_field<T>` method calls for runtime-shaped `unknown` values
+
 ### Docs
 
 - brand legacy releases as BAML v0 (#4297) - ([5dbd250](https://github.com/boundaryml/baml/commit/5dbd25084b39523854a8967da5f5363b7368f3a6)) - Sam Lijin
--- a/baml_language/crates/baml_builtins2/baml_std/baml/core.baml
+++ b/baml_language/crates/baml_builtins2/baml_std/baml/core.baml
@@ -4,14 +4,6 @@ function deep_copy<T>(value: T) -> T throws never {
     $rust_function
 }
 
-/// Builtin companion for operations available on dynamically shaped values.
-class Unknown {
-  /// Assert-read a named field from a runtime-shaped class instance.
-  function get_field<T>(self, name: string) -> T throws baml.reflect.errors.CompilationError {
-    reflect.class.get_field<T>(self, name)
-  }
-}
-
 /// Implemented by all concrete types.
 /// Can be used to check or bound a type to only include concrete types.
 interface Concrete {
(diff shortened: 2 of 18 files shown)


antoniosarosi:
# Adversarial review — get_field method ergonomics

**Verdict: SHIP-WITH-FIXES.** The mechanism (a `baml.Unknown` companion class + two receiver-table rows, parallel to `baml.Int`/`baml.String`) is MORE contained than the brief's anticipated compiler-fallback: verified with a branch-built CLI that typed receivers still get E0007 for unknown members (including `get_field` itself on a class lacking it) — the universal method is structurally unreachable for typed receivers. Semantics parity proven on the 4-way matrix (missing-field / wrong-T × both spellings, identical E0001s); method-form turbofish incl. `unreflect(t)` mint identity works; `baml.Unknown{}` misuse yields catchable errors consistent with `baml.Int` precedents; targeted suites green.

## Pre-merge fixes
1. **F1**: the shadowing test is a tautology (its receiver is a typed class — the new arm is never consulted; the test passes with the feature reverted). Replace with a diff-sensitive assertion (typed receiver LACKING get_field still E0007s) and state in the PR body that member-shadowing holds BY CONSTRUCTION (companion-class mechanism), not by tested ordering.
2. **F2**: negative coverage must run through BOTH spellings (currently missing-field only-old, wrong-T only-new) — two extra assertions.
3. **F3**: CHANGELOG.md Features line (language-surface addition) + conventional-commit type on the commit.
4. **F4**: discoverability — add `Ty::BuiltinUnknown` to `completions_for_ty_members` (one line) and mention `get_field` in the `unknown` keyword-docs topic (describe-command snapshot will update).
5. **F5**: move `class Unknown` after `TaggedString` in core.baml to avoid the two line-number-only snapshot dirties.
6. Cheap comments: cross-reference the two new table rows to their sibling tables (F6), "users never construct this class" + primitive-spelling cross-ref docstrings (F7).

## Recorded, not blocking
- F8: watch the Size Gate (stdlib bytecode grew; baselines re-baked recently with tens-of-KB margins).
- F9 (pre-existing, ticket): `unknown` receivers in CALLEE position suppress the unresolved-member diagnostic (`infer.rs:5470-5473`) contradicting the field road's explicit rule — `u.no_such_method()` compiles silently on canary today; this PR makes such spellings routine, raising the priority.
- Design ratification note for the maintainer: `baml.Unknown` is now the permanent extension point for unknown-receiver methods — its first real member table. Consistent with the Int/String companions; flagged for explicit sign-off rather than STOP since the mechanism is strictly more conservative than the pre-ratified fallback rung.

PR #4467, opened 2026-08-17
fix: preserve mutable identities across loop calls

antoniosarosi:
## Summary

- preserve mutable map identity when a map allocated before a loop is used by a callee inside the loop
- detect CFG paths that revisit the map-allocation use without re-executing its definition, while retaining the existing loop-header guard
- add regressions for both the formerly failing no-caller-read shape and the formerly masking caller-read control
- add a changelog entry

## Root cause

Cross-block virtualization only rejected uses that were themselves loop headers. In the common `header -> body(use) -> header` shape, an empty map allocated before the loop was sunk into the body, so each iteration passed a fresh map and callee mutations did not accumulate.

The new check is intentionally limited to map allocations. An earlier broad CFG guard passed all runtime tests but changed seven unrelated string/copy/closure bytecode snapshots; those pending snapshots were rejected and are not part of this PR. The final implementation keeps all existing snapshots byte-identical.

## Verification

- focused regressions: 2/2 passed
  - no caller read: `"1,2,3,"`
  - caller read control: `"1,2,3, caller=3"`
- all eight standalone repros under `thoughts/antonio/repros/map-aliasing/` produced their expected output; m1-m4 and the sequential-call negative control remained unchanged
- exact pinned gate:
  - `CARGO_BUILD_JOBS=8 CARGO_INCREMENTAL=0 rustup run 1.93.0 cargo insta test --test-runner nextest -p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface --all-features --unreferenced=reject`
  - 3,753/3,753 passed, 24 skipped
  - no unreferenced snapshots; no snapshots to review
- pre-commit workspace clippy/format checks passed

## Release A/B benchmark audit

Patched / canary baseline medians:

- class method call 100k: 29.46 ms / 31.30 ms
- array build + sum 100k: 79.46 ms / 80.66 ms
- call chain 100 x 10k: 112.7 ms / 104.8 ms
- pure call 1m: 151.0 ms / 149.5 ms
- shared array, 1 writer + 4 readers: 23.23 ms / 23.16 ms
- shared array push, 4 x 25k: 22.46 ms / 12.06 ms (highly noisy)

There was no systematic regression: the isolated call-chain delta and concurrent-push result were noisy, while the other workloads were flat or faster.

## Scope audit

The PR contains exactly three files: emitter analysis, two regression tests, and the changelog. It contains no TypeScript bridge rebuild output.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where mutable maps, lists, and class instances could be recreated during loop iterations, causing unexpected state changes when modified by called functions.
  - Preserved object identity across loops and nested data structures, including class fields and returned values.
  - Ensured fresh allocations created inside loops continue to behave correctly.

- **Tests**
  - Added comprehensive coverage for mutation, aliasing, control flow, and optimized execution scenarios.

### DIFF
--- a/baml_language/CHANGELOG.md
+++ b/baml_language/CHANGELOG.md
@@ -26,6 +26,7 @@ This changelog covers the independent `baml_language` release line. It does not 
 
 ### Fixes
 
+- Fixed a silent miscompile where a map, list, or class instance allocated before a loop and mutated by a callee called inside that loop was re-allocated on every iteration. ([#4467](https://github.com/BoundaryML/baml/pull/4467)) - Antonio Sarosi
 - Restored E0007 diagnostics for method calls on `unknown` receivers, preventing unresolved calls from reaching an internal VM error. ([#4466](https://github.com/BoundaryML/baml/pull/4466)) - Antonio Sarosi
 - Rejected runtime-checked arguments on indirect calls with E0010; release builds had previously compiled them while silently omitting the check. ([#4460](https://github.com/BoundaryML/baml/pull/4460)) - Antonio Sarosi
 - Enabled runtime package compilation from `baml run` scripts and expressions. ([#4451](https://github.com/BoundaryML/baml/pull/4451)) - Antonio Sarosi
--- a/baml_language/crates/baml_compiler2_emit/src/analysis.rs
+++ b/baml_language/crates/baml_compiler2_emit/src/analysis.rs
@@ -1487,17 +1487,25 @@ fn can_be_virtual(
             }
         }
 
-        // Check if the use block is a loop header (has back-edge predecessors)
-        // If so, be conservative and don't inline
+        // Preserve the existing protection for values used directly in a loop
+        // header.
         let use_preds = predecessors
             .get(&use_loc.block)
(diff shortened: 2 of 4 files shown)


antoniosarosi:
Fix round pushed in `9f8902fcf`.

- Generalized the loop repeat-use guard with an exhaustive identity-allocation predicate for maps, arrays, byte arrays, class aggregates, closures, and bound methods.
- Added runtime-output coverage for the ReturnPhi CFG, the inside-loop freshness control, list/string/nested-list mutations, direct class mutation, class-held maps/lists, and O2.
- Added hand-built MIR assertions for both repeated (`Real`) and non-repeated (`Virtual`) allocations.
- Updated the existing lambda bytecode snapshot to pin its closure allocation before the loop.
- Moved and broadened the changelog entry under `baml_language/CHANGELOG.md`.

Validation: full pinned Rust 1.93 gate passed 3,762/3,762 tests with 24 skipped; all doctests passed; zero unreferenced snapshots; zero snapshots to review. The commit hook also passed workspace Clippy. No enqueue action taken.

PR #4470, opened 2026-08-17
fix: reject non-data LLM output schemas

antoniosarosi:
## Summary

- reject non-data realized types at the shared LLM render boundary with catchable E0164 diagnostics
- report nested failures with their full class-field path and ignore fields marked `@skip`
- reject realized generic classes before the current unsubstituted renderer can silently erase their schema and proceed to provider I/O
- replace the low-level `unreachable!` fallback with a fallible formatter error

## Root cause

`type._validate_renderable` guarded open interfaces, conflicting runtime definitions, and empty enums, but did not recursively reject non-data leaves. Those leaves reached `sys_ops::output_format`; `never` and its family aborted the process, while the first fix made unsubstituted generic-class fields return an error that throws-never helper callers swallowed into an empty schema.

## Compatibility / strictness

This intentionally makes previously degraded renders fail loudly with E0164: examples include a `uint8array` field that rendered as `RESULT=[]`, a `type` field, and a `string | uint8array` field that rendered only the string branch. Skipped fields remain ignored, matching the renderer. Generic functions whose realized outputs are non-generic remain supported.

## Deliberate deferrals

- Real substitution of realized generic class arguments into `ClassDefinition.field_type` in `sys_ops::output_format` is a separate migration follow-up. Until then, generic classes fail safely instead of aborting or rendering without a schema.
- Static rejection is intentionally not implemented: this is the runtime-only shared seam, matching E0159/E0161/E0162 precedent.
- The throws-never low-level output-format helpers are not routed through `_validate_renderable`; their comment is narrowed because they may still degrade an unsupported formatter error to an empty string.

## Tests

- root and nested non-data diagnostics, including `unknown` and `type`
- runtime-minted nested class definitions
- skipped non-data and open-interface fields
- generic-class rejection before provider I/O
- generic-function concrete-output positive control
- existing E0159/E0161/E0162 precedence coverage
- full pinned gate: 3,761/3,761 passed, 24 skipped; doctests clean; zero unreferenced or pending snapshots

## Summary by CodeRabbit

* **Bug Fixes**
  * LLM output schemas with unsupported non-data types now return catchable `E0164` diagnostics instead of crashing or being silently ignored.
  * Improved validation across nested types, unions, aliases, generic classes, and skipped fields.
  * Rendering now reports clear errors for unsupported types, including `never`, `uint8array`, and `type`.
  * Valid data-class fields continue to render successfully.

* **Documentation**
  * Added changelog coverage for the updated diagnostic behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,6 @@ All notable changes to this project will be documented in this file. See [conven
 
 ## [0.226.0](https://github.com/boundaryml/baml/compare/0.225.0..0.226.0) - 2026-08-17
 
-### Bug Fixes
-
-- return a catchable diagnostic when non-data LLM output types reach schema rendering
-
 ### Features
 
 - **(runtime)** make connection pooling configurable ([#3975](https://github.com/BoundaryML/baml/pull/3975)) - ([521919c](https://github.com/BoundaryML/baml/commit/521919c7f58b0b930ae1709df1882651db8ca865)) - Dex Hunter
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. See [conven
 
 ## [0.226.0](https://github.com/boundaryml/baml/compare/0.225.0..0.226.0) - 2026-08-17
 
+### Bug Fixes
+
+- return a catchable diagnostic when non-data LLM output types reach schema rendering
+
 ### Features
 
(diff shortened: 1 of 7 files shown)


antoniosarosi:
Applied the single adversarial-review fix round in `ad62eccc8`.

The guard now matches renderer `@skip` behavior (including the pre-existing E0161 hole), nested E0164 failures name the full field path, and generic classes fail before unsubstituted fields can silently erase the schema and reach provider I/O. The language changelog now discloses the strictness increase; generic-class substitution and static rejection are explicitly deferred.

Full pinned gate: **3,761/3,761 passed**, 24 skipped; doctests clean; no unreferenced or pending snapshots.

PR #4473, opened 2026-08-17
fix: diagnose unspecialized reflected generics

antoniosarosi:
## Summary

- diagnose genuinely unspecialized reflected generics with dedicated E0165 diagnostics
- preserve extraction and listing for reconstructable generated companions of generic LLM functions
- declare the reachable `baml.reflect.errors.CompilationError` channel on `reflect.call_any`
- document the current reflection limit without claiming that a specialization API exists

## Root cause and review correction

`Package.get_function` asks the VM to reconstruct a callable signature. For an unspecialized generic original, reconstruction returns `None`; the old code silently returned `Ok(None)`. The contract-mismatch path applies only when a signature was successfully reconstructed but does not satisfy the requested `AnyFunction` pins.

The first fix incorrectly gated on whether a function declared generic parameters. Generated `$render_prompt`, `$parse`, and `$stream` companions inherit their source function's generic parameters but reconstruct successfully with an empty supplied frame. This round instead diagnoses only an actual signature-reconstruction failure and uses the VM helper solely to distinguish an incomplete generic callable from a genuinely non-callable value. Reconstructable companions remain extractable and listed.

The diagnostic is now E0165:

> generic function `{name}` cannot be extracted through reflection: reflected packages cannot supply type arguments yet

`Package.get_function` intentionally supplies the package-qualified `display_local_name`; `reflect.call_any` has no package context and supplies the bare declared name.

No specialization API or compiler inference behavior is added. That design remains explicitly reserved for the human. `reflect.signature` still uses its older non-callable error for an unspecialized generic; correcting that requires a throws-channel/API decision and is deliberately deferred.

## Validation

- generic LLM companion extraction and package-listing regressions
- exact E0165 code/message and qualified-versus-bare naming coverage
- pinned `AnyFunction<Returns = string, Throws = never>` control
- exhaustive `call_any` catch updated for the declared `CompilationError` channel
- compiler diagnostics, runtime package, `call_any`, CLI describe, and type-kind coverage
- the package-listing snapshot's unrelated delta is only a one-line source shift from the added stdlib documentation line
- full pinned gate before the final canary rebase: 3,759/3,759 passed, 24 skipped; all doctests passed; zero unreferenced or pending snapshots
- post-rebase validation on canary `9d24fbaeb`: formatting passed; 38 diagnostics/CLI/type-kind tests passed; all 40 `reflect_call_any` and `runtime_package_compile` tests passed (one pre-existing skip)

## Summary by CodeRabbit

* **Bug Fixes**
  * Reflection now reports a clear compilation diagnostic when unspecialized generic functions are retrieved or invoked.
  * Unspecialized generic functions are excluded from reflected function listings until explicitly specialized.
  * Dynamic invocation error reporting now includes compilation errors alongside argument errors.
  * Reflected function descriptions now include available documentation.
* **Documentation**
  * Added guidance explaining generic function specialization requirements and reflected function availability.


antoniosarosi:
Adversarial review round pushed in `b268f661a` (rebased onto current canary). The `get_function` decision now keys on actual callable-signature reconstruction failure, so generic LLM companions that inherit type parameters but reconstruct with an empty frame remain extractable and listed. The round also assigns dedicated E0165, declares `CompilationError` on `reflect.call_any`, updates the exact diagnostics/docs, and adds companion extraction/listing plus exhaustive throws-channel regressions. Full pinned gate before the final canary rebase: 3,759/3,759 passed, 24 skipped, doctests clean, zero pending/unreferenced snapshots. Post-rebase: formatting, 38 diagnostics/CLI/type-kind tests, and all 40 reflection runtime tests passed (one pre-existing skip). `reflect.signature` is intentionally deferred because changing its current error requires a throws-channel/API design decision.

PR #4479, opened 2026-08-17
fix(ci): sccache log path

harivansh-afk:
Two small changes that cost nothing on your runners and unbreak CI for anyone running this repo's workflows outside it.

**sccache log path.** `.envrc` wrote the sccache error log to a hardcoded `/tmp`, which fails on any runner with an isolated or read-only `/tmp`. It now respects `TMPDIR` and behaves exactly as before everywhere else.

**benchmarks queue-wedge.** `benchmarks build (baml)` targets an ARM runner label that only exists in this repository. On a fork the job queues for 24 hours holding the whole run open, which also blocks re-running any other job in that run. It is now gated on `github.repository`.

Found while running this repo's full CI end to end on our own runner fleet (context in the PRs that follow). Both fixes have been live on the fork since 2026-08-16.

## Summary by CodeRabbit

* **Chores**
  * Updated development environment setup to support explicit Nix configuration.
  * Preserved temporary-directory handling for build cache error logs, with `/tmp` used as a fallback.


sxlijin:
I prepared the requested follow-up, but GitHub would not accept the maintainer edit to this fork branch. The remote head is still `a14fde0d8d8e5ea60f525909f7e4c7550ac9fe29`; pushing the validated commit to `indexable-inc/baml:up/ci-hygiene` failed with:

```
remote: Permission to indexable-inc/baml.git denied to sxlijin.
fatal: unable to access 'https://github.com/indexable-inc/baml.git/': The requested URL returned error: 403
```

The prepared result does exactly these three things:

1. Reverts this PR's `.github/workflows/ci.yaml` change completely, restoring `if: needs.determine_changes.outputs.run_perf == 'true'`.
2. Makes Nix explicit in `.envrc`:
   ```sh
   # Use nix only when explicitly requested; otherwise, use mise and brew.
   if [ "${BAML_SETUP:-}" = "nix" ] && command -v nix-shell >/dev/null 2>&1; then
   ```
3. Replaces the TMPDIR explanation with one line:
   ```sh
   # sccache needs a writable log file; some CI environments use a non-standard TMPDIR (for example, /var/folders/.../T/ on macOS).
   export SCCACHE_ERROR_LOG="${TMPDIR:-/tmp}/sccache-baml.log"
   ```

I validated the prepared patch with `bash -n .envrc`, `git diff --check`, and a fake-`nix-shell` behavior probe: unset `BAML_SETUP` takes the non-Nix path, while `BAML_SETUP=nix` takes the Nix path. A contributor with write access to the fork will need to apply/push it because maintainer edits are not working despite `maintainer_can_modify: true`.

harivansh-afk:
Applied your prepared patch verbatim, all three parts (5fc869533): the ci.yaml gate change is fully reverted (file now byte-identical to canary), .envrc uses nix only under `BAML_SETUP=nix`, and the TMPDIR comment is your one-liner. Re-ran your validation on the pushed tree: `bash -n .envrc`, `git diff --check`, and the fake-nix-shell probe (unset BAML_SETUP takes the mise path; BAML_SETUP=nix takes the nix path).

PR #4480, opened 2026-08-17
test: make wall-clock-bound tests robust under CI load

harivansh-afk:
Two families of tests encoded idle-machine timing as correctness, and both flake under real CI load with no product defect present.

- The sdk cancellation fixtures (java, rust, typescript) asserted that cancellation completes inside 0.5s while the operations being cancelled slept only 2s. Under contention an otherwise-green run measured 3.7s for the cancel path. The bound is now 5s and the cancelled sleeps 60s - so the bound has headroom, and a regression that *ignored* cancellation can no longer finish inside it and pass by accident.
- The bex_events prof soak test bounded each round's flush ack at 1s. Under a full CI fan-out one round measured >15s natively, and under Miri the 1s bound failed deterministically on some hosts. Each round is now bounded at 30s native / 1 min Miri: a wedged consumer still fails fast with the named panic (64 x 30s = 32 min, inside the job timeout), but machine speed is no longer part of the assertion.

We hit both while running this suite at high parallelism on 64-core machines; the numbers above are from those runs. The fixes make the tests honest on any hardware, including yours.

## Summary by CodeRabbit

* **Tests**
  * Improved cancellation test reliability across Java, Rust, and TypeScript SDKs by allowing additional time for cancellation and timeout operations.
  * Enhanced long-running soak-test validation with environment-appropriate acknowledgment timeouts.
  * These updates reduce timing-related test failures, particularly in slower or instrumented environments.

### DIFF
--- a/baml_language/crates/bex_events/src/prof/consumer.rs
+++ b/baml_language/crates/bex_events/src/prof/consumer.rs
@@ -1015,6 +1015,22 @@ mod tests {
         const ENGINE: u64 = 0x50AC_0001;
         let rounds: u64 = if cfg!(miri) { 4 } else { 64 };
         let per_round: u64 = if cfg!(miri) { 20 } else { 500 };
+        // Per-round wedge bound, not a latency assertion. Under Miri's
+        // interpreter a 1 s wall-clock bound is machine-marginal and failed
+        // deterministically on some hosts, so allow a minute there. Natively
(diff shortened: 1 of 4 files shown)


harivansh-afk:
Folded into #4483: these bounds exist because the ix preview runs the suite at higher parallelism than 16-vCPU runners, so the change travels with the machines that need it (disclosed in that PR's body) instead of asking for standalone review.

PR #4481, opened 2026-08-17
chore(ci): make attw faster, skip opaque binary blob

harivansh-afk:
The attw gate packs the working tree verbatim, so the tarball it checks carries the debug `.node` binary: **630MB, 99.7% of the archive**, gzipped on every CI run and never loaded by attw (it type-checks the JS/DTS surface only).

The interesting part: your release workflow already runs `rm -f dist/*.node` before publishing - native binaries ship as per-platform sub-packages - so the gate was validating a tarball 630MB larger than anything a consumer can actually install.

This packs from a staging directory that excludes `*.node`, with an otherwise identical file set and identical attw argv. Verdicts are byte-identical on the current tree, and the check now matches the published shape instead of accidentally diverging from it.

Measured: **61.0s → 5.1s** for the attw step in CI (59.4s → 4.2s locally).

## Summary by CodeRabbit

* **Bug Fixes**
  * Package validation now runs against a temporary package copy, avoiding interference with concurrent tests and the source package.
  * Validation results and error reporting are preserved, with temporary files cleaned up automatically.
  * Native add-ons are excluded from the validation package to improve compatibility and ensure accurate checks.

* **Chores**
  * Updated package validation to use the local checker consistently across development and automated tests.


sxlijin (commented):
The staging wrapper itself is well-scoped and `node --check baml_language/sdks/typescript/bridge_typescript/typescript_src/attw-check.js` passes. There is one applicability gap against current canary, though: the generated TypeScript SDK harness still emits

```
pnpm exec attw --pack --profile esm-only
```

at `baml_language/sdk_tests/harness_setup/src/typescript.rs:205`. That invokes the dependency binary directly and bypasses the new `package.json` script, so the CI path this PR is intended to speed up does not use `attw-check.js`.

Please either change that command to `pnpm run attw` in this PR, or merge #4485 first (it already makes that exact harness change). With #4485 first, this PR remains independently useful and should merge cleanly; without it, this patch has no effect on the generated SDK-test attw invocation. There is no path or semantic conflict with #4483 or the already-merged #4499.

sxlijin:
Why this needs a wrapper instead of a flag:

`attw --pack` first runs `npm pack` over the package and then analyzes the resulting tarball. ATTW has `--ignore-rules` for suppressing diagnostic classes and `--exclude-entrypoints` for changing which package entrypoints it analyzes, but it has no file/glob option that removes an arbitrary payload from the tarball. In particular, neither option prevents npm from reading and gzip-compressing the 630 MiB debug `.node` file, which is the expensive part. The CLI option semantics are documented here: https://www.npmjs.com/package/@arethetypeswrong/cli#options

Changing the package's normal publish/include rules is also the wrong layer. The local debug build leaves the addon under `dist/`, while the release workflow runs `napi artifacts` and publishes native binaries in per-platform optional-dependency packages; the umbrella tarball ATTW is meant to model does not contain that local debug blob. The SDK harness also runs ATTW concurrently with fixture tests that load the real addon, so temporarily moving or deleting it would introduce a race.

That is why the script:

1. makes an isolated staging directory,
2. copies the publish-shaped `dist/` tree minus only `.node`,
3. copies the real `package.json` unchanged so exports/types are analyzed exactly,
4. invokes the ATTW JavaScript entrypoint directly so the same path works without shell/`.bin` shim differences on Windows, and
5. cleans up in `finally` while preserving ATTW's exit status.

The measured cost in the patch is 59.4s with the 630 MiB addon versus 4.2s without it, with byte-identical ATTW output; the complexity is isolating the packaging input safely, not changing the type analysis.

harivansh-afk:
Now self-contained (6b9f817b0): the generated harness invokes `pnpm run attw` at the call site you flagged, carried in this PR itself since #4485 is closed. No sequencing dependency on any other PR. It is also safe on trees without this PR's package.json change - the `attw` script already exists on canary with the old command, so the harness change is behavior-neutral there.

PR #4482, opened 2026-08-17
chore: speed up python sdk tests, build bridge once instead of twice

harivansh-afk:
The python sdk test setup builds the engine bridge twice, and the first build is 100% discarded.

`uv sync` performs a PEP 660 editable install, and editable builds inherited `tool.maturin.profile = release`: fat LTO, codegen-units = 1, the full engine graph compiled from scratch. Nothing ever consumes that artifact - `maturin develop` runs immediately after and replaces the extension with a dev-profile build. On a cold runner the discarded release build cost 199s of a 281s setup; even on a warm hosted runner the pair was 109.5s of a 112s step.

Fix: pin `editable-profile = "dev"` (and floor maturin at 1.10, where the key landed - older maturin ignores it and silently builds release again). Both steps now build the same profile the test run itself uses and share one set of artifacts. The artifact under test is unchanged; Windows already had the equivalent behavior.

Measured: setup **281.7s → 119.6s** in CI.

## Summary by CodeRabbit

- **Chores**
  - Improved the Python development setup to use a consistent editable development configuration.
  - Streamlined fixture setup so local development builds can reuse artifacts more efficiently.
  - Updated packaging configuration to improve compatibility with current Python build workflows.
  - Added documentation clarifying the development setup and build behavior.


sxlijin:
I ran a controlled before/after benchmark to verify the cache-sharing claim.

Method: both cases used the exact PR head (`8cd077fdd4`) on an Apple M3 Max with Rust 1.93.0, uv 0.11.3, and Maturin 1.14.1. The “without” worktree reversed only this PR's `pyproject.toml` setting/version-floor/comment. Each case had a fresh venv, uv cache, and `CARGO_TARGET_DIR`; sccache was disabled, `CARGO_INCREMENTAL=0`, and `CARGO_BUILD_JOBS=8`. I populated Python dependencies with `uv sync --group dev --no-install-project` before timing, then timed an editable `uv sync --group dev` followed immediately by `maturin develop -v`.

| Case | Editable `uv sync` | Following `maturin develop` | Combined | Cargo evidence |
| --- | ---: | ---: | ---: | --- |
| Without `editable-profile` | 270.24s | 110.99s | 381.23s | 400 compile lines in **release**, then 400 compile lines again in **dev** |
| With `editable-profile = "dev"` | 140.27s | 11.23s | 151.50s | 400 compile lines in **dev**, then only 8 Python-binding crates rebuilt; the engine graph was `Fresh` |

That is a 229.73s (60.3%) reduction for the two-command sequence. The without case left both a 1.5 GiB `target/release` tree and a 3.8 GiB `target/debug` tree. The with case produced only the 3.8 GiB debug tree. The raw Cargo summaries were:

- without: `Finished release ... in 4m 23s`, followed by `Finished dev ... in 1m 47s`
- with: `Finished dev ... in 2m 17s`, followed by `Finished dev ... in 8.09s`

The CI steps that are intended to share Cargo work are:

1. **Build tests with --timings** runs `cargo test --no-run ...`, which populates the workspace's dev/debug target before setup.
2. **Run sdk test setup** triggers `sdk_tests/crates/python_pydantic2/setup.sh` through nextest. Its first fixture `uv sync` performs the PEP 660 editable build of `baml_bridge`.
3. The same setup script finishes with one `maturin develop`, also against the shared workspace target.

Without this setting, Maturin makes the editable build inherit its ordinary wheel `profile`, which defaults to `release` here (fat LTO, one codegen unit, size optimization). Cargo keys artifacts by profile, so that `target/release` graph neither consumes the dev artifacts from **Build tests with --timings** nor helps the final dev-profile `maturin develop`. With `editable-profile = "dev"`, all three operations target the debug/dev graph. Normal wheel builds still use the existing release `profile`; only PEP 660 editable installs change.

sxlijin:
`editable-profile = "dev"` affects only PEP 660 editable builds, such as `uv sync` installing `baml_bridge` as editable. It is not used by `maturin build`, so our release workflow remains on its explicit `--profile release-bridge-python` build and needs no changes.

sxlijin:
We are moving this change to #4533 and will enqueue that upstream-owned PR instead. #4533 preserves the substantive commit from this PR (`beaf4248bfc993c4f8dc2818a824991b9a5b3c3d`) with @harivansh-afk as its GitHub-recognized author, while applying the requested comment cleanup.

Why we cannot update this branch directly:

- GitHub reports `maintainerCanModify: true` on this PR, and the authenticated maintainer has admin/push permission on `BoundaryML/baml`.
- A direct push to `indexable-inc/baml:up/python-bridge-once` fails with `403 Permission to indexable-inc/baml.git denied to sxlijin`.
- The GitHub App contents API fails independently with `403 Resource not accessible by integration`.
- Fetching the branch works, so this is specifically a write-authorization failure rather than a missing branch or stale-head problem.

The likely root cause is that `indexable-inc/baml` is an **organization-owned** fork. GitHub's maintainer-edit documentation scopes this mechanism to pull requests from forks in personal accounts and describes the checkbox as applying “on user-owned forks.” Our best reading is that the `maintainerCanModify` flag is present, but it does not grant a BoundaryML maintainer or GitHub App write access across the separate `indexable-inc` organization boundary; the writer would also need permission in that organization. GitHub documents that distinction here: https://docs.github.com/en/pull-requests/how-tos/work-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Because both supported write paths fail consistently, #4533 is the reviewable and queueable upstream replacement.

PR #4484, opened 2026-08-17
test(typescript-web): bound the vitest worker fan-out on big hosts

harivansh-afk:
Your #4462 fixed the sharp end of this: the workerd-spawning `vitest_workers` fixtures are now serialized, so only one workerd fleet exists at a time. This PR bounds the other axis - how big that fleet can get.

vitest sizes its worker pool at `cores - 1`. On the 16-vCPU runners this suite was tuned on that means 15 workers; on a 64-core host it means **63 workers, 63 workerd processes, each ~1 GiB anon under V8** - we measured the suite OOM-killing itself at 22 resident workerd (~13 GiB RSS + 14 GiB zram) and confirmed the kill chain with cgroup `oom_kill` A/B runs. The webview vitest configs now pin `maxWorkers`, so `#4462's serialization x this cap` = a bounded peak on any machine.

No effect at your current envelope (the caps sit at or above what 16 vCPUs produce naturally); it makes the suite portable to anything bigger.

## Summary by CodeRabbit

- **Build & Release**
  - Improved development and CI environments for more consistent Rust builds.
  - Added support for cross-compiling applications for macOS, Windows, and Linux.
  - Improved musl-based builds by reusing already available compiler tools.

- **CI & Infrastructure**
  - Added automated management and health reconciliation for self-hosted build runners.
  - Enhanced CI fallback behavior across different runner environments.

- **Tests**
  - Improved browser test startup and dependency handling for more reliable test execution.

### DIFF
--- a/typescript2/app-vscode-webview/vitest.config.ts
+++ b/typescript2/app-vscode-webview/vitest.config.ts
@@ -1,52 +1,70 @@
-import { defineConfig } from 'vitest/config';
-import react from '@vitejs/plugin-react';
 import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
+import react from '@vitejs/plugin-react';
 import { playwright } from '@vitest/browser-playwright';
+import { defineConfig } from 'vitest/config';
 
 const projectRoot = dirname(fileURLToPath(import.meta.url));
 
 export default defineConfig({
+  define: {
+    __DEV__: true,
+  },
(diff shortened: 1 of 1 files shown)


harivansh-afk:
Folded into #4483 for the same reason as #4480: the worker caps are a no-op at the 16-vCPU envelope and a requirement on the preview's 64-core machines, so they ship with the preview.

PR #4486, opened 2026-08-17
[4/5] ci: run the Linux lanes on the ix pool, hosted fallback intact

harivansh-afk:
> **Stack 4/5** &middot; parent: #4485 &middot; [this PR's own diff](https://github.com/indexable-inc/baml/compare/up/hermeticity...up/ix-lanes) (GitHub shows ancestor commits below until the parent merges, then this diff collapses to its own).

The flip: `runs-on` swaps for every Linux lane. Same-repo events target `[self-hosted, ix-<family>]`; **fork PRs keep the exact Blacksmith runner each job uses today** - untrusted code never touches persistent VMs that later run trusted, secret-bearing jobs. (That fallback is also why this PR's own CI runs on Blacksmith: it comes from a fork. The same contents running on the pool: [the fork's trunk waves](https://github.com/indexable-inc/baml/actions/workflows/ci.yaml?query=branch%3Acanary).)

What the pool buys, measured against this repo's own Blacksmith medians (6 green runs):

| job | Blacksmith median | ix, warm |
|---|---|---|
| size-gate (linux) | 310s | 30s |
| cargo doc | 173s | 24s |
| Playwright | 182s | 57s |
| webview typecheck | 185s | 65s |
| cargo test (gnu) | 198s | 92s |
| snapshot tests | 470s | 395s |

Also in here: jobs enter the nix toolchain shell via the `setup-ci-shell` seam from #4483 (fallback arm unchanged for hosted runners), `checkout` moves to `clean: false` so ignored build state survives on persistent runners (a no-op on ephemeral ones), an `ix-cross.yml` lane adds zig/xwin windows + darwin compile checks that hosted runners had no capacity for, and lane-level parallelism pins keep tools honest on 64 cores (they size to `nproc` otherwise).

Reverting this one PR - or flipping any job's `runs-on` back - restores today's behavior exactly. The pool machinery itself carries no state your CI depends on.

## Summary by CodeRabbit

- **New Features**
  - Added support for cross-compiling CLI releases for macOS and Windows across multiple architectures.
  - Centralized release platform definitions for consistent packaging across supported SDKs and artifacts.
  - Added configurable test fixture and source locations for SDK and language tooling.

- **Bug Fixes**
  - Improved Windows archive extraction by safely overwriting existing files.
  - Improved generated test handling with clearer errors and fallback output locations.
  - Prevented unsupported fork jobs from waiting on unavailable runners.

PR #4487, opened 2026-08-17
[5/5] ci: compile the heavy Rust lanes through nix, cached on ix's binary cache

harivansh-afk:
> **Stack 5/5** &middot; parent: #4486 &middot; [this PR's own diff](https://github.com/indexable-inc/baml/compare/up/ix-lanes...up/l2-lanes) (GitHub shows ancestor commits below until the parent merges, then this diff collapses to its own).

The heavy Rust lanes (msrv, wasm, gnu, musl) get a second compile path: a per-rustc-unit nix derivation graph, substituted from ix's binary cache, where a trusted builder publishes the graph for every trunk rev. A PR then compiles only its own delta - changed units and their reverse-dependencies - instead of the whole workspace. Think per-crate sccache, except the cache key is the full input closure, so it can never serve a stale hit.

The design constraint throughout is **arm equivalence**: both arms compile the same program (same profile, same `debug_assertions`; only debuginfo is dialed down, and the gnu/musl test lanes run real binaries so even that is scoped to lanes that never execute what they build... the msrv lane is `--no-run` by construction). Where equivalence could not be proven, the lane keeps cargo - the probe **fails open**: no published graph for a rev, no nix on the runner, any error at all → the job takes the exact cargo path it takes today. There is no state where this PR makes a lane worse than the status quo.

Measured on the worst lane on the board - msrv, whose unique toolchain nothing else warms: **cold start 224s vs your 228s warm Blacksmith median**. A freshly rebuilt pool recovers to warm speed on its first wave instead of after days of cache accretion.

Requires nothing from you beyond the stack below it; cache population for this repo's revs is on our side.


harivansh-afk:
Superseded by #4483, which carries the compile-cache machinery inside the non-gating preview workflow instead of the gating lanes.

PR #4491, opened 2026-08-18
feat(reflect): add read-only AnyClass reflection

antoniosarosi:
## Summary

Adds read-only value-side class reflection through `baml.AnyClass`, with runtime narrowing for both static and runtime-minted class instances. Also activates the existing E0113 diagnostic so a fall-through `let ... else` branch is rejected instead of reaching the MIR-inserted `unreachable` at runtime.

The implementation follows the [feasibility report](#feasibility-report) and keeps the surface read-only.

## Ratified rulings

1. Of the nine sealed reflection kind views, only `baml.reflect.class.Type` inhabits `baml.AnyClass`; the other eight do not.
2. This is read-only: there is no `set`, `set_field`, write native, or mutation API.
3. The interface lives at `baml.AnyClass`, alongside `baml.AnyFunction`.
4. Field handles reuse `baml.reflect.class.Field`, and metadata reuses `baml.reflect.Meta`.

## Implementation

- Defines the seven AnyClass operations as pure-stdlib interface defaults backed by the existing reflection natives, plus a blanket implementation.
- Adds compiler-derived class-only membership and matching compiler/runtime impl guards, including runtime-minted classes and `reflect.class.Type`.
- Applies that class-only membership oracle consistently to requires-closure validation, recursive compiler bound discharge, and recursive VM proof. Ordinary dispatch is unchanged.
- Keeps AnyClass's blanket implementation out of ordinary concrete-member lookup and inference-variable impl probing. This prevents inherited names such as `get` from shadowing map methods or user-defined generic methods.
- Scans type-side field metadata in `has_field` and `get_field`, so `get` no longer allocates one bound field handle per class field and `get_field` mints at most the single matching handle.
- Makes `reflect.class.Field` value-bound when obtained from an instance; type-side fields retain a null owner. The exact field-handle spelling is `read<T>()` and `metadata()`: `get` and `attributes` would also compile, but the distinct names avoid giving one name two different meanings on a Field value. The brief explicitly allowed adapting these signatures to the proven implementation shape.
- Filters describe output through the same class-membership rule, so enums and the eight excluded kind views do not advertise AnyClass.
- Wires the existing E0113 factory when a `let ... else` branch can fall through. This is isolated in the second commit.
- Snapshots and restores flow narrowing while inferring a diverging `let ... else` branch, so assignments on the failure path cannot affect the reachable success continuation; diagnostics and throw effects remain accumulated.

## Coverage

- Runtime-minted Vetrec-shaped class: narrow once, then read five consecutive optional-string fields.
- Every AnyClass method on static and runtime-minted instances.
- Missing field versus present-null behavior, and catchable generic type mismatch.
- Static classes and `reflect.class.Type` accepted; primitives, enums, arrays, maps, and the other eight reflection kind views rejected.
- Reflection parity, static metadata, and ownerless type-side field handles.
- User/inherent methods colliding with all seven AnyClass method names.
- Generic out-of-body `get` lookup through an inferred receiver.
- E0125 rejects a primitive in an AnyClass requires closure; compiler and VM bounded-impl membership agree for primitive and real-class cases.
- Describe output for eligible and excluded builtins.
- E0113 for a non-diverging branch, with `return` accepted and the current loop-divergence limitation pinned.
- Post-let hover coverage proves that a literal assignment before `return` in the else branch does not narrow the outer local on the success path.

## Snapshot audit

- Stdlib PPIR/MIR/codegen snapshots add the AnyClass interface, seven default methods, its blanket impl, and the two Field methods.
- CLI builtin-list snapshot adds AnyClass and reflects the Field surface; class/user describe snapshots show the blanket only on eligible classes.
- The final review round updates only the three expected CLI source locations shifted by the expanded stdlib implementation; no snapshot metadata churn was accepted.
- TIR package snapshot adds AnyClass plus `Field.metadata`/`Field.read`.
- Optimized and unoptimized bytecode snapshots move deterministic callable indices by nine: seven AnyClass methods plus two Field methods. No unrelated bytecode shape changed.

## Feasibility report

Source assessment: `thoughts/antonio/anyclass-feasibility.md` (coordinator working-tree artifact, intentionally excluded from the product diff). It established that typed `let ... else` interface narrowing, generic virtual dispatch, and six of the seven read operations already worked in pure BAML. The implementation here supplies the report's missing class-only membership rule, mirrors it in runtime impl resolution, omits the deferred write path, and fixes the independently reproduced E0113 hole.

## Gate

```text
CARGO_BUILD_JOBS=8 CARGO_INCREMENTAL=0 rustup run 1.93.0 cargo insta test --test-runner nextest -p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface --all-features --unreferenced=reject
```

- 3,788/3,788 tests passed; 24 skipped.
- Doctests passed.
- No unreferenced snapshots, pending snapshots, or `*.snap.new` files.
- `git diff --check` passed.

No enqueue requested; this PR is ready for coordinator review.

## Summary by CodeRabbit

* **New Features**
  * Added `baml.AnyClass` for read-only class reflection, including field discovery, metadata, type information, and runtime field reads.
  * Added support for narrowing runtime-created class values and reporting typed reflection mismatches.
* **Bug Fixes**
  * `let ... else` branches now require divergence, with clearer diagnostics for invalid branches.
  * Improved type-flow handling so else-branch values do not incorrectly affect surrounding code.
* **Documentation**
  * Documented `AnyClass` reflection and narrowing capabilities in the changelog.


antoniosarosi:
Final adversarial-review round is pushed in `43d6d7670`.

Addressed the five scoped findings:

- E0125 requires-closure validation now uses the class-membership oracle for `AnyClass`.
- Compiler bound discharge applies the derived membership guard recursively in `resolve_within_depth`.
- VM membership applies the same guard recursively in `prove`.
- Removed the redundant derived-membership conjunct from impl enumeration.
- `AnyClass.has_field`/`get_field` now scan type-side fields; `get_field` mints at most the matching bound handle, avoiding the per-field allocation cliff in `get`.

Added negative primitive/bounded-impl cases and a positive real-class requires/dispatch case. Dispatch code is unchanged.

Full pinned gate: 3,787/3,787 passed, 24 skipped; doctests passed; no unreferenced or pending snapshots. Final audit found no `*.snap.new`, no bridge TypeScript paths in the commit, and a clean `git diff --check`.

antoniosarosi:
CodeRabbit micro-round is pushed in `de8e85802`.

Verified the finding independently before applying the proposed shape: the new regression failed on the prior head because an else-only assignment leaked and the post-let hover was `value: \"else-only\"` instead of the declared `value: int | string`.

`finish_let_else` now snapshots and restores `self.flow` around else-branch inference alongside its existing divergence isolation. Pending diagnostics and throw effects remain accumulated; the full existing throw/catch let-else fixtures pass.

Validation:

- Focused let-else fixtures: 19/19 passed.
- Full pinned gate: 3,788/3,788 passed, 24 skipped.
- Doctests passed; no unreferenced or pending snapshots; no `*.snap.new` files.
- Commit scope is two files only; `git diff --check` is clean and no bridge TypeScript paths are included.

PR #4493, opened 2026-08-18
feat(reflect): apply the AnyClass follow-up rulings

antoniosarosi:
## Summary

`x?.m<T>()` compiled clean and died at runtime with

```
VM internal error: could not realize type template:
template references frame type-arg slot 0 but the frame has 0 type args
```

while the equivalent `if let` / let-else spelling worked. Found during #4493 on
the `baml.AnyClass` surface (`value.get_field(name)?.read<T>()`).

The optional chain was dropping the call's type arguments. This lowers
`x?.m(...)` as the guarded *method call* it is, so `?.` decides **whether** the
call happens and never **how** it is made.

## Root cause

`baml_compiler2_mir::lower::lower_call` recognizes a method call by its callee
shape: `AstExpr::MemberAccess` (`x.m`) or `AstExpr::Path` (`x.m` written as a
path). An `AstExpr::OptionalMemberAccess` callee matched neither, so it fell
through to the final "the callee is an opaque callable value" branch
(`lower.rs:8504` on canary): `MakeBoundMethod` for the receiver, then
`Terminator::Call` with a non-constant callee.

`baml_compiler2_emit::emit` (`emit.rs:2311`) lowers a non-constant callee to
`Instruction::CallIndirect` — which has no `ntypeargs` field. So the `ntypeargs`
MIR computed was silently discarded. Verified by instrumenting `lower_call`:
the `?.` call site really does compute `ntypeargs=1` and hand it to an indirect
call. The `LoadType` operands were pushed and then stranded below the callee
frame (a leaked operand-stack slot per call), and the callee frame was seeded
only with whatever the `BoundMethod` had curried — the receiver's class-level
args, and nothing else.

That last detail explains the two shapes of the failure:

| receiver | frame the callee got | outcome |
| --- | --- | --- |
| plain class | `[]` | `slot 0 but the frame has 0 type args` |
| generic class | `[class T]` | `slot 1 but the frame has 1 type args` |

## Blast radius (measured on canary, before the fix)

Broken:

- `x?.m<T>()` — explicit method type args
- `x?.m(v)` — type args **inferred** at the call site
- `iface?.m<T>()` and `concrete?.m<T>()` where `m` comes from an `implements` block
- `Box<int>?.both<U>()` — class args present, method args missing (slot shift)
- `a?.b()?.c<T>()` — chained
- `value.get_field(name)?.read<T>()` — the reported shape

Already working, and still working:

- `x?.m()` with no type args anywhere
- `Box<int>?.describe()` — class args only (the `BoundMethod` curried these)
- a null receiver: the chain short-circuits and the callee never runs

The stdlib already carries a workaround for this bug class —
`baml_std/ai/ns_clients/clients.baml:219`: *"`?.method()` on an interface-typed
optional trips a VM bug; if-let is the reliable form."* Left in place here (out
of scope); it can be simplified as a follow-up.

## Fix

`lower_call` now diverts an `OptionalMemberAccess` callee to
`lower_optional_method_call`, which emits the null test (joining the enclosing
`OptionalChain`'s shared null exit, exactly like `lower_optional_call` does) and
then re-enters the ordinary call lowering with the callee viewed as a plain
`MemberAccess`. The callee's *expression id* is unchanged, so every TIR lookup
— resolution, call plan, receiver type — still keys on the node the type
checker recorded.

Past the guard the receiver is non-null, so the receiver type narrows
(`T | null` → `T`) for the three lookups that read it: interface dispatch, union
dispatch, and the receiver's class-level type-arg prefix. That is the same
narrowing `dispatch_target_for_member_access` already applied to `x?.field`;
`try_lower_interface_dispatch` now shares that helper outright. Without the
narrowing, dispatch declines on the `Class<..> | null` union and the class
prefix comes back empty, shifting every De Bruijn slot the method's own args
occupy.

Net effect: `x?.m<T>()` emits exactly what `x.m<T>()` emits — a direct
`Call`/`VirtualCall` with the type args leading — under a null branch. The
stranded operand-stack slot goes away with it.

## Tests

New `crates/baml_tests/baml_src/ns_optional_chain_type_args/` runtime-output
suite (10 `test` blocks), each pairing the optional spelling with the
non-optional one it must agree with: explicit type args vs. let-else, null
receiver short-circuit, inferred type args, class-only args, class + method args
(slot order), interface and concrete-receiver interface methods, the chained
form with both a present and a null receiver, a non-generic `?.` negative
control, and the `get_field(name)?.read<T>()` reflection shape — including that
a type argument the value does not fit is still a `baml.errors.TypeMismatch`,
not a frame-layout failure.

## Not in scope

Two adjacent pre-existing bugs, unchanged by this PR and not type-arg related:

1. **`?.` evaluates its receiver twice.** `get(c, "yes")?.raw` calls `get`
   twice — once for the null test, once for the access. True of `x?.field` on
   canary as well; this PR keeps the count at two.
2. **`x?.to_string()` ICEs** (`MIR failed to resolve field access .to_string
   against class definition ...`). The `to_string`/`to_json`/`from_json` sugar
   fallbacks match on `MemberAccess`/`Path` only (`is_sugar_callee`), so they
   never fire for an optional-chained receiver.

## Summary by CodeRabbit

* **New Features**
  * Infinite `while` loops without reachable `break` statements are now recognized as diverging, improving `let … else` flow analysis.
  * Added clearer diagnostics for attempts to construct invalid built-in companion types.

* **Breaking Changes**
  * Renamed reflection API `Field.read<T>()` to `Field.value<T>()`.

* **Bug Fixes**
  * Prevented invalid built-in companion class construction while preserving support for user-defined classes with similar names.


antoniosarosi:
Amendment round: the revised ruling 1, plus the review's must-fix and three follow-ups. Head is now `948320f99`.

### Ruling 1 revised — `metadata()` stays

`Field.metadata()` keeps its name and the `meta` storage field is exactly as #4491 shipped it. `read<T>()` → `value<T>()` stands, as does the recomposed `AnyClass.get<T>`.

Measured against `origin/canary`, the revert is exact: `class.baml` differs by **one line** (the `read` → `value` rename), and `type_kinds.rs`, `runtime_classes_and_composites.rs` and the VM's `Field` construction are byte-for-byte back to canary. Snapshots were **regenerated**, not hand-reverted — `package_items` lands on `methods: [metadata, value]`, and the describe listing's `class.baml` offsets return to canary's values (29/43/54/60/102/134/139) while `core.baml` keeps its +1 shift from the longer `get` body. A side benefit: `Field.metadata` lowers to a direct field read (`load_field .meta`) again instead of a method call.

### MUST-FIX — `break` in the C-style update slot

Fixed as prescribed. `loop_body_breaks` now takes `after: Option<StmtId>` and seeds the walk with it; the call site passes the outer `*after`, not the binding shadowed by the `if let` above it; `StmtId` joins the `flow.rs` imports.

`for (let i = 0; 1 == 1; i = break) {}` is now a **fourth rejection case** in `let_else_loop_with_break_does_not_diverge.baml`, and it produces E0113 with no parse errors — the update slot really does parse and really does hold a reachable `break`, which is exactly why missing it was unsound.

### Also fixed

1. **Changelog** — three entries under the top block: the new accepted divergence form (Features), `Field.read` → `value` (Breaking and compatibility changes), and the E0166 sealing plus the ICE it removes (Fixes). PR link and attribution per the #4470/#4473/#4491 convention.
2. **E0166 wording** — `baml.Bool` and `baml.Null` have entirely empty bodies (verified: 0 methods each, against 9–45 for the others), so telling them they "only carry the methods of `bool`" was false. `BuiltinCompanion` now reports `carries_methods`, and the shared factory picks *"it is only the companion of `bool`, whose values come from literals"* for those two. Everything else is unchanged.
3. **Divergence recognizer's reach** — kept as-is per the ruling, with the doc comment rewritten to state what it actually accepts: the literal, constant folds, flow-narrowed bindings, and calls declared `-> true`. It also records *why* narrowed bindings are sound — a loop havocs every binding its body assigns before the condition is checked, so a binding the loop could falsify is no longer narrowed at that point. `if (c is true) { while (c) {} }` is now an acceptance fixture, written as an `if`/`else` where the other arm throws, so the `while` is load-bearing rather than incidentally rescued.

### Gate

Full pinned gate, clean: **3,792/3,792 passed**, 24 skipped, doctests passed, `no unreferenced snapshots found`, no pending snapshots. Clippy clean; pre-commit hooks (fmt + clippy) passed on the commit.

PR #4495, opened 2026-08-18
fix: thread type arguments through optional-chained calls

antoniosarosi:
## Summary

`x?.m<T>()` compiled clean and died at runtime with

```
VM internal error: could not realize type template:
template references frame type-arg slot 0 but the frame has 0 type args
```

while the equivalent `if let` / let-else spelling worked. Found during #4493 on
the `baml.AnyClass` surface (`value.get_field(name)?.value<T>()`).

The optional chain was dropping the call's type arguments. This lowers
`x?.m(...)` as the guarded *method call* it is, so `?.` decides **whether** the
call happens and never **how** it is made.

## Root cause

`baml_compiler2_mir::lower::lower_call` recognizes a method call by its callee
shape: `AstExpr::MemberAccess` (`x.m`) or `AstExpr::Path` (`x.m` written as a
path). An `AstExpr::OptionalMemberAccess` callee matched neither, so it fell
through to the final "the callee is an opaque callable value" branch
(`lower.rs:8504` on canary): `MakeBoundMethod` for the receiver, then
`Terminator::Call` with a non-constant callee.

`baml_compiler2_emit::emit` (`emit.rs:2311`) lowers a non-constant callee to
`Instruction::CallIndirect` — which has no `ntypeargs` field. So the `ntypeargs`
MIR computed was silently discarded. Verified by instrumenting `lower_call`:
the `?.` call site really does compute `ntypeargs=1` and hand it to an indirect
call. The `LoadType` operands were pushed and then stranded below the callee
frame (a leaked operand-stack slot per call), and the callee frame was seeded
only with whatever the `BoundMethod` had curried — the receiver's class-level
args, and nothing else.

That last detail explains the two shapes of the failure:

| receiver | frame the callee got | outcome |
| --- | --- | --- |
| plain class | `[]` | `slot 0 but the frame has 0 type args` |
| generic class | `[class T]` | `slot 1 but the frame has 1 type args` |

## Blast radius (measured on canary, before the fix)

The decision was made purely from the callee's AST shape, before anything about
the surrounding expression was consulted — so this was a **whole-operator
outage for generics**, not a corner case. `?.` + a generic method failed
wherever it was written:

- receiver forms: plain class, generic class, interface-typed, concrete class
  with an `implements` method, a bounded type variable (`<T: Iface>`)
- call forms: explicit type args (`x?.m<T>()`), type args **inferred** from the
  arguments (`x?.m(v)`), chained (`a?.b()?.c<T>()`), the reported reflection
  shape (`value.get_field(name)?.value<T>()`)
- syntactic positions: statement position, `let` initializer, argument to
  another call, operand of a binary operator, inside a template-literal
  interpolation, inside a `for`/`while` body, inside a lambda body, wrapped in
  parentheses

Already working, and still working:

- `x?.m()` with no type args anywhere
- `Box<int>?.describe()` — class args only (the `BoundMethod` curried these)
- a null receiver: the chain short-circuits and the callee never runs

The stdlib already carries a workaround for this bug class —
`baml_std/ai/ns_clients/clients.baml:219`: *"`?.method()` on an interface-typed
optional trips a VM bug; if-let is the reliable form."* Left in place here (out
of scope); it can be simplified as a follow-up.

## Fix

`lower_call` now diverts an `OptionalMemberAccess` callee to
`lower_optional_method_call`, which emits the null test (joining the enclosing
`OptionalChain`'s shared null exit, exactly like `lower_optional_call` does) and
then re-enters the ordinary call lowering with the callee viewed as a plain
`MemberAccess`. The callee's *expression id* is unchanged, so every TIR lookup
— resolution, call plan, receiver type — still keys on the node the type
checker recorded.

Past the guard the receiver is non-null, so the receiver type narrows
(`T | null` → `T`) for the three lookups that read it: interface dispatch, union
dispatch, and the receiver's class-level type-arg prefix. That is the same
narrowing `dispatch_target_for_member_access` already applied to `x?.field`;
`try_lower_interface_dispatch` now shares that helper outright. Without the
narrowing, dispatch declines on the `Class<..> | null` union and the class
prefix comes back empty, shifting every De Bruijn slot the method's own args
occupy.

Net effect: `x?.m<T>()` emits exactly what `x.m<T>()` emits — a direct
`Call`/`VirtualCall` with the type args leading — under a null branch. The
stranded operand-stack slot goes away with it.

### Two follow-ups the normalized callee needed

Re-entering the ordinary call lowering means the *normalized* callee (`x.m`)
and the *arena node* (`x?.m`) disagree, and two places still read the arena
node directly:

1. **A callable-valued field went from two receiver evaluations to three.**
   `x?.cb(1)`, where `cb` is a function-typed field, resolves to a field, so it
   declines every direct-dispatch path and lowers the callee as a value — via
   `lower_to_operand(callee)`, which re-lowered the original `x?.cb` node under
   the guard and emitted a *second* null test plus a third receiver
   evaluation. When the third evaluation yielded null (a side-effecting
   receiver), the field read aborted with `VM internal error: type error:
   expected instance, got any`. `lower_normalized_callee_operand` now lowers
   the member access itself at the three fallback sites, restoring the
   two-evaluation shape the plain method path already had.
2. **A sys-op method reached through `?.` lost its `sys_op` opcode.**
   `sys_op_callee` / `sys_op_synthetic_type_arg_count` matched only
   `MemberAccess`, so `f?.text()` on a `baml.fs.File?` emitted a plain `call`
   of a body-less `$rust_io_function` instead of `sys_op baml.fs.File.text`.
   That path also skips the omitted-default materialization that only sys-op
   callees get, so `ctx?.output_format_with(prefix = "…")` emitted
   `load_const <omitted>` sentinels that would reach the engine. Both matches
   now accept the optional shape, and `f?.text()` emits the same `sys_op` the
   plain spelling does.

## Tests

New `crates/baml_tests/baml_src/ns_optional_chain_type_args/` runtime-output
suite, each shape pairing the optional spelling with the non-optional one it
must agree with: explicit type args vs. let-else, null receiver short-circuit,
inferred type args, class-only args, class + method args (slot order),
interface and concrete-receiver interface methods, the chained form with a
present receiver, a null receiver, and a null arising at the *second* stage, a
non-generic `?.` negative control, and the `get_field(name)?.value<T>()`
reflection shape — including that a type argument the value does not fit is
still a `baml.errors.TypeMismatch`, not a frame-layout failure.

Plus, for the follow-ups above:

- **Receiver evaluation count**, asserted through a side-effect counter for
  both a callable field and a real method: `?.` evaluates its receiver exactly
  twice, including when the receiver starts returning null after the second
  evaluation (the shape that aborted).
- **Sys-op through `?.`**: `f?.text()` vs. `file.text()` on a `baml.fs.File?`.
  The opcode itself is pinned by the namespace's bytecode snapshot.
- **Union-typed receivers**, which `?.` could not reach before: a
  `Dog | Cat | null` receiver dispatching on the runtime class, and a
  `Pair<int> | Pair<string> | null` receiver whose arms differ only in their
  class type args.

`s15_sweep_baml_src` gains its first two hir_ty error-channel entries
(`expected Speaker, got Cat | Dog`). Those are the union-receiver tests, and
they are pre-existing hir_ty imprecision rather than anything `?.` introduces:
the plain `a.speak<int>()` spelling on a `Dog | Cat` parameter records the
identical entry (checked by adding one and re-running the sweep — entries went
2 → 4). Dispatch, type args and runtime results are all correct; the corpus
simply had no union receiver calling a shared-interface method until now.

## Not in scope

Two adjacent pre-existing bugs, unchanged by this PR and not type-arg related:

1. **`?.` evaluates its receiver twice.** `get(c, "yes")?.raw` calls `get`
   twice — once for the null test, once for the access. True of `x?.field` on
   canary as well; this PR keeps the count at two, and now pins it with a test.
2. **`x?.to_string()` ICEs** (`MIR failed to resolve field access .to_string
   against class definition ...`). The `to_string`/`to_json`/`from_json` sugar
   fallbacks match on `MemberAccess`/`Path` only (`is_sugar_callee`), so they
   never fire for an optional-chained receiver.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed optional-chained generic method calls so type arguments are preserved and initialized correctly.
  - Improved dispatch across classes, interfaces, unions, inferred and explicit generics, and chained nullable calls.
  - Ensured null receivers short-circuit safely without repeated evaluation.
  - Corrected builtin I/O default argument handling and reflective field type validation.

- **Tests**
  - Added comprehensive coverage for optional chaining, generic methods, dispatch, null handling, system operations, evaluation counts, and type mismatch errors.

- **Documentation**
  - Documented the fixes in the changelog.


antoniosarosi:
Fix round. Rebased onto canary (post-#4493), applied the must-fix regression, closed the four test gaps, and broadened the blast-radius section of the description.

**Rebase.** `Field.read<T>` → `value<T>` in the three reflection tests; changelog `### Fixes` head resolved as a union of the #4493 and #4495 entries; all bytecode snapshots regenerated, never hand-edited.

**Must-fix: the callable-field regression.** Verified independently before touching anything. A side-effect counter on the receiver of `x?.cb(1)`, where `cb` is a function-typed *field*, read `3` on the pushed branch against `2` for the method spelling `x?.m(1)`; a receiver that turns null on its third evaluation aborted with `VM internal error: type error: expected instance, got any`. Cause is as reported — the fallbacks at `lower.rs:8482/8503/8509` re-lower the original `OptionalMemberAccess` arena node under the guard. Applied the reviewed `lower_normalized_callee_operand` helper; the counter reads `2` for both spellings and the aborting shape returns its value.

**A second arena-node reader, found while closing the sys-op test gap.** `sys_op_callee` / `sys_op_synthetic_type_arg_count` matched `MemberAccess` only, so `f?.text()` on a `baml.fs.File?` emitted `call baml.fs.File.text` where `file.text()` emits `sys_op baml.fs.File.text`. Bytecode dumps for both spellings, before and after. The plain-`call` path also skips the omitted-default materialization that only sys-op callees get: `ctx?.output_format_with(prefix = "…")` emitted eight `load_const <omitted>` sentinels — the exact hazard the comment at `lower_call_arg_operands` warns about — against `load_const null` for the plain spelling. Both matches now accept the optional shape, and the two spellings emit identical operand sequences.

**Test gaps closed**, all in `ns_optional_chain_type_args`:

- receiver evaluation count via a side-effect counter, for a callable field and for a method, plus the receiver that turns null after the second evaluation
- sys-op through `?.`: `f?.text()` vs `file.text()`; the opcode itself is pinned by the namespace bytecode snapshot
- union receivers: `Dog | Cat | null` dispatching on the runtime class, and `Pair<int> | Pair<string> | null` whose arms differ only in their class type args
- chained `a?.b()?.c<T>()` with null arising at the *second* stage

One thing to flag: `s15_sweep_baml_src` gains its first two hir_ty error-channel entries, from the union-receiver tests. Pre-existing imprecision, not `?.` — adding a plain `a.speak<int>()` on a `Dog | Cat` parameter takes the count 2 → 4, so the optional and plain spellings record it identically. Dispatch, type args and runtime results are all correct; the corpus simply had no union receiver calling a shared-interface method before. Noted in the suite and in the description.

**Gate** (pinned 1.93.0, `--all-features`, `--unreferenced=reject`, `-p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface`): **3792 passed, 0 failed**, no unreferenced snapshots, clean tree. A later comment-only edit to the suite shifted the sweep snapshot's byte offsets; that snapshot was regenerated and `baml_src` re-run (3108 BAML tests, 0 failed).

PR #4508, opened 2026-08-18
fix: materialize reassigned short-circuit locals

### LINKED ISSUE #4421: [bug] for loop with multi-variable reassignment + array push hangs (0.15.0) / throws spurious VM type error (0.16.0)
### Product

BAML

### Describe the bug

While writing a markdown parser in pure BAML (a `for` loop over `content.lines()` that reassigns two outer-scope `let` variables of different types — a `string` and a `bool` — inside an `if`/`else if`, with an `Array.push` in the `else if` branch), I hit two related failures depending on toolchain version:

- **canary `0.15.0`**: the `test` block **hangs indefinitely**. `baml test` (and `baml test -i "<selector>"` selecting exactly that one test) never returns — not even after 600s. `/proc/<pid>/wchan` for the hung `baml-cli test` process showed `futex_wait_queue`, and its open file descriptors were only a local `.bamlprof` write handle and two `AF_UNIX` socketpairs (no TCP/network sockets) — so this is an internal deadlock/livelock, not a network stall. Disabling telemetry (`~/.baml/telemetry.toml` → `enabled = false`) did not change the behavior.
- **canary `0.16.0`** (installed via `baml toolchain update`): the same test no longer hangs, but fails fast with a VM-internal type error instead of a normal assertion failure or compile error — `baml check` reports no errors for the same file:

  ```
  VM internal error: type error: expected string, got bool
  ```

  or, depending on the exact function body, at a different line:

  ```
  VM internal error: type error: expected int, got string
  ```

  Both come from inside `run_test`/`test_child` in `<builtin>/testing/registry.baml`, i.e. this is a VM/runtime bug, not a bug in my `.baml` source (which `baml check` accepts).

I was unable to fully isolate this to a small standalone repro I could reliably reproduce fresh outside the project where I found it (see "Reproduction Steps" — I've included both what I could reproduce and what I could not, since I'd rather be accurate than overclaim a clean minimal case). Rewriting the same logic functionally (`.lines().map(...).reduce(...)`, no outer-variable reassignment across loop iterations) made the problem disappear entirely, which is what led me to suspect the *combination* of (a) a `for` loop, (b) reassigning 2+ outer-scope variables of different types across `if`/`else if` branches, and (c) `Array.push` into an accumulator, is implicated — rather than any single one of those in isolation (a loop that only reassigns one `int` counter, e.g. `n = n + 1`, was fine at 12+ iterations in my testing).

### Reproduction Steps

**What reliably failed** (observed multiple times, both the 0.15.0 hang and the 0.16.0 type error), inside a larger project (~10 `.baml` files under `baml_src/`, `baml.toml` at a monorepo root, an accumulated `.baml/profiles/` cache directory from many prior `baml check`/`baml test` runs over months):

```baml
class Fact {
    subject_id: string,
    section: string,
    text: string,
}

function parse_subject_file(content: string, subject_id: string) -> Fact[] {
    let facts: Fact[] = [];
    let current_section = "";
    let in_scope = false;
    for (let raw_line in content.lines()) {
        let line = raw_line.trim();
        if (line.starts_with("## ")) {
            let heading = line.substring(3, line.length());
            in_scope = heading != "One-Sentence Takeaway" && heading != "References & Rabbit Holes";
            current_section = heading;
        } else if (in_scope && line.starts_with("- ")) {
            let _ = facts.push(Fact { subject_id: subject_id, section: current_section, text: line.substring(2, line.length()) });
        }
    }
    facts
}

test "repro" {
    let content = #"# Created Pipeline (Marketing)

---

## 1. First Idea

- Bullet one here.

## 2. Second Idea

- Bullet two here.
"#;
    let facts = parse_subject_file(content, "x");
    assert.equal(facts.length(), 2)
}
```

Bisection notes from the original failure (all observed as either hangs on 0.15.0 or the type error on 0.16.0, all in the *original* project):
- Removing the `---` line (a bare 3-char line, matches neither branch) made it pass.
- Removing either heading/bullet pair (leaving one `## `/`- ` pair instead of two) made it pass.
- The trigger did **not** depend on the literal characters — swapping the `---` line for any other single extra no-op line (e.g. a line reading `z`, or a second blank line) reproduced the same failure. It seemed to correlate with total loop-iteration count crossing some small threshold (~9→10 lines) in that specific file, more than with the *content* of the lines.
- A `.reduce()`-based rewrite of the same function (no `for`, no outer-variable reassignment — fold state through the accumulator's return value instead) never reproduced the bug at any size I tried.

**What I could NOT reproduce cleanly**: re-creating what looks like the identical function + test (as pasted above) in a brand-new `baml init`-scaffolded project, and re-adding it back into the original project as a new, isolated, uniquely-named test, both pass without incident on `0.16.0`. So the trigger appears to depend on something about the surrounding project/file state that I have not identified — possibly cumulative program size/complexity in the enclosing scope, something in the `.baml/profiles/` cache, or something else entirely. I'm reporting this as accurately as I can rather than shipping a repro I can't actually verify reproduces for you; happy to provide more detail (full transcript, exact repo state at time of failure, etc.) if useful.

### BAML Version

canary 0.15.0 (hang) and canary 0.16.0 (VM type error) — wrapper 0.2.0

### Language/Framework

Other

### Operating System

Linux (Debian 13 "trixie", kernel 6.12.85, x86_64)

sxlijin:
## Issue Reference

- [x] Fixes #4421

## Changes

- Restrict short-circuit phi stack-carrying to the return place and unnamed compiler temporaries, matching the existing conservative treatment of named source locals in MIR copy propagation.
- Materialize named locals assigned from short-circuit expressions so reassignment cannot leave earlier values stranded on the VM stack, while preserving the optimization for nested compiler-generated temporaries.
- Add the reported multi-variable loop and array-push scenario as O1 and O2 regression coverage, including both evaluated and direct short-circuit paths.
- Regenerate the affected bytecode corpus snapshots in the current per-namespace layout.

## Testing

- [x] `cargo fmt -p baml_compiler2_emit -p baml_tests --check -- --config imports_granularity="Crate" --config group_imports="StdExternalCrate"`
- [x] `cargo clippy -p baml_compiler2_emit --all-targets -- -D warnings`
- [x] `cargo test -p baml_compiler2_emit` (46 passed)
- [x] `cargo test -p baml_tests --test loop_reassignment` (2 passed)
- [x] `cargo test -p baml_tests --test baml_src` (3 Rust tests and 3,163 BAML tests passed)
- [x] `cargo test --lib` passed through the affected compiler and BAML suites (including 1,413 `baml_tests` tests); the local macOS run then encountered an environment-only `/install/lib/libpython3.13.dylib` loader path, and `bridge_python --lib` passes when pointed at the installed Python 3.13 library
- [x] GitHub Actions language and runtime workflows, including Rust unit, snapshot, lint, SDK, platform, size, and benchmark instrumentation jobs

## PR Checklist

- [x] I have commented the non-obvious optimization constraint

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed incorrect handling of variables reassigned across loop branches, preventing hangs and inaccurate type errors.
  - Improved nested short-circuit expression handling, including values carried across loop iterations and reassigned variables.

- **Tests**
  - Added regression coverage for loop-branch reassignment scenarios.
  - Added standard and optimized execution tests for nested short-circuit behavior and variable materialization.

### DIFF
--- a/baml_language/crates/baml_compiler2_emit/src/analysis.rs
+++ b/baml_language/crates/baml_compiler2_emit/src/analysis.rs
@@ -1208,12 +1208,25 @@ fn is_short_circuit_phi(local: Local, du: &LocalDefUse, body: &MirFunctionBody) 
         return false;
     }
 
-    // One of the defs must be a ShortCircuit terminator targeting this local.
-    du.all_defs.iter().any(|&(block_id, ref stmt_ref)| {
-        *stmt_ref == StatementRef::Terminator
+    let use_block = du.uses[0].block;
+
+    // `_0` is the return place; all other named locals are source variables
+    // and may be reassigned. Keep them in slots rather than stack-carrying a
+    // value across control-flow edges. This mirrors MIR copy propagation,
+    // which only optimizes unnamed compiler temporaries for the same reason.
+    if local != Local(0) && body.local(local).name.is_some() {
+        return false;
+    }
+
+    du.all_defs.iter().any(|&(block_id, statement_ref)| {
+        statement_ref == StatementRef::Terminator
             && matches!(
                 &body.block(block_id).terminator,
-                Some(Terminator::ShortCircuit { destination: Place::Local(l), .. }) if *l == local
+                Some(Terminator::ShortCircuit {
+                    destination: Place::Local(destination),
+                    join,
+                    ..
+                }) if *destination == local && *join == use_block
             )
     })
 }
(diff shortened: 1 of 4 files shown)


sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
CI note: the optional CodSpeed analysis compares rebased head `6afa84e4` with the pre-rebase PR head `7385f91d`, not with `canary`. Its two reported regressions are `parallel_sum_4x250k` and `parallel_sum_sequential_1m`, which contain only arithmetic loops and cannot execute the changed short-circuit analysis path. Rerunning the instrumented benchmark job reproduced the baseline artifact; both required workflows and all branch-protection contexts are green.

sxlijin:
Addressed in the stacked PR #4544 at c103d16243. The upper PR replaces the candidate predicate, so the final O1 expectations are now PhiLike for both unnamed and sound named locals, and Real for reassigned named and unnamed locals. The tests call AnalysisResult::analyze and assert the refined classifications. Validated with the 56-test baml_compiler2_emit suite, focused runtime regressions, clippy with warnings denied, and the full baml_src corpus.

Issue #4506, opened 2026-08-18
[bug] 0.17.0 canary: internal compiler error on unknown method when generic type arguments are inferred


ritunjaym:
### Product

BAML

### Describe the bug

Calling a method that does not exist on a value whose **generic type arguments were inferred** — rather than written explicitly — raises an internal compiler error instead of a normal diagnostic.

```
internal compiler error: MIR failed to resolve field access .zzz_probe_beta against class definition 'Repeat' (module_path: ["baml", "iter"]). This class should be in class_fields but isn't.
```

The trigger is inference of generic type arguments through a static factory function. It is **not** specific to the stdlib and **not** related to the package boundary: a generic class defined in local user code reproduces it just as readily, and the compiler's own message names the user module.

The minimal control pair — same class, same bogus method, only the construction style differs:

```baml
baml.iter.Repeat.new(7).zzz_probe_beta()                       // internal compiler error
baml.iter.Repeat<int> { value: 7, count: 3 }.zzz_probe_beta()  // error[E0007] — correct
```

Writing the type argument explicitly gives the correct diagnostic. Letting it be inferred from `new(7)` crashes the compiler.

A purely user-space class shows the same thing, with no stdlib involved:

```
internal compiler error: MIR failed to resolve field access .zzz_probe_delta against class definition 'ProbeGen2' (module_path: ["user", "zzprobe"]). This class should be in class_fields but isn't.
```

Non-generic classes are unaffected in every case, since they have no type arguments to infer.

Observed on `canary` at `851e763dd326659b1b1f4afab6cef5fe7b499caf`, `baml-cli 0.17.0`, Rust 1.93.0, macOS arm64.

### Reproduction Steps

**Minimal repro, no stdlib types involved.** Add a file under `baml_language/crates/baml_tests/baml_src/ns_zzprobe/probe.baml`:

```baml
class ProbeGen2<T> {
  v: T,

  function new(v: T) -> ProbeGen2<T> throws never {
    ProbeGen2<T> { v: v }
  }
}

function p() -> int throws unknown {
  ProbeGen2.new(1).zzz_probe_delta()
}

test "p" { assert.equal(p(), 1) }
```

Run:

```bash
cd baml_language
cargo build --workspace
./target/debug/baml-cli test --from crates/baml_tests/baml_src
```

Result:

```
AGGREGATE FAIL [outcome=fail]
[...]

PR #4510, opened 2026-08-18
feat(stdlib): add take/skip/take_while/skip_while iterator adapters to baml.iter

ritunjaym:
## Issue Reference

Related to #4506 — discovered while investigating that compiler crash.

**This PR should not be linked as closing that issue.** The ICE is a separate compiler-side defect and remains present after this lands; these adapters are an independent contribution that happened to surface it.

- [ ] ~~This PR fixes/closes #[issue number]~~ — intentionally left unchecked, see above

## Changes

`baml.iter` had no way to bound an iterator.

`Repeat.new(value)` defaults to `count = -1` and repeats forever ([`iter.baml:287`](https://github.com/BoundaryML/baml/blob/canary/baml_language/crates/baml_builtins2/baml_std/baml/ns_iter/iter.baml)):

```baml
function new(value: T, count: int = -1) -> Repeat<T> throws never
```

and both `collect` and `reduce` carry the doc comment *"Do not call this method on an infinite iterator as it will never terminate."* But nothing in the module could make an infinite iterator finite — the positional form `Repeat.new(v)` was effectively unusable, since the only way to bound it was to pass `count` explicitly.

This adds the four bounding adapters:

| Method | Behaviour |
| --- | --- |
| `take(n)` | Yields at most the first `n` elements; stops advancing the source afterwards. `n <= 0` yields nothing. |
| `skip(n)` | Discards the first `n` elements, yields the rest. `n <= 0` yields everything. |
| `take_while(p)` | Yields while `p` returns `true`, **stops** at the first `false`. |
| `skip_while(p)` | Discards leading elements while `p` returns `true`, then yields everything remaining. |

```baml
baml.iter.Repeat.new(0).take(5).collect()    // [0, 0, 0, 0, 0] — now terminates
baml.iter.Range.new(0, 5).skip(2).collect()  // [2, 3, 4]
```

### Implementation

Each adapter follows the existing pattern in the same file, the one used by `StepBy`, `Chain` and `Peekable`: a class holding the source state, an `implements Iterable` returning `self`, an `implements Iterator` with `next()`, plus a one-line default method on the `Iterator` interface.

`Take` and `Skip` are `<T, E>` and preserve the source's error type. `TakeWhile` and `SkipWhile` are `<T, E, E2>` with `type Error = E | E2`, matching how `Filter` and `FilterMap` combine the source's error type with the predicate's.

Semantics follow Rust's `Iterator`, which the module already models:

- `take_while` **stops** at the first failing element rather than filtering past it — `[2, 4, 5, 6, 8].take_while(is_even)` yields `[2, 4]`, not `[2, 4, 6, 8]`
- `skip_while` only skips *leading* elements — `[1, 2, 3, 1, 2].skip_while(x < 3)` yields `[3, 1, 2]`, keeping the trailing values the predicate would match

The four interface methods are placed after `peekable` so the lazy adapters stay grouped ahead of the consumers; the four classes are placed after `Chain` to keep the adapter classes together.

No Rust changes — these are pure BAML, with no `$rust_function` bindings and no new native code.

## Testing

- [x] Unit tests added/updated — 15 new tests in `crates/baml_tests/baml_src/ns_iter/iter.baml`
- [x] Tested in `baml-cli test --from crates/baml_tests/baml_src` and `cargo test -p baml_tests --test baml_src`

Full suite green, no regressions:

```
Finished 3087 passed, 0 failed, 3087 total
```

3072 before this change, +15 new tests. `cargo test -p baml_tests --test baml_src` also passes (3/3, including the regenerated bytecode snapshot).

Coverage:

- **Motivating case** — `iter_take_bounds_infinite_repeat`: `Repeat.new(0).take(5)` terminates and yields `[0, 0, 0, 0, 0]`
- **Boundaries** — `take(0)`, `skip(0)`, take more than available, skip past end
- **Chaining** — `Range.new(0, 100).skip(3).take(4)`
- **Semantics** — that `take_while` stops rather than filters, and `skip_while` only skips leading elements; both `none_match` and `all_match` predicate cases

Tests were written before the implementation and confirmed failing first (`error[E0007]: has no member 'take'`), then passing.

## Screenshots

N/A — no user-facing UI.

## PR Checklist

- [x] My code follows the style guidelines of this project — verified against `baml-cli fmt`, which reports no changes for the stdlib file
- [x] I have commented my code, particularly in hard-to-understand areas — each adapter has a doc comment with inline examples, matching the surrounding style
- [x] I have made corresponding changes to the documentation — the doc comments are the documentation for these methods

## Additional Notes

The bytecode snapshot `crates/baml_tests/snapshots/baml_src/iter.snap` is regenerated because it records every function in `baml_src`, so adding tests changes it. The diff is additive apart from an insta `assertion_line:` metadata line that regeneration drops.

I left the missing non-bounding adapters (`enumerate`, `zip`, `scan`, `nth`, `position`, `last`, `inspect`, `fuse`) out of scope. `enumerate` and `zip` in particular would need a public pair type, since BAML has no tuples — that seemed like a decision worth having separately rather than smuggling into this PR.

Happy to adjust naming, semantics, or placement if any of it cuts against the module's intended direction.

Investigated and implemented with AI-assisted tooling, reviewed and verified by me before submitting.

## Summary by CodeRabbit

* **New Features**
  * Added lazy iterator operations for limiting results with `take` and skipping items with `skip`.
  * Added predicate-based traversal with `take_while` and `skip_while`.
  * Adapters support chaining, finite and infinite sources, source exhaustion, and error propagation.

* **Tests**
  * Added comprehensive coverage for bounds, predicates, chaining, source errors, and terminal iterator behavior.


ritunjaym:
Fixed — remaining is now only decremented after source.next() confirms an item was yielded, not before. Also set remaining = 0 on Done so a finished source can't be advanced again on a later call.

Same issue existed in Skip.next() (same ordering problem in the loop), fixed there too.

Added tests covering both failure modes: a source that throws mid-iteration and is retried, and a source that returns Done once before resuming, to confirm the adapters don't lose or re-spend quota in either case.

One side effect worth flagging: this makes Take/Skip fused (they permanently stop once the source reports Done), matching Rust's Take. The rest of the module's adapters don't guarantee fusing, so there's a small inconsistency there now. Happy to discuss if that's worth addressing separately.

2kai2kai2 (approved):

sxlijin:
Thanks for the PR! @2kai2kai2 is the owner of our standard library now and will be seeing this PR through to merge

PR #4515, opened 2026-08-18
tests: relocatable test binaries - resolve corpus, snapshot and fixture paths via env overrides

harivansh-afk:
Several test suites bake absolute paths at compile time (`CARGO_MANIFEST_DIR`, workspace-relative `include_str!`, sibling-crate assumptions) or assume a rustup-managed toolchain. The binaries then only work when compiled and run inside one full workspace checkout. Running this repo's CI from nix-prebuilt test binaries (the non-gating ix preview, #4483) is what surfaced each of these, but every fix stands alone: a test binary should not stop working because it was compiled somewhere else.

Every change is inert in the current layout - each new env seam defaults to exactly today's behavior when unset, and the suites pass identically:

- **baml_tests**: one crate-level `manifest_dir()` honoring `BAML_TESTS_DIR`; all sixteen path sites (corpus dirs, snapshot dirs, build.rs's generated tests - which now bake only the relative subpath) go through it. The Prompt Fiddle demo read becomes a runtime read with `BAML_PROMPTFIDDLE_DEFAULT_BAML` as override.
- **baml_cli / baml_lsp2_actions**: a crate-local `file_snapshot!` macro honoring `BAML_CLI_SNAPSHOT_DIR` / `BAML_LSP2_SNAPSHOT_DIR`; call sites converted mechanically.
- **pack-host setup**: honors `CARGO_BUILD_TARGET` and publishes the built host's path instead of trusting sibling-crate layout; the nextest setup filter selects it by `package(=baml_cli) and kind(=test)` instead of exact binary name (trimmed cargo metadata declares only lib targets, so `binary(=pack_e2e)` hard-errors there; the package+kind form fires in exactly the same runs today).
- **sdkgen_cpp pb_generation**: workspace root via `BAML_WORKSPACE_ROOT` (protoc dies 'Could not make proto path relative' from a store-copy manifest dir).
- **sdk_test_go / sdk harness**: crate dir env override; build diagnostics embedded at compile time; a dangling mise resolution treated as not-resolved.

Same class as the already-merged #4499; split out so #4483 stays new-files-only. Merge order against #4483 does not matter.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved native package-host detection and fallback behavior across build profiles and platforms.
  - Tool resolution now safely falls back when configured paths are missing or invalid.
  - Build diagnostics remain available in more test and build environments.

- **Improvements**
  - Added configurable paths for test fixtures, snapshots, workspace roots, and generated SDK checks.
  - Snapshot and fixture handling is more portable across local and relocated checkouts.
  - Expanded validation across CLI, language tooling, compiler, and SDK workflows.


harivansh-afk:
Withdrawn for the same reason as #4485 - the caching problem moves to our side of the runner boundary, so relocatability stops being needed at all. The branch stays up: the diff doubles as documentation of the undeclared build inputs any per-crate build of baml would have to account for.

PR #4516, opened 2026-08-19
Carry minted type identity through interface dispatch

antoniosarosi:
Follows #4501, which carried runtime *definitions* through interface dispatch and closed
by naming what it did **not** fix:

> **Identity does not cross dispatch, only definitions do.** `type.of<T>()` inside an
> interface-impl method re-mints […] Nothing here depends on mint equality; making
> identity survive dispatch means carrying exact values through `realize_frame`, which is
> a separate change.

This is that change. Antonio ratified carrying the exact values through, on the same
discipline as the defs carry.

## The hole

`VirtualCall` resolves the impl from the receiver's realized `Self` and seeds the callee
frame from `resolver.realize_frame`, which produces **realized types only**. The exact
`TypeValue`s the caller minted reach the callee frame's `FrameTypeMetadata.values` from
`append_virtual_method_type_args` alone — i.e. from *method-level* type arguments. The
receiver's class-level slots got definitions but no values, so `LoadType(TypeArgRef)` in
the body fell through to `alloc_static_type_with_defs` and derived a **fresh static
digest**.

The result was a type value that named the same definition and rendered, parsed and
reflected identically, but was `==`-distinct from the one the caller passed:

```baml
class Holder<T> {
    implements Probe<T> {
        function same(self, t: type) -> bool { type.of<T>() == t }   // false
    }
}
```

Structural checks all passed, so the failure was silent: a map keyed by type missed, an
`==` against a stored type went the wrong way, and nothing reported an error.

The direct paths were already correct — a generic function call and a generic-class
instance method both thread the call site's type-argument operands, values included. Only
the resolver path dropped them.

## The fix

A runtime definition records the mint it was created with (`RuntimeTypeProvenance`), so
the caller's identity can be **read back off the definition** instead of derived afresh —
never a new mint, which is what BEP-066 I-1 requires. The interface operand already
carries the definitions the receiver's class-level slots name (that is #4501's overlay),
so those two facts together reconstruct the exact value:

- `BexVm::runtime_declaration_identity(definition_ptr)` rebuilds the minted value for a
  runtime class or enum, adding the definition's own pointer back to its provenance defs.
  `type.of_value` was already doing exactly this inline; both now share the one
  construction.
- `BexVm::minted_declaration_value(ty, defs)` resolves a frame slot's realized type
  against the operand's overlay, **only for a mint-unique name** — see below. The
  reconstructed value must also describe the same type or it is refused, so a decorated or
  parameterized spelling can never borrow a definition's mint.
- `VirtualCall` fills the owner slots with what that recovers and hands them to the callee
  frame ahead of any method-level slots, which keep their existing positions.

### Why recovery is restricted to `$dyn` names

A `DynTypeDefs` is keyed by `QualifiedTypeName`, and **only a `runtime_local` name carries
its mint in the name** (`user.$dyn.N.Foo` — what `reflect.class.new` and `reflect.enum.new`
produce). A static declaration and a compiled package's declaration are both plain
`user.Foo`. An overlay also reaches a frame whether or not the spelling being recovered is
the one that pulled it in, because `LoadType` staples the whole frame overlay onto anything
materialized there.

Matching an ordinary name against the overlay therefore answers from a *different*
definition. Both shapes are now regressions in this PR, and both were verified failing
without the `is_runtime_minted` gate:

| shape | ungated | gated |
| --- | --- | --- |
| static `Holder<Item>` in a frame that also bound a compiled package's `Item` | `type.of<T>() == type.of<Item>()` is **false**, `== package_item` is **true** | `true`, `false` |
| two compiled packages each declaring `Item`, dispatching on `Holder<B>` | `== b_item` is **false**, `== a_item` is **true** | `false`, `false` |

The first is a regression against canary, which got it right by never recovering at all.
The second is `==` lying about a type it is not, which is worse than not knowing. So
anything but a mint-unique name declines and the body re-derives normally.

## Cost

The added work is gated on the interface operand carrying definitions at all. A static
interface operand leaves `iface_defs` empty, so the owner-slot walk never runs and the
values vec never allocates; the existing no-type-args fast path is entered on exactly the
same condition as before. The four `interfaces/*` speedtest workloads are static-interface
dispatch and are structurally untouched by that reasoning — but note honestly that **no
bench covers the runtime-definitions dispatch path**, and CodSpeed was not run for this
branch, so the "definitions present" case is argued, not measured.

When the operand does carry definitions, the walk is `O(owner slots)` (0–2 in practice)
map lookups, and each *recovered* slot clones the definition's provenance defs — the same
`O(defs)` shape as the overlay clone #4501's F4 note flagged. Note the frames this runs in
are exactly the ones where `LoadType`'s static cache is already disabled (a non-empty
overlay disables it), so the recovery does not add an allocation that was previously
avoided. `Arc<DynTypeDefs>` remains the lever for both; it is not a drop-in, because GC
forwarding rewrites the pointers *inside* a `DynTypeDefs` in place, so sharing would have
to be unshared again exactly where it pays off. That reasoning is now a comment at the
clone rather than only in a PR body.

## Documented gaps

Both are the same shape — the receiver is the only thing that could carry the identity, and
`Instance` drops its class type-argument *values* at construction (it stores realized types
only; `Object`'s 64-byte assert and its Borsh wire form make a values lane a change to the
object model, not an implementation detail).

**Derived runtime types.** `t.array()`, `t.optional()` and `type.meta(…)` mint a fresh
runtime id per evaluation and attach it to no definition, so there is nothing to read back:
`Holder<RuntimeOutput[]>`'s impl body still sees a re-derived value. The alternative —
deriving a derived type's mint from its parts — would make `t.array() == t.array()` true in
ordinary code too, and contradicts I-1's "one per constructor evaluation".

**Declarations from a compiled `reflect.Package`.** Their names are not mint-unique, which
is exactly the restriction above. Their *definitions* still travel (#4501), so an impl body
can read, render and parse the type; it just does not hold the caller's identity token for
it. `runtime_package_declarations_keep_definitions_but_not_identity` pins that contract
rather than leaving it silent. The honest futures are to give compiled-package declarations
mint-unique names, or to carry the values on the receiver — the same lever as the derived
case.

Neither is a ruling this PR should make; written up separately.

## Follow-up noted in code, not fixed

`execute_call_from_locals_offset_with_type_args` restores `pending_call_type_values`
(the rooted copy) before reading `options.type_values`, which borrows an unrooted caller
local. A collection in between would leave those pointers stale. It is unreachable as
written — the callee-entry helper pushes a frame and sizes the eval stack with no TLAB
allocation, and the native path that can allocate pushes no bytecode frame, so the write
guard declines — and it predates this PR (#4501 introduced the lane). Recorded as a comment
because this lane now carries recovered identities too.

## Tests

Runtime-output oracles in `runtime_type_bindings.rs`:

| Test | Covers |
| --- | --- |
| `minted_type_identity_survives_interface_dispatch` | implements-block method, inherited default method, two-hop dispatch out of an impl body |
| `interface_impl_methods_look_up_a_type_keyed_registry` | the registry pattern — entries keyed by the call site's type value, matched by `==` inside the impl, with a same-shape same-name entry as the in-test miss |
| `dispatch_identity_separates_distinct_mints_and_leaves_static_generics_alone` | negative control (two mints of the same shape stay unequal) and static generics (`Holder<string>`) unchanged in both directions |
| `dispatch_identity_covers_owner_and_method_slots_together` | a generic method on a generic impl: owner slot recovered, method slot supplied, neither crossing |
| `dispatch_identity_covers_a_runtime_enum_slot` | the enum arm of the recovery, which nothing else reaches |
| `static_class_slots_are_not_answered_from_a_same_named_runtime_definition` | the static-vs-package collision above |
| `same_named_declarations_from_two_packages_do_not_cross_match` | the two-packages collision above |
| `runtime_package_declarations_keep_definitions_but_not_identity` | the documented gap, pinned |

Unit tests in `vm.rs` pin the owner/method slot alignment: recovered owner values precede
method-level slots, and a non-generic method still receives them.

## Verification

Focused: `runtime_type_bindings` (18/18), `bex_vm` unit + integration. The two collision
regressions were re-verified failing with the name gate removed (`false|true` in both
cases). Full pinned gate below.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved runtime-minted type identity across interface, inherited default-method, and multi-hop dispatch.
  * Improved type comparisons and registry lookups for runtime-generated classes and enums.
  * Maintained correct runtime package ownership during reflection.
  * Fixed type propagation through sparse owner slots and non-generic method dispatch.
  * Corrected truthiness handling for supported values and negation.

* **Tests**
  * Added coverage for runtime classes, enums, separate packages, same-named declarations, and dispatch scenarios.


2kai2kai2:
I am eliminating minted types in favor of a more stable representation that should eliminate the problem altogether. It should cover this case as well.

antoniosarosi:
Review round applied. The blocking finding was correct and I reproduced both regressions before fixing them.

## The blocking fix

`minted_declaration_value` looked slots up in the operand's `DynTypeDefs` by `QualifiedTypeName`, but only a `runtime_local` name (`user.$dyn.N.Foo`, from `reflect.class.new` / `reflect.enum.new`) carries its mint in the name. A static declaration and a compiled package's declaration are both plain `user.Foo` — and `LoadType` staples the whole frame overlay onto anything materialized in a frame that touched a runtime type, so the overlay is present whether or not the spelling being recovered is the one that pulled it in. The lookup could therefore answer with a *different* definition's mint.

Recovery is now gated on `name.is_runtime_minted()` in both arms (Class keeps `args.is_empty()` as well); everything else declines and re-derives normally.

Verified independently by removing the gate and re-running the new regressions:

| shape | ungated | gated |
| --- | --- | --- |
| static `Holder<Item>` in a frame that also bound a compiled package's `Item` | `false\|true` — `type.of<T>() != type.of<Item>()`, and `==` the package's mint | `true\|false` |
| two compiled packages each declaring `Item`, dispatch on `Holder<B>` | `false\|true` — `false` against its own type, `true` against A's | `false\|false` |
| `runtime_package_declarations_keep_definitions_but_not_identity` | `true\|Item` | `false\|Item` |

The first is a regression against canary (which was right by never recovering at all); the second is `==` asserting a type it is not, which is worse than not knowing. Both are `$dyn`-invisible, so the existing negative control could not have caught either.

## Consequences handled

1. **The package-declaration test flipped** and is renamed `runtime_package_declarations_keep_definitions_but_not_identity` — definitions still travel (#4501), identity does not. The claim moved into the PR body's documented-gaps section, next to the derived-types gap. Both are the same shape (only the receiver could carry it, and `Instance` drops its class type-arg values), and the write-up names the honest futures for each: mint-unique names for compiled-package declarations, or receiver-carried values.
2. **Two colliding-name regressions added** — the static-vs-package shape and the two-packages shape.
3. **PR body corrected.** "Recovery finds nothing for a static declaration" was false as written and is gone; the overlay-stapling behaviour that made it false is now stated as the reason for the name gate. The cost section also says plainly that no bench covers the runtime-definitions dispatch path and that CodSpeed was not run, so the "definitions present" case is argued rather than measured. It additionally notes that the frames this runs in are exactly the ones where `LoadType`'s static cache is already disabled, so the recovery adds no allocation that was previously avoided.
4. **Promoted to end-to-end**: `dispatch_identity_covers_owner_and_method_slots_together` (a generic method on a generic impl — owner slot recovered, method slot supplied, neither crossing) and `dispatch_identity_covers_a_runtime_enum_slot` (the enum arm, which nothing else reached). Both green.
5. **Follow-up recorded, not fixed**: the #4501-era window where `pending_call_type_values` is restored before `options.type_values` is read off an unrooted caller local. It is unreachable as written — the callee-entry helper only pushes a frame and sizes the eval stack with no TLAB allocation, and the native path that can allocate pushes no bytecode frame, so the write guard declines — but the lane now carries recovered identities too, so it is a comment at the site rather than tribal knowledge.

Focused suite: `runtime_type_bindings` 18/18. Full pinned gate below.

antoniosarosi:
Gate for the review round (toolchain 1.93.0, `--all-features --unreferenced=reject`):

```
Summary [1685.129s] 3814 tests run: 3813 passed (11 slow), 1 failed, 24 skipped
FAIL baml_tests::baml_src baml_test
```

The one failure is environmental, not this change: inside it, 3113 of 3114 stdlib cases pass and the single failing case is a local media fetch timing out —

```
ai.errors.NetworkFailure { provider: "media", detail: "media URL http://127.0.0.1:44855/doc.pdf body could not be read: ... error decoding response body ..." }
```

Re-run alone on an idle machine: **3114 passed, 0 failed**. (The box was running two agents' gates; the same test also hit the nextest slow-timeout in earlier runs. Nothing in this PR touches media fetching.)

Also green: `runtime_type_bindings` 18/18, `cargo test -p bex_vm`, `cargo clippy -p bex_vm --all-features --all-targets`, `cargo fmt --check`.

PR #4518, opened 2026-08-19
Diagnose inline unreflect type arguments that escape their call

antoniosarosi:
Closes the deferred half of B-1582 item 1 (the `STOP.md` at the repo root), implementing
ruling **(A)**: an inline `unreflect(expr)` type argument is legal only while the runtime
type stays out of the expression's published static type.

## The rule

`unreflect(v)` written inline introduces a parameter that is rigid for **one call**. The
call site publishes `occurrence_ty` (the parameter's first interface bound, or `unknown`)
in its place. So the check is an occurs-check on the callee's declared **result** type:

| declared result | verdict | why |
|---|---|---|
| `-> T` | **legal** | occurrence-substitution types a *value*. Its static type erases to `unknown`, the runtime tag rides on the value, and nothing asserts more. This is `sap.parse` / `Extract$parse<unreflect(t)>` — the supported dynamic path, used all over tests and demos. |
| `-> Wrapper<unknown>` | **legal** | declared erasure is the author's contract; `T` is not mentioned. |
| result not mentioning `T` | **legal** | `Extract$render_prompt<unreflect(t)>(..)` consumes the type inside the call. |
| `-> Wrapper<T>`, `-> T[]?`, `Agent<T>.new`, `Holder<unreflect(t)> { .. }` | **E0168** | the occurrence substitutes *into a type constructor*. The published type then asserts something about the value that stops being true the moment the call returns — and every later dispatch re-derives the receiver's class arguments from exactly that published type. |

That is the dividing line the brief asked for, stated positively: the bare-parameter result
is the single shape where occurrence-typing describes a value rather than lying about a
constructor.

The carve-out is deliberately **exactly one shape deep**: `-> T` is legal, anything
containing `T` is not. I kept the predicate literal rather than curating a list of
constructors that are supposedly safe to substitute under — one rule, no boundary to
relitigate per constructor, and the escape hatch is one line of user code. Relaxing it
later is additive; the `-> T?` case that sits closest to the line is written up under
Residuals.

One consequence worth stating up front: **the ticket's repro has *two* inline slots, and
both are named.** `ai.Agent<unreflect(t)>.new(..).run(DynamicOutput@spec<unreflect(t)>())`
reports twice — `@spec<T>` embeds the parameter the same way `Agent<T>` does. That matches
the #4501 scenario, which already writes `DynamicOutput@spec<Out>()`.

## Implementation

Two sites, one diagnostic.

- **Call result typing** — `infer.rs::report_runtime_type_escape`, called from the two
  existing recorders (`record_runtime_dependent_arguments` for source signatures,
  `record_external_runtime_dependent_arguments` for mounted ones) that between them cover
  every `write_call_type_args` road. The predicate is
  `infer.rs::runtime_param_escapes_result`: false when the result *is* the parameter, else
  `ty_mentions_param` — the same occurs machinery the runtime-parameter plumbing already
  uses. No interprocedural analysis.
- **Class literals** — `lower_expr_body.rs::lower_object_literal`. A class literal's result
  is `C<…T…>` by construction, so the answer never depends on a callee signature and the
  report is made where the written source is still at hand. The slot used to be *dropped*
  by `collect_constructor_path`, which is what left an error-recovery type in the
  instantiation and hit `runtime_ty.rs:252`
  `unreachable!("`Error` is not a valid `RuntimeTy`")`. It now holds its place (so
  inference does not also report a missing type parameter) and the diagnostic fires long
  before lowering. **No path reaches that `unreachable!` any more** — it is still live if
  called directly, but every real entry point (`run`, `pack`, `check`, runtime
  `Package.compile`) gates on error diagnostics before lowering, and `Package.compile`
  surfaces this one as a catchable E0168.

## The diagnostic (E0168)

Built by one shared factory, `runtime_type::runtime_type_must_be_named`, with an E-1 oracle
row in `constructors_own_code_and_complete_message`. (This started as E0167; #4498 landed
that code for `ConditionAlwaysConstant` while this branch was gating, so it rebased onto
E0168 — the next free code, with the same `// E0167 is owned by …` marker comment the file
already uses for E0164.)

```
E0168

  × this runtime type must be given a name before it can be used here
   ╭─[app.baml:5:12]
 5 │     Holder<unreflect(t)> { label: "h" }
   ·            ──────┬─────
   ·                  ╰── a type created at runtime only lasts for one call when written inline with `unreflect(...)`, but the value this expression creates would still need it afterwards
   ╰────
  ╰─▶   ☞ name the type first, then use the name:
        │     type Out = unreflect(t);
        │     Holder<Out> { label: "h" }
```

The rewrite is read back out of the author's own source, not reconstructed:
`RuntimeTypeNameRewrite::from_source` takes the written expression and the byte range of its
`unreflect(...)` slot and substitutes. The class-literal site has the CST; the call site
assembles it in `TirDiagnostic::render_with_type_refs`, the first point that holds both the
file text and the resolved spans (inference sees arena ids and no source at all — hence the
new `DiagnosticLocation::UnreflectArg { carrier, enclosing }`, which carries both spans
together). A half that would not print cleanly — multi-line, empty, past a length budget —
is dropped rather than guessed at, and the suggestion degrades to
`type Out = unreflect(...);`.

The `unreflect(...)` slot needed a span of its own (`AstSourceMap::unreflect_arg_spans`):
the carrier expression's span covers only `t`, not the marker and parens.

## Tests

`crates/baml_tests/tests/runtime_type_escape.rs`, thirteen cases.

Refused: the ticket's verbatim `ai.Agent<unreflect(t)>.new(..).run(DynamicOutput@spec<unreflect(t)>())`
(both slots); `STOP.md`'s minimal panic repro `Holder<unreflect(t)> { label: "h" }`; the
static-constructor sibling `Holder<unreflect(t)>.new("h")`; `-> Wrapper<T>`; `-> T[]?`.

Two rendered snapshots pin headline + note + rewrite together, one per road, because the
code+message assertions would not notice either half degrading: the class-literal report
(span and rewrite from the CST) and the call-site report (span from
`DiagnosticLocation::UnreflectArg` through `AstSourceMap`, rewrite assembled in
`render_with_type_refs` from the file text).

Accepted (each a pin from the brief): `Extract$parse<unreflect(t)>` (`-> T`);
`Extract$render_prompt<unreflect(t)>` (result never mentions `T`); `-> Wrapper<unknown>`;
the lexical `type Out = unreflect(t)` binding on the very shapes the inline one is refused
for; and the negative control — a user class *named* `unreflect`, a local binding named
`unreflect`, and a function called as `unreflect(3)` (the exact `unreflect(` shape the
type-argument lookahead keys on).

`applying_the_suggestion_compiles_and_runs` asserts the refusal and then runs the program
E0168 literally spells: it is the #4501 Agent scenario with `type Out = unreflect(..)` in
front and `Out` in the slots — compiles, dispatches through `implements Runner<Out>`, parses
the reflected output type, returns `Pixel`.

## Not done

The stretch LSP quickfix is **skipped**. `baml_lsp2_actions::fixes` has no text-edit concept
at all today — `FixKind` has a single `OpenInPlayground` variant and no `WorkspaceEdit` path
through to the LSP layer — so this would have meant building the codebase's first
diagnostic-driven edit action, which is its own change. The rewrite is already computed and
attached to the diagnostic, so a later quickfix has nothing left to derive.

## Residuals — named, not implemented

Three things the ruling's paragraph does not settle. All are conservative-side gaps (they
refuse more, or report less, never accept a lie), so none blocks this PR; flagging them for
a call rather than guessing.

1. **The `throws` clause is unchecked.** The occurs-check reads the declared *result* only.
   `f<unreflect(t)>() -> int throws Boom<T>` publishes exactly the same lying constructor in
   the effect channel and is accepted today. The wrinkle that stopped me extending it: a
   `throws` clause is often *inferred* rather than written (`FunctionSignature::throws` with
   `throws_declared: false`), so the check would start firing on effects the author never
   spelled — which is a different conversation from "you wrote this type argument".
2. **`-> T?` vs `x?.m<unreflect(t)>()` is asymmetric.** A declared `-> T?` is refused
   (`Optional` is a constructor, so the parameter occurs), while an optional-chained call to
   a `-> T` method is accepted — and both publish `unknown?`. The strict side is the safe
   one and relaxing it later is additive, but the two spellings arriving at the same
   published type by different verdicts is worth a deliberate answer.
3. **Cosmetic: a class literal's carrier expression is never lowered.** The literal road
   reports from AST lowering and leaves the slot as an error-recovery type, so
   `Holder<unreflect(nope)> { .. }` reports only E0168 — the "unresolved name `nope`"
   appears once the user applies the rewrite. The call road, which does lower its carrier,
   reports both at once.

## Gate

`cargo insta test --test-runner nextest -p baml_tests -p baml_cli -p baml_lsp2_actions
-p baml_lsp2_actions_tests -p baml_surface --all-features --unreferenced=reject`, pinned
1.93.0, `CARGO_INCREMENTAL=0`, caps `CARGO_BUILD_JOBS=24` / `NEXTEST_TEST_THREADS=24` —
**3820 passed, 0 failed, 24 skipped**, no unreferenced snapshots, no snapshots to review. `cargo fmt --all --check` and `cargo clippy --all-targets --all-features` over
the touched crates clean. Rebased onto canary after #4498 landed (E-code collision, see
above); the gate below is the post-rebase run. Not enqueued.


antoniosarosi:
Review round 1 applied, plus one thing the review could not have known about.

**The code is now E0168, not E0167.** #4498 (truthiness, B-1563) landed on canary while this
branch was gating and took E0167 for `ConditionAlwaysConstant` — GitHub flagged the branch
`CONFLICTING` and the conflict was literally the two enum variants appended at the same
spot. Rebased onto `0f130a2a8`, renumbered to the next free code, and left the same
`// E0167 is owned by …` marker comment the file already uses for E0164. Everything below
and in the PR body reads E0168; the review's E0167 references map onto it one-for-one.

**F1 (must-fix) — done.** `the_report_at_a_call_site_names_the_slot_and_spells_the_fix`,
directly after `a_result_that_wraps_the_parameter_is_refused`, is an inline snapshot of the
full rendered report for `wrap<unreflect(t)>(1)`. You were right that nothing pinned this
road: the class-literal snapshot exercises the CST path, while the call site resolves its
span through `DiagnosticLocation::UnreflectArg` → `AstSourceMap::unreflect_arg_span` and
builds its rewrite in `TirDiagnostic::render_with_type_refs` from the file text — two
distinct mechanisms that `assert_escapes` (code + message only) would have watched degrade
in silence. Snapshotted what exists, unchanged:

```
E0168

  × this runtime type must be given a name before it can be used here
   ╭─[test.baml:6:57]
 6 │ function main(t: type) -> unknown throws unknown { wrap<unreflect(t)>(1) }
   ·                                                         ──────┬─────
   ·                                                               ╰── a type created at runtime only lasts for one call when written inline with `unreflect(...)`, but the value this expression creates would still need it afterwards
   ╰────
  ╰─▶   ☞ name the type first, then use the name:
        │     type Out = unreflect(t);
        │     wrap<Out>(1)
```

**F2 — done.** "The panic is gone" now reads "no path reaches that `unreachable!` any more",
with the reason spelled out: the `unreachable!` is still live if called directly, but every
real entry point (`run`, `pack`, `check`, runtime `Package.compile`) gates on error
diagnostics before lowering, and `Package.compile` surfaces this one as a catchable E0168.

**Residuals — added to the PR body, not implemented.** All three are conservative-side (they
refuse more, or report less, never accept a lie), so none of them blocks:

1. `throws` is unchecked — `-> int throws Boom<T>` publishes the same lying constructor in
   the effect channel. The wrinkle I wrote up: a `throws` clause is frequently *inferred*
   (`throws_declared: false`), so extending the occurs-check there starts firing on effects
   the author never spelled, which is a different conversation from "you wrote this type
   argument".
2. The `-> T?` / `x?.m<unreflect(t)>()` asymmetry — the declared optional is refused, the
   optional-chained call to a `-> T` method is accepted, and both publish `unknown?`. Strict
   side is the safe one and relaxing is additive, but it deserves a deliberate answer.
3. Cosmetic — a class literal's carrier is never lowered, so `Holder<unreflect(nope)> { .. }`
   reports only E0168 and the unresolved-name error appears after applying the rewrite. The
   call road, which does lower its carrier, reports both at once.

**Gate** (post-rebase, on the exact head here): **3820 passed, 0 failed, 24 skipped**, no unreferenced snapshots, no snapshots to review. `cargo fmt --all --check` and
`cargo clippy --all-targets --all-features` over the touched crates clean. **Not enqueued.**

PR #4519, opened 2026-08-19
Add specialize() and generic-function listing to reflection

antoniosarosi:
Implements the reserved half of
[B-1582](https://linear.app/boundaryml2/issue/B-1582) item 3 — the ratified
specialization API. #4501 fixed everything about runtime types that did not need
a new surface; this is the surface.

## What it looks like

```baml
let descriptor = pkg.functions().get("root.Extract$render_prompt")
    ?? throw "not listed"

descriptor.is_generic()          // true
descriptor.generic_params()      // [ GenericParam { name: "T" } ]

let specialized = descriptor.specialize([record.as_type()])
let render = specialized.get<PromptFn>() ?? throw "no callable"
render("records").text()         // embeds the runtime class's schema
```

`baml.reflect.function.Type` gains:

| method | contract |
|---|---|
| `is_generic(self) -> bool throws never` | still expects type arguments |
| `generic_params(self) -> GenericParam[] throws never` | names + count, declaration order |
| `specialize(self, args: type[]) -> Type throws CompilationError` | arity + bounds checked |
| `get<F>(self) -> F? throws CompilationError` | the callable, through an `F` contract |

`Package.functions()` now lists **every** declared function, generics included.

## Why the listing changed

The omission was never a decision: `functions()` `filter_map`ped over
`function_type`, which returned `None` whenever `callable_signature` failed —
which is exactly what an unspecialized generic does (`TyTemplate::substitute`
hits `TypeArgRefOutOfRange`, and `.ok()` erases it). "Listed but not extractable"
already existed on canary for generic companions since #4501. This PR makes both
states first-class and actionable instead of a dead end.

## How it works

A reflection kind view *is* the `Object::Type` value (`as_type` returns the
receiver), so a descriptor has to be a `type` value that also remembers its
callable. Two additive, provenance-only payload fields:

- **`TypeValue.callable: HeapPtr`** — the `Object::GenericFunction` a descriptor
  was reflected from, null everywhere else. Outside the identity tuple: `==`/`Hash`
  stay mint-only, so two descriptors of equal type remain equal type values. GC
  traces it exactly like `owner`.
- **`GenericFunction.exact_type_values: Option<Box<[Option<TypeValue>]>>`**
  (`#[borsh(skip)]`, `None` for every compile-time instantiation, so pooled
  interned objects stay byte-identical) — the exact `type` values behind
  `type_args`. `execute_call_from_locals_offset` seeds the callee's
  `FrameTypeMetadata` from it, which is how `LoadType` hands the body's
  `type.of<T>()` back the caller's own minted value with its `DynTypeDefs`
  overlay attached. Without that lane the specialized `$render_prompt` companion
  would render a bare unresolvable name instead of the runtime class's schema.

Everything else is assembly of parts that already existed:

- **arity/genericity** — `type_args.len() < generic_param_bounds.len()`, the same
  question `unspecialized_generic_callable_name` asks (`vm.rs:2636`).
- **bounds** — the proof `validate_runtime_generic_bounds` runs before entering a
  runtime-checked generic call: substitute the bound's `args`/`assoc` against the
  completed frame, then `ImplResolver::type_implements`. Rooted at the *supplied
  value's* dynamic world (`for_value`) so a runtime-minted type's impls are
  visible, and reported as a typed diagnostic instead of a bare "mismatched types".
  `baml.AnyClass` keeps its #4493 carve-out for free.
- **`specialize`** — build a `GenericFunction` with the completed `type_args` plus
  the exact values; `callable_signature` then reconstructs. Specialize ≈ "make
  `callable_signature` succeed".
- **`get<F>`** — `Package.get_function`'s contract check, factored into
  `check_function_contract` and shared verbatim.

## Contract changes to existing pins

Each of these is a deliberate change to something previously pinned:

1. **`Package.functions()` lists unspecialized generics.**
   `function_listing_omits_unspecialized_generics` →
   `function_listing_includes_unspecialized_generics`, now also asserting
   `is_generic()` on the generic entry and its absence on the concrete one.
   The stdlib docstring's "unspecialized generic functions are omitted" claim is
   deleted.
2. **`function.Type.params()` / `return_type()` gained a throws channel**
   (`throws never` → `throws baml.reflect.errors.CompilationError`). An
   unspecialized generic descriptor has no realized function type to decompose;
   reading one was an `unreachable!` before it was reachable, and is now E0165.
   `type_kinds.rs`'s `read_views` helper declares the channel accordingly.
3. **E0165's two messages changed.** Both said reflection "cannot supply type
   arguments yet". It can now, so both name the route that works:
   `Package.functions()` → `specialize`. Extraction *by name* is still refused —
   a name lookup has nowhere to put type arguments — so
   `unspecialized_generic_get_function_reports_reflection_limit` and the four
   `reflect_call_any` message pins keep their shape and take the new text.
   `generic_function_companion_extraction_reports_reflection_limit` is unchanged
   in behaviour; its doc comment now points at the route that does work.
4. **New E0169 `ReflectSpecializationFailed`**, appended at the end of
   `DiagnosticId` (borsh discriminants are declaration-ordered), with six shared
   factories and their oracle rows in `runtime_type.rs`'s message table: arity
   mismatch, bound violation, not-generic, already-specialized, not-a-descriptor,
   and the unreconstructible-signature backstop. (E0167 went to #4498's
   always-constant-condition lint and E0168 to #4518's escaping-`unreflect`
   diagnostic while this branch was open; E0169 is the next free code at the
   rebase head.)

## Review round

**GC edges are structural now (blocker).** `TypeValue` carries three heap
pointers — its owning package, its definition overlay, and (new here) a
descriptor's callable — and six sites walked that set by hand: the collector's
major/forwarding/young arms, a frame's exact type arguments, the pending-call
lane, and the `runtime_type` provenance on classes and enums. Adding a field
meant editing all six, and missing one leaves a dangling pointer that
`get_object`'s unchecked deref turns into UB rather than a panic — which is
exactly what the first cut of this PR did. The walk now lives on the payload as
`TypeValue::gc_edges()` / `forward_gc_edges()` (with the same pair on
`DynTypeDefs` and `RuntimeTypeProvenance`, and a `young_edges` filter for the
minor-collection arms), and every site calls it. The six-copy pattern is gone,
so the next pointer-bearing field cannot repeat this. Two pre-existing gaps fell
out: a runtime class field's type value never had its `owner` forwarded, and the
class/enum provenance arms duplicated the same walk a third time.

Covered by a unit test that roots and forwards a callable-bearing frame value,
plus two BAML tests that collect between every step — one holding a descriptor
across collections and then specializing/extracting/calling it, one specializing
with a runtime-minted class and reading `type.of<T>()` back out of the callee
after a collection.

**Bounds resolve in the descriptor's world, not the caller's.** `lookup_interface`
goes through `package_for_type`, which roots at the *executing frame's* runtime
package. A bound declared inside a `Package.compile`d package is `Local` to that
package, so proving it from a host call site found no interface, no rules, and
rejected every argument — including conforming ones. `ImplResolver` now resolves
the interface name against its `root_package` first and falls back to the
lexical lookup (the name-resolution half of what `for_package` already did for
rules), and `specialize` roots at the descriptor's package. The fallback matters:
a goal can equally name an interface the inspected package *borrowed* from a
mounted dependency, which is local to that dependency and only resolvable the
lexical way — rooting alone broke `scenario_6`'s `PlanThenAct implements
app.AgentAction`. Pinned by a test that compiles a package declaring the
interface, a conforming class, and a non-conforming one: before the fix the
*conforming* type was rejected.

**Exact values are positional, not structural.** `params()`/`return_type()`
matched the supplied values against the reconstructed types by structural
equality, so a parameter the author wrote as a concrete type could be handed the
caller's type argument whenever the two happened to coincide. The mapping is now
read off the callable's own `TyTemplate`s — a position reports an exact value
only when it is written as exactly that type parameter (`TyTemplate::TypeArgRef`);
a nested occurrence (`T[]`) decomposes normally and keeps the overlay. Pinned by
a test whose second parameter is declared as the very class supplied for `T`.

**Also:** re-specializing an already-bound descriptor gets its own message
(telling the caller it "is not generic" pointed at the wrong mistake); a fully
supplied frame that still fails to reconstruct now throws instead of silently
producing an `unknown` descriptor that denies being generic; the call hot path
reads a `GenericFunction`'s two carried lanes from one deref; and `is_generic`
no longer allocates a name and an argument vector to answer an arity question.

## Notes

- A descriptor is still a `type` value, so descriptor **equality is type
  equality** — mint-only, and a specialization's mint is the static digest of
  its reconstructed function type. Two descriptors specialized from different
  runtime classes therefore compare equal when their signatures do not mention
  the type parameter (a `$render_prompt` companion is exactly that shape), even
  though their `return_type()`s differ. That is the BEP-066 rule working as
  designed — the descriptor denotes a function type, not an instantiation — but
  it is worth knowing before anyone keys a cache on one.
- The arity message is phrased against the parameters *still* awaiting
  arguments. Specialization is all-at-once today, so that is always the full
  count; if partial specialization ever lands, the message already says the
  right thing.

## Deferred

- **Bounds in `generic_params()`.** A bound's args are `TyTemplate`s over the
  callee's own frame, so `T extends Comparable<T>` has no `type` value to report
  and every workaround is a policy choice (drop silently / substitute `unknown` /
  render a string / introduce a `Bound` row). Shipped names + count; the ruling is
  written up separately. `specialize` enforces every bound regardless.
- **Static sugar `specialize<T1, …>()`.** Did not fall out cheaply — it needs a
  turbofish-to-`type.of<T>()` desugaring at the call site rather than a native.
- **Family specialize.** Companions are specialized individually, as ratified
  (`GenericList` and `GenericList$render_prompt` are separate entries).

## Tests

New `crates/baml_tests/tests/reflect_specialize.rs`, 13 cases: the item-3 flow end
to end with a runtime-minted type (asserting the rendered prompt carries the
runtime class's fields), the same shape with a static type, the `is_generic`
truth table over all four cells, `generic_params` names/count, specialized
signature readback, mint identity on both the descriptor and the callable side,
arity mismatch, an interface-bound violation, an `AnyClass`-bound violation,
specialize-on-non-generic, the unspecialized signature read, a non-descriptor
function type, and contract enforcement on extraction.

## Snapshot churn

Eight files — **6 modified, 1 added, 1 deleted** — every one a consequence of
adding one class and four methods to the stdlib:

1. `baml_cli__…__describe_package_functions_documents_unspecialized_generic_omission.snap`
   — **deleted with its test.** It existed to pin the omission contract in
   `baml describe`, and that contract is gone. Replaced by
   `…__describe_package_functions_documents_the_generic_listing_contract.snap`,
   asserting the new docstring instead.
2. `baml_cli__…__render_builtin_package_listing.snap` — one added row,
   `class baml.reflect.function.GenericParam`, plus the line-number shifts in
   `reflect.baml` from the docstring edits.
3. `baml_tests__compiler2_tir__phase5__snapshot_baml_package_items.snap` —
   `baml.reflect.function` gains `class GenericParam { methods: [] }`, and
   `class Type`'s method list gains `is_generic, generic_params, specialize, get`.
4-6. `__baml_std__` `03_ppir`, `04_5_mir`, `06_codegen` — the same class (with
   its generated `GenericParam$stream` companion) and the same four builtin
   methods, at each stage. `05_diagnostics` is unchanged: the stdlib still
   compiles clean.
7-8. `bytecode_format__bytecode_display_expanded{,_unoptimized}` — global slot
   indices shift by exactly +4 (`call 917` → `call 921` and so on), the four new
   builtin functions in the global table. No instruction changes. Regenerated
   after the rebase rather than hand-merged, since the base numbers moved too.

The reflection test suites themselves are `assert_eq!` on engine values, so they
contribute nothing here.

## Summary by CodeRabbit

* **New Features**
  * Added reflection support for generic functions, including parameter inspection, type specialization, signature access, and callable retrieval.
  * Generic functions and generated companions now appear in package function listings.
  * Added validation for specialization arguments, including count and type-bound checks.
  * Expanded truthiness behavior for the `!` operator beyond boolean values.

* **Bug Fixes**
  * Improved reflection errors and guidance for incomplete, invalid, or unavailable generic signatures.
  * Improved runtime type diagnostics with clearer messages and suggested corrections.


antoniosarosi:
Review round addressed. Rebased onto canary (`a06a1e9cb`, picking up #4498/#4503/#4518) and force-pushed; the PR body is updated throughout.

## F1 — GC edges are structural now (blocker)

Confirmed and fixed the way you asked. `TypeValue` carries three heap pointers (owner, definition overlay, and the descriptor callable this PR added) and six sites walked that set by hand — the collector's major/forwarding/young arms, a frame's exact type arguments, the pending-call lane, and the `runtime_type` provenance on classes and enums. The walk now lives on the payload:

- `TypeValue::gc_edges()` / `forward_gc_edges()`
- the same pair on `DynTypeDefs` and `RuntimeTypeProvenance`
- `BexHeap::young_edges(...)` so a minor-collection arm is its major twin plus one filter

Every site calls those; the six-copy pattern is gone. Two pre-existing gaps fell out of the consolidation: a runtime class field's type value never had its `owner` forwarded, and the class/enum provenance arms open-coded the same walk a third time.

Tests: `frame_exact_type_values_root_and_forward_every_edge` (unit, clones the `frame_type_metadata_is_included_in_vm_gc_roots` shape with a callable-bearing value and asserts both the definition and the callable are rooted *and* forwarded), plus both repro shapes as runtime tests with explicit `baml.sys.collect_garbage()` — `a_descriptor_survives_collection_between_every_step` (collects between list, specialize, extract and call) and `a_specialized_body_reads_its_own_type_across_collection` (collects between specialize, extract and call, then reads `type.of<T>()` back out of the callee and its fields).

## F2 — bounds in runtime-compiled packages

Took the preferred fix: `ImplResolver` resolves the interface *name* against its `root_package` (via a new `BexVm::lookup_interface_in` / `package_for_type_in`), and `specialize` roots at the descriptor's package. One correction to the plan — rooting alone regressed two existing tests (`scenario_6_extracts_and_calls_the_live_aliased_function`, `scenario_four_pattern_two_mounts_a_runtime_type_as_a_static_name`): a goal can equally name an interface the inspected package *borrowed* from a mounted dependency (`PlanThenAct implements app.AgentAction`), which is local to that dependency and only resolvable the lexical way. So it is root-first with a lexical fallback; the inspected package still wins a same-name collision, and resolution only picks which interface object keys the rule lookup.

Pinned by `bounds_declared_in_a_runtime_package_are_provable`, which compiles a package declaring the interface, a conforming class and a non-conforming one. Verified it fails without the fix — and fails on the *conforming* type, exactly the false rejection you described.

## F3 — positional, not structural

`signature_type_value` no longer matches by `RealizedTy` equality. `descriptor_exact_positions` reads the mapping off the callable's own `TyTemplate`s: a position reports an exact value only when it is written as exactly that type parameter (`TypeArgRef(i) -> exact[i]`); a nested occurrence (`T[]`) decomposes normally and keeps the overlay. Pinned by `exact_values_are_reported_by_slot_not_by_shape`, whose second parameter is declared as the very class supplied for `T`; verified it fails under the old structural rule.

## The rest

- **(a) E0169, not E0168.** Your correction was right and I had already hit it at the rebase: #4498 took E0167 and #4518 shipped E0168, both on canary. `ReflectSpecializationFailed` is E0169, variant appended last, `// E0167 is owned by ...` marker convention preserved. All message pins and oracle rows renumbered.
- **(b)** Snapshot paragraph corrected to the actual **6 modified + 1 added + 1 deleted**, with the `__baml_std__` stages and both bytecode snaps itemized. Bytecode snaps regenerated post-rebase, not hand-merged (the base slot numbers moved: `call 917` → `call 921`).
- **(c)** `specialize_already_specialized` — "generic function `X` is already specialized; every type parameter is bound". Test: `respecializing_reports_that_it_is_already_bound`.
- **(d)** The `RealizedTy::unknown()` fallback is now `specialize_signature_unreconstructible`. Nothing known reaches it, which is the point: the impossible case is catchable rather than a descriptor that silently reports `unknown` and denies being generic.
- **(e)** One `match self.get_object(callee_ptr)` yields both carried lanes, behind a `CarriedTypeArgs` alias (clippy's `type_complexity` otherwise).
- **(f)** `DescriptorCallable` holds two `usize`s; the name is fetched lazily only on a diagnostic path, so `is_generic` allocates nothing.
- **Notes** added to the PR body: descriptor equality is type equality (mint-only, so two descriptors specialized from different runtime classes compare equal when their signatures do not mention the parameter — a `$render_prompt` companion is exactly that shape), and the arity message already reads correctly if partial specialization ever exists.

## Gate

Full pinned gate on the rebased tree, caps 24/24: **3838 tests run, 3838 passed**, `no unreferenced snapshots found`, `no snapshots to review`, exit 0. `cargo fmt` + clippy clean on the six touched crates. Not enqueued.

antoniosarosi:
CI red is fixed. Both failures were real, and neither was a feature-flag problem — my pinned gate simply does not run the crates that caught them.

## 1. `cargo doc`

`lookup_interface_in` is `pub` and its doc linked `[`Self::package_for_type_in`]`, which is private. The doc job runs with `RUSTDOCFLAGS: -D warnings`, so `rustdoc::private_intra_doc_links` failed the whole crate. Named in plain backticks now.

Verified with the job's exact command: `RUSTDOCFLAGS="-D warnings" cargo doc --all --no-deps` → exit 0.

## 2. The three platform jobs

Not a default-features build failure — the build was fine. `cargo nextest run` fails fast, so the job died 578 tests in and reported no per-test output beyond the one failure:

```
FAIL baml_compiler_diagnostics runtime_type::tests::constructors_own_code_and_complete_message
  left:  "...its signature does not reconstruct              even with every type argument supplied"
  right: "...its signature does not reconstruct even with every type argument supplied"
```

`specialize_signature_unreconstructible` had lost the backslash on its continued string literal (my edit, not rustfmt — `cargo fmt --check` is clean on the restored form), so the message it built carried the continuation's indentation and no longer matched its oracle row. That row lives in `baml_compiler_diagnostics`' own unit tests.

**Why my gate missed it:** the pinned gate is `-p baml_tests -p baml_cli -p baml_lsp2_actions -p baml_lsp2_actions_tests -p baml_surface`. `baml_compiler_diagnostics`, `bex_vm`, `bex_heap` and the rest of the workspace's unit tests are not in it, and nothing in `baml_tests` reaches this particular diagnostic (nothing produces it — it is the impossible-case backstop). The oracle row was the only thing pinning it, and it lives in the crate the gate skips. I ran the platform job's exact invocation this time.

## Verification on the pushed head (`2e072369d`)

| command | result |
|---|---|
| `cargo doc --all --no-deps` with `RUSTDOCFLAGS=-D warnings` | exit 0 |
| `cargo nextest run --workspace --exclude baml_tests --exclude "sdk_test_*" --exclude baml_bridge -E 'not binary(=pack_e2e)'` (default features — the failing job's command) | **5198 passed, 0 failed** |
| `cargo test --no-run --workspace --exclude …` (default features, the build step) | clean |
| `cargo fmt --all --check` | clean |
| clippy `--all-features --all-targets` on the six touched crates | clean |
| full pinned gate `--all-features --unreferenced=reject` | re-running on this head; it was 3838/3838 on `1570ca900`, and the delta is one string literal in an unreachable diagnostic plus a doc comment |

Pushed, not enqueued.

antoniosarosi:
Pinned gate confirmed on the pushed head `2e072369d`: **3838 tests run, 3838 passed**, `no unreferenced snapshots found`, `no snapshots to review`, exit 0. All six local checks in the table above are now green on this commit.

PR #4520, opened 2026-08-19
fix: make generated gitignore configurable

### LINKED ISSUE #4373: sdkgen_rust: generated .gitignore ('*') makes 'git add' of the generated crate a silent no-op — document or make it configurable
The generated crate ships a `.gitignore` containing `*`, which follows from a regenerate-on-build philosophy.

Repos that **commit** the generated artifact — we do, so that CI can gate regeneration byte-parity — run `git add crates/baml_sdk` and stage **only the `.gitignore` itself**. No error, no warning, exit 0.

This cost us a real commit: our first spike commit shipped an empty crate this way, and it was only caught later. The workaround is `git add -f`, which you have to already know.

## Suggestion

A generator option (`gitignore = false`), or a line in the generated crate's README naming the `-f` requirement for commit-the-artifact workflows. Either removes the trap; the option is nicer because it makes the repo's intent explicit at the config site.

## Why a repo would commit generated code

Committing it makes the artifact reviewable in PRs and lets CI prove that a regeneration is byte-identical to what is checked in — which is how you catch a generated seam silently drifting from its source.

---
**Environment**: `0.15.1-nightly.20260731.a` (rust generator), Linux x86_64. We consume the generated crate as a committed path dependency in a production Rust service, as an in-process alternative to shelling the CLI per call (~11ms vs ~1.5–5s of process overhead).

sxlijin:
## Summary

- add a `gitignore = false` generator setting while preserving the current default
- safely remove the writer-owned catch-all when the setting is disabled and preserve custom `.gitignore` files on subsequent generation
- document the setting for committed Rust SDK crates

## Root cause

The shared SDK output writer always installed a catch-all `.gitignore`, so a normal `git add` of a generated crate silently staged only that file.

## Validation

- `cargo fmt --all -- --config imports_granularity="Crate" --config group_imports="StdExternalCrate"`
- `cargo clippy -p baml_cli --all-targets --all-features -- -D warnings`
- `cargo test -p baml_codegen_types --lib`
- `cargo test -p baml_cli --lib generator_gitignore`
- `cargo test -p baml_cli generate_rust_can_disable_generated_gitignore --test exit_code_e2e`
- `cargo test -p sdkgen_csharp --lib`
- GitHub Actions: all 53 executed checks pass across Linux, Windows, macOS, musl, wasm, SDK, snapshot, documentation, MSRV, pre-commit, and size gates

Fixes #4373


sxlijin:
@coderabbitai review

PR #4522, opened 2026-08-19
fix: generate baml_sdk in next to `baml.toml` by default

### LINKED ISSUE #4376: sdkgen_rust: default output_dir writes OUTSIDE the project directory (<project>/../baml_sdk)
With no `output_dir` set, the rust generator writes to `<project>/../baml_sdk` — a **sibling of the project root**, outside the project directory entirely.

For a repo-rooted BAML project (`baml.toml` at the repo root, `baml_src/` beside it) that means `baml generate` creates a directory in the repo's **parent** — outside version control, outside the container build context, and in a shared checkout, potentially beside somebody else's project.

`output_dir = "crates"` works fine and is what we use, so this is a default-value report rather than a blocker.

## Suggestion

Default to a path inside the project (e.g. `./baml_sdk`). A generator writing outside its own project on the default path is surprising in a way that is easy to not notice: it succeeds, and the artifact is simply somewhere unexpected.

sxlijin:
[sheep council user ](https://github.com/BoundaryML/baml/issues/4376)pointed out that if you have baml.toml in the root of your repo, baml generate defaulting to ../baml_sdk means that the generated baml_sdk is outside of our repo, which seems strictly wrong

i agree with this conclusion and i'm going to flip the default for output_dir to ./baml_sdk

for comparison, the way Cargo.toml handles this is that Cargo.toml is a sibling of src/, whereas our baml.toml pattern is to have baml_src/baml.toml - whereas instead we should have baml.toml and baml_src/ as siblings

## Summary

- default an omitted `output_dir` to the BAML project root so the generated `baml_sdk` stays inside the project
- preserve explicit `output_dir` behavior
- add resolver-level and real CLI filesystem regressions, and update the generator documentation

## Root cause

Generator configuration moved from source files, where paths were relative to `baml_src`, to `baml.toml`, where paths resolve relative to the project root. The former `..` default was retained during that migration, changing its effective destination to the parent of the project.

## Validation

- `cargo test -p baml_cli generator_output_defaults_to_project_root`
- `cargo test -p baml_cli render_keyword_generator`
- `cargo test -p baml_cli --test exit_code_e2e generate_rust_default_output_stays_inside_project`
- `cargo test -p baml_bridge --lib`
- `cargo clippy -p baml_cli --all-targets -- -D warnings`
- `cargo fmt --check -- --config imports_granularity=Crate --config group_imports=StdExternalCrate`

Fixes #4376

## Summary by CodeRabbit

* **Bug Fixes**
  * Generator output now defaults to the project directory instead of its parent.
  * Rust SDK files are created in the project’s `baml_sdk` folder by default.
  * Generator naming conventions are now validated consistently across targets, with clearer target-specific guidance.
  * Verified that default Rust SDK generation succeeds without creating files outside the project.

* **Documentation**
  * Clarified required naming conventions and Go’s additional SDK import path setting.
  * Updated Java generation documentation to reflect the revised default output location.

### DIFF
--- a/baml_language/crates/baml_builtins2/keyword_docs/baml_keywords.yaml
+++ b/baml_language/crates/baml_builtins2/keyword_docs/baml_keywords.yaml
@@ -272,7 +272,7 @@ generator:
     output_dir = "."
     naming_convention = "preserve-case"
   details: |
-    Generators are declared in `baml.toml` as `[generator.<name>]` tables, where `<name>` is an arbitrary identifier. `output_type` and `naming_convention` are required. `output_dir` is optional and defaults to `..`; BAML creates a `baml_sdk` directory inside it for every generator.
+    Generators are declared in `baml.toml` as `[generator.<name>]` tables, where `<name>` is an arbitrary identifier. `output_type` and `naming_convention` are required. `output_dir` is optional and defaults to `.`; BAML creates a `baml_sdk` directory inside it for every generator.
 
     Accepted `output_type` values:
 
--- a/baml_language/crates/baml_cli/src/generate.rs
+++ b/baml_language/crates/baml_cli/src/generate.rs
@@ -697,8 +697,8 @@ fn discover_generators(root: &Path) -> (Vec<GeneratorDef>, Vec<Diagnostic>) {
             sdkgen_go::DEFAULT_MAX_TYPED_UNION_ARITY
         };
         // `output_dir` is resolved relative to the project root and defaults
-        // to "..", with the target-owned generated directory appended.
-        let raw_output_dir = generator.output_dir.as_deref().unwrap_or("..");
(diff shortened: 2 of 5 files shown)


sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai rate limit

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai rate limit

sxlijin:
@coderabbitai review

sxlijin:
@coderabbitai full review

sxlijin:
@coderabbitai full review

PR #4529, opened 2026-08-19
Make a let binding that lives in a global behave like an ordinary binding

antoniosarosi:
Two defects from MIG_BRIEF Fix 4(b), shipped together because they are one story — *a `let`
binding that lives in a global should behave like an ordinary binding* — and because
splitting them would leave a catchable→uncatchable regression on canary in between.

## 1 — a method call on such a binding reached the VM with no receiver

```baml
client MyClient = openai.ResponsesClient.new(model = "gpt-4o-mini", …)
function main() -> string { MyClient.id() }   // VM internal: expected instance, got any
```

**This is ordinary BAML, no Session anywhere** — a `client` declaration lowers to a top-level
`let`, so it took the same defect. That is the largest behavioral surface of this PR and it
is verified by A/B against canary's lowering.

The cause: a top-level `let` is an initialized **global**, not a lexical local, so MIR's
`place_for_path` correctly finds nothing. Several roads read that absence as *"no receiver"*
rather than *"not a local"* — a single-segment receiver became `Constant::Null`, a
member-access base failed `base_is_value` so the receiver was dropped from the call, and the
container/interface dispatch block was gated on `local_for_path`. The null **is** the `any`
in the message: `Type::of` maps `ValueKind::Null` to `ObjectType::Any`. Field access and
indexing were never affected, because those roads already loaded the global — which is why
the bug presented as "method dispatch on a binding, specifically".

All roads now go through one `load_top_level_let_root`, factored out of the field-chain road
that already did this inline.

### The `.length()` holdout, and why my first diagnosis was wrong

An earlier revision of this PR shipped with `v.length()` still broken and blamed "container
methods whose owner is generic" / `builtin_kind: Some(Vm)` / emit's inline opcodes. **All
three were wrong discriminators.** `v.join(…)`, `m.keys()` and every other container method
already worked off the loaded global — A/B verified.

The real cause is one line of asymmetry in MIR: the `.length()` special case emits
`Rvalue::Len(place)`, and `Len` is the **only** consumer that takes a `Place` where every
other takes an `Operand`. The temp `lower_item_ref` defines is
`Use(Constant::GlobalItem)`, which emit's analysis classifies as a pure constant and
*virtualizes* — re-emitted at each use rather than stored. Operand consumers re-emit it
happily; the Place road read a slot nothing had written. Handing out a materialized copy
makes the place a real defined local. `v.length()` and `m.length()` are fixed, with
regressions.

Note-only, spotted there: that same match tests `"baml.string.length"` lowercase against a
class named `baml.String` — a stale dead branch.

## 2 — a literal binding took the literal's type

`s.eval("let n = 5")` bound `n` at the type `5`. A `let` item has no declaration signature,
so a reference recovers its type from the initializer's inference result —
`type_of_expr[root_expr]`, the **expression's** type. Ordinary `let` never binds that: it
applies `widen_fresh` first. This road skipped that step.

Unconditional here because a Session binding cannot opt out: `lower_session_let` refuses any
pattern ascription, so `let n: int = 5` is not a legal submission at all (pinned as a test,
since it is the reason there is no annotation branch).

### Consequences of the widening, all deliberate and all pinned

- **An eval contract naming the literal no longer accepts the binding.** `s.eval<5>("n")`
  was accepted and is now `submission result has type int, which is not a subtype of
  requested contract 5`. This is the PR's cleanest oracle — it flips exactly at the change.
- **Match exhaustiveness moves to the base type.** `match (n) { 5 => … }` on a session
  binding is now `non-exhaustive match on type int; missing: _`. The mirror also holds: a
  complete `true`/`false` match on a bool binding is now legal where it previously matched
  the literal `true` alone.
- The eval **result contract** itself still reads the unwidened initializer type
  (`let_initializer_type`), so `s.eval<5>("5")` is unchanged.

## Correction: this PR does have one regressing spelling

An earlier revision claimed "there is no spelling that regresses". **That is false**, and
here is the case: an interface-dispatched method on such a binding — `n.compare(m)`, or a
session-local `implements` on a primitive — reaches a **pre-existing broken road** and fails
with an uncatchable `InvalidArgumentCount { expected: 2, got: 1 }`, where before the widening
the same spelling was a catchable `E0007: type \`5\` has no member \`compare\``.

Verified identical before and after this PR's MIR change, so the receiver fix does not cause
it; what the widening changes is **reachability** — a literal-typed binding had no members at
all, so users never got there. Filed as MIG_BRIEF Fix 11 with the evidence. Stated here
rather than papered over, because it is the honest cost of the widening.

## What this does not fix

- **Interface dispatch on a top-level-let receiver** — above; MIG_BRIEF Fix 11.
- **A union-of-fresh-literals initializer does not widen.** `let picked = if (c) { 1 } else
  { 2 }` binds `1 | 2` where ordinary `let` binds `int`. `widen_fresh` keys on freshness, and
  union canonicalization at `finish()` drops it, so by the time a reference reads the
  recorded initializer type there is nothing left to widen. Closing it means recording the
  let's *binding* type as its own product of the let's inference — which then also has to
  avoid changing the result contract above.
- **`to_string` reports as absent** on these bindings (`type \`int\` has no member
  \`to_string\``), because the universal sugar road and the member walk these paths use are
  two different tiers and only the former carries universal members. Cosmetic today — both
  spellings error — but filed as MIG_BRIEF Fix 9, and the test that pins one of these
  messages says so in its comment.
- **Mounted-package class methods inside a Session** (E0099 shadowing, then "no member") —
  pre-existing, notebook-relevant, filed as MIG_BRIEF Fix 10.

## Related, filed not fixed

A Session assignment does not typecheck against its binding (`n = "seven"` on an `int`
binding compiles), because an assignment lowers to a fresh `let __gen = (value)` plus a
`commit_global` with no check. MIG_BRIEF Fix 8, with a recommendation appended: check
through `let_initializer_type` while leaving `let` free to re-bind. Its urgency rose with
this PR — an unchecked assignment used to be mostly inert on a `5`-typed binding, and now
produces uncatchable crashes downstream.

## Cost

`load_top_level_let_root` runs a linear scope scan (`resolve_name_at_in_scope`) on three
more roads than before. Measured in review: **+1.8% debug-compile time**. The lever if that
ever matters is memoizing the resolution per `(expr, name)` — deliberately not done here,
because caching the *loaded local* (rather than just the resolution) would have to prove the
first load dominates the second use, and the two can land in different blocks. Relatedly,
the dispatch-block `or_else` emits a dead duplicate global read when that block declines the
call; both reads are pure constant fetches, and there is now a comment saying so.

## Tests

`crates/baml_tests/tests/runtime_session.rs`, 27/27:

| Test | Covers |
| --- | --- |
| `session_top_level_lets_widen_literal_initializers` | the contract oracle — `s.eval<5>("n")` refused post-widening |
| `session_let_widening_is_visible_through_member_resolution` | int, string and bool each name their **base** type (annotated re: the `to_string` gap) |
| `session_let_widening_moves_match_exhaustiveness_to_the_base_type` | both directions of the exhaustiveness flip |
| `session_let_rebinding_across_submissions_is_unaffected` | widening changes the binding, not the values |
| `session_let_annotations_are_still_rejected` | why widening is unconditional |
| `session_let_narrowing_still_sees_the_literal` | `if (n is 5)` still narrows |
| `method_calls_on_session_let_bindings_dispatch` | every road the fix touches: primitive companion, container-as-`Call`, container-as-`Rvalue::Len`, a session-declared class's own method, a reflection handle's `fields()` |
| `method_calls_on_a_session_binding_work_in_its_own_submission` | the defect never needed two submissions |
| `session_binding_field_access_and_indexing_still_work` | the controls that always worked |
| `client_declaration_methods_dispatch` | `MyClient.id()` in ordinary BAML |

## Verification

Focused: `runtime_session` 27/27. A/B against canary's lowering for the client case, the
`join`/`keys`-vs-`length` split, and the pre-widening `E0007` for the interface-dispatch
regression. Full pinned gate below.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Session `let` and `client` bindings so literal values correctly widen to their base types.
  * Fixed method calls on bindings, including primitive, container, user-defined class, reflection, and client values.
  * Improved member access, indexing, rebinding, match exhaustiveness, and evaluation behavior.
  * Preserved constructed type identity through interface dispatch and inherited default methods.
  * Improved failure handling for unsupported interface dispatch.
* **Tests**
  * Added regression coverage for binding behavior, method dispatch, type identity, and related Session operations.


antoniosarosi:
Gate (toolchain 1.93.0, `--all-features --unreferenced=reject`):

```
Summary [2779.054s] 3827 tests run: 3826 passed (25 slow), 1 failed, 24 skipped
FAIL baml_tests::baml_src baml_test
```

Environmental, not this change: inside it 3160 of 3163 stdlib cases pass, and all three failures are local media fetches timing out against the test's own ephemeral HTTP server —

```
ai.errors.NetworkFailure { provider: "media", detail: "media URL http://127.0.0.1:34525/cat.png ... could not be fetched: Timeout" }
```

Re-run alone on an idle machine: **3163 passed, 0 failed**. (Two agents were gating concurrently; nothing here touches media fetching.)

Also green: `runtime_session` 24/24, `cargo clippy -p baml_compiler2_mir --all-features --all-targets`, `cargo fmt --check`.

antoniosarosi:
Merged canary (which had moved past #4516 and #4518) and re-gated on the merge result — clean this time, no environmental flake:

```
Summary [2884.273s] 3835 tests run: 3835 passed (23 slow), 24 skipped
info: no unreferenced snapshots found
```

The merge also resolves a duplicate changelog entry that a three-way merge left on canary: #4516's corrected wording and the wording it replaced both survived, because they landed in separate commits and #4518 inserted at the same anchor. The stale line is dropped here, so canary ends up with one entry per PR.

antoniosarosi:
Review round applied. Every finding verified independently before acting on it; the two that changed my mind are called out below.

## 1 — residual fixed, and my diagnosis was wrong twice

`v.length()` / `m.length()` now work. The reviewer's cause is right and mine was not: it is **not** generic owners, **not** `builtin_kind`, and **not** in emit. A/B on the branch:

| | before the 6-line fix | after |
| --- | --- | --- |
| `v.join("-")`, `m.keys().join("-")` | `a-b` / `k` — **already worked** | unchanged |
| `v.length()`, `m.length()` | `expected array, got any` | `2` / `2` |

So "every other container method already works" is exactly right, and the discriminator is that `Rvalue::Len` is the **only** consumer taking a `Place` where every other takes an `Operand`. The temp `lower_item_ref` defines is `Use(Constant::GlobalItem)`, which emit's analysis classifies as a pure constant and *virtualizes* rather than storing; Operand consumers re-emit it, the Place road read a slot nothing had written. Handing out a materialized copy makes the place real.

MIG_BRIEF 4(b) corrected in place — it was sending the next agent to `baml_compiler2_emit`. Bonus find noted in the PR body: the same match tests `"baml.string.length"` lowercase against class `baml.String`, a stale dead branch.

## 2 — client declarations, verified as the largest surface

A/B against canary's `lower.rs`: `MyClient.id()` in **ordinary BAML** was `VM internal error: expected instance, got any` (inside `openai.ResponsesClient.ai.Client.id`) and now returns `openai/gpt-4o-mini`. Test + changelog sentence added; the changelog now leads with the global-binding framing rather than the Session framing.

## 3 — claim corrections, all measured

- **(a) "no spelling regresses" was false, and is retracted.** `n.compare(m)` on a session binding: pre-widening a catchable `E0007: type \`5\` has no member \`compare\``, post-widening an uncatchable `InvalidArgumentCount { expected: 2, got: 1 }`. Verified identical before and after the MIR receiver fix, so the widening is what changes reachability, not the receiver fix. Filed as **MIG_BRIEF Fix 11**.
- **(b)** Both widening consequences pinned as tests: `match (n) { 5 => … }` → `non-exhaustive match on type int; missing: _`, complete `true`/`false` match on a bool binding now legal, and `s.eval<5>("n")` → `submission result has type int, which is not a subtype of requested contract 5`.
- **(c)** Union-of-fresh-literals non-widening added to "what this does not fix", with the freshness-through-joins reason.

## 4 — test hygiene

- **F5**: the flagship test is now the contract oracle (`s.eval<5>("n")`), which flips exactly at the change. The member-resolution test is kept for the int/string/bool sweep with a comment saying plainly that it pins the *type* each message names, not the message, and that it should be re-pointed if the universal-members gap is fixed. That gap is filed as **MIG_BRIEF Fix 9**.
- **F6**: `method_calls_on_session_let_bindings_dispatch` now covers a primitive companion, a container-as-`Call`, a container-as-`Rvalue::Len`, a session-declared class's own method (`p.greet()`), and a reflection handle (`cls.fields()[0].name`) — each of which was a VM error or a compiler panic before.

## 5 — F8 notes

The dropped `path_root_ty` guard is called out in the commit message as a deliberate precedence change (a callee path records no root, so requiring one would have kept exactly this case broken). The dead duplicate global read at the dispatch-block `or_else` has a comment; not fixed, because memoizing the *local* would have to prove domination across blocks.

## 6 — filed, not fixed

**Fix 9** (universal `to_string` unreachable through the member walk), **Fix 10** (mounted-package class methods lost in sessions), **Fix 11** (interface dispatch on top-level-let receivers). **Fix 8** stays unimplemented with the reviewer's recommendation and reasoning appended, including why the widening raises its urgency.

## Gate

Merged post-#4519 canary (changelog union clean) and re-gated:

```
Summary [1460.437s] 3856 tests run: 3856 passed (12 slow), 24 skipped
info: no unreferenced snapshots found
```

Focused: `runtime_session` 27/27. Perf noted in the body: +1.8% debug-compile from the extra linear scope scans, with the memoization lever and why I did not take it.

PR #4531, opened 2026-08-19
Type-check a Session assignment against its binding

antoniosarosi:
In a Session (the notebook), assigning to a variable skipped type checking entirely. This compiled:

```baml
s.eval(#"let n = 5"#);
s.eval(#"n = "seven""#);   // no error!
s.eval(#"n.abs()"#);        // VM crash you cannot catch
```

In ordinary BAML, `n = "seven"` on an int is a compile error. Now sessions behave the same way:

```baml
s.eval(#"n = "seven""#);
// error[E0001]: mismatched types — expected int, found string   (identical to ordinary code)

s.eval(#"let n = "seven""#);   // still fine — re-declaring at a new type is how you change it
s.eval(#"n += 1"#);            // compound assignments checked the same way
```

## How

Instead of adding a second checker, the session rewrite now emits the assignment *as an assignment* (`let __target = <the binding>; __target = value`), so the ordinary checker runs and the error message is the ordinary one by construction — same code, same wording, byte for byte (pinned by tests that compile the same program both ways and compare).

## Notes

- The value is spliced in **unwrapped** because parentheses change verdicts in ordinary BAML too: `n += 1.5` on an int compiles (and panics at runtime — a pre-existing ordinary-code hole), while `n += (1.5)` is a compile error. Sessions now match both behaviors exactly, including the hole. Before this PR, a session's `n += 1.5` produced 6.5 by accident; now it lands where ordinary code lands.
- The synthesized local uses a name (`__baml_assign_...`) that no user binding can ever collide with — during review we found that a user variable literally named `target_1` could silently capture the old name and turn an assignment into a no-op.


antoniosarosi:
Review round applied — blocker fixed, both hardenings in, changelog corrected off a measurement. Pushed as `053f52f`.

## Blocker — the synthesized local could collide with a user binding

Reproduced against this branch before fixing it. A user binding mints
`__baml_session_{seq}_{name}`, so a binding literally called `target_1` mints
`__baml_session_N_target_1` — exactly what the per-statement local was named when the
assignment was statement 1 of submission N. The block-local then shadows the global, and the
rewritten value reads its own copy:

```
s.eval("let n = 5")
s.eval("let target_1 = 99\nn = target_1")
s.eval<int>("n")
```

| | answer |
|---|---|
| canary | `99` |
| this branch, before the fix | **`5`** — silent, no diagnostic |
| this branch, after | `99` |

Fixed as suggested: the local is now `__baml_assign_{sequence}_{index}`, a root prefix
`internal()` cannot produce, so no user name can ever land on it. The submission above is
pinned as a fixture (`a_session_assignment_does_not_shadow_a_user_binding`), and I checked the
neighbouring shapes while I was there — same name across two submissions, and an ordinary name
in one submission — both already correct, and still correct.

## Non-blocking, same round

**Changelog — measured rather than assumed.** I ran the `n += 1.5` shape on true canary
(`git show origin/canary:…runtime_compile.rs`) rather than on this branch with the change
stashed, because HEAD already carried it. Canary answers `6.5`: the float really was committed
into the `int` binding and read back, so this is a live behavior change, and the changelog now
says so — *"used to commit the float and read back `6.5`, and now compiles and fails at runtime
exactly as the same line does in ordinary BAML."*

**Both test hardenings.** `a_session_assignment_is_no_stricter_than_ordinary_code` now asserts
all three halves: the ordinary spelling compiles, the *parenthesized* spelling does not
(`E0001: mismatched types: expected `int`, found `float`` — the freshness quirk the splice
avoids, so the tripwire can actually trip), and the ordinary program fails at RUNTIME, which is
where the Session now lands too.

**Span remapping filed** as MIG_BRIEF Fix 12: session diagnostics carry generated-module
offsets under the submission's own name, nothing maps them back, and no test asserts
`e.diagnostics[0].span` at all. Pre-existing, measurably worsened here (one line became four),
with the fix sketched — `lower_session_submission` already computes each statement's
`source_range`; it just never reaches `owned_diagnostic`.

## Gate

Full pinned gate on the final tree: **3,864/3,864 passed**, 24 skipped, no unreferenced
snapshots, doctests clean (1,447 s). CI-mirror extras (this touches `bex_project`, out of gate)
**5,200/5,200** (694 s); `RUSTDOCFLAGS="-D warnings" cargo doc --all --no-deps` clean;
`cargo fmt --all --check` and `cargo clippy --all-targets --all-features` clean. Caps 24/24,
pinned 1.93.0, `CARGO_INCREMENTAL=0`. Not enqueued.

PR #4593, opened 2026-08-25
fix(compiler): reject imprecise throws unknown declarations

codeshaunted:
## Stack context

[#4626](https://github.com/BoundaryML/baml/pull/4626), the independent deterministic LSP test fix that formed the lower stack layer, is merged into `canary`. [#4621](https://github.com/BoundaryML/baml/pull/4621), whose pure-BAML compiler-test harness is used here, is also merged. Review this PR relative to `canary`.

## Problem

BAML already infers the values that can escape through `throw`. An explicit `throws` clause is a bound on those inferred values, but `unknown` was treated as an automatic escape hatch from unused-contract diagnostics.

That allowed declarations such as these to compile without explaining the real behavior:

```baml
function DoesNotThrow() -> int throws unknown {
  1
}

function ThrowsOnlyInt() -> int throws unknown {
  throw 1
}
```

The first function does not throw. The second throws only `int`. In both cases, `throws unknown` is broader than the implementation and hides useful compiler information.

This precision requirement is specific to `unknown`. An ordinary explicit bound may remain conservative: `throws A | B | C` is valid when the function only throws `A | B`.

## After this PR

Both imprecise `unknown` declarations are rejected with E0097.

For a function that does not throw:

```text
`throws unknown` is unnecessary: BAML infers thrown types automatically, and this function does not throw. Remove the declaration; write an explicit `throws` type only to bound what may escape
```

For a function whose inferred thrown type is narrower:

```text
`throws unknown` is imprecise: this function only throws `int`. BAML infers thrown types automatically, so remove the declaration; write `throws int` only to explicitly bound what may escape
```

There are three intended remedies.

1. Let BAML infer the thrown type:

```baml
function ThrowsOnlyInt() -> int {
  throw 1
}
```

2. Keep an explicit, precise bound when the API contract should spell it out:

```baml
function ThrowsOnlyInt() -> int throws int {
  throw 1
}
```

3. Keep `throws unknown` at a real unknown boundary:

```baml
function Rethrow(value: unknown) -> int throws unknown {
  throw value
}
```

## Behavioral boundaries

| Declaration and implementation | Result |
| --- | --- |
| `throws A \| B \| C`, only `A \| B` escapes | Compiles; ordinary explicit bounds may be wider than inferred throws |
| `throws unknown`, no escaping throw | E0097: unnecessary |
| `throws unknown`, only `int` escapes | E0097: imprecise |
| `throws string \| unknown`, only `string` escapes | E0097: imprecise |
| `throws unknown`, an `unknown` value escapes directly | Compiles |
| `throws Alias`, where `type Alias = unknown`, and an unknown value escapes | Compiles |
| `throws unknown`, unknown escapes through a called function | Compiles |
| `throws string \| unknown`, both string and unknown can escape | Compiles |
| `throws unknown \| UnusedError`, unknown escapes but `UnusedError` does not | Compiles with the existing E0097 warning for the unused concrete member |
| Signature-only interface requirement with `throws unknown` | Remains a valid contract |

### Concrete superset bounds

The compiler does not require every concrete member of an explicit `throws` union to appear in the current function body. A bound such as `throws A | B | C` may intentionally reserve `C` for implementations or future behavior even when inference currently finds only `A | B`.

`unknown` is different because it admits every value and therefore erases all useful precision when no unknown-typed value can escape. Only that unused open member becomes an error.

### Written union surfaces

Finalizing `unknown | SomeError` produces semantic `unknown`, which would normally erase `SomeError`. The coverage check therefore preserves the written surface for open unions so that an unused concrete member still receives the existing warning.

Other contracts continue to use the finalized surface. This avoids false unused-member warnings for types that reduce through projections or aliases.

### Aliases and propagated throws

A direct `unknown`, an alias that resolves to `unknown`, and an `unknown` value propagated through a call all satisfy the contract. The diagnostic is based on the inferred escaping type, not only on a syntactic `throw` in the current function.

### Compiler acceptance coverage

The former `ns_fixtures/ns_throws_unknown` corpus fixture has been replaced with explicit `AssertCompiles` cases. Direct and aliased unknown bounds, mixed unions, signature-only interface bounds, ordinary unknown value flow, and callback error translation now state their expected zero-error behavior at the public compiler boundary.

Iterator functions that either do not throw or fully handle their errors no longer declare `throws unknown`.

The broader Rust integration-test programs also omit 177 executable `throws unknown` convenience clauses. Signature-only interface requirements, callback and function types, pattern checks, and the dedicated precision tests retain their explicit open contracts.

## Diagnostic assertions

The public behavior is covered in pure BAML at:

```text
crates/baml_tests/baml_src/ns_compiler/ns_throws/ns_unknown_contracts/precision.baml
```

The tests use the shared `reflect.Package._compile` harness from #4621:

- `AssertRejected` and `ExactDiagnostic` lock the exact E0097 code and message for direct, union, aliased, no-throw, and narrower-throw cases.
- `AssertCompiles` locks concrete superset bounds, accepted unknown boundaries, transitive propagation, interface signatures, and callback translation.
- A focused compiler test verifies that an unused concrete member beside `unknown` remains a warning; successful `_compile` results intentionally do not expose warnings.

## Validation

- [x] 9 pure-BAML `reflect.Package._compile` contract tests
- [x] Compiler/corpus snapshot coverage updated for the removed fixture
- [x] Mechanical audit of 177 executable test-helper migrations
- [x] Intentional signature and function-type `throws unknown` surfaces preserved
- [x] Diff and whitespace checks
- [x] CI after integration-test migration

---

## Summary by CodeRabbit

* **New Features**
  * Added clearer diagnostics for missing required fields in class constructors and object literals.
  * Improved guidance when generic functions or methods are used without specialization.
  * Enhanced validation and messaging for `throws unknown` contracts, including imprecise or unused error types.

* **Bug Fixes**
  * Prevented duplicate diagnostics for generated stream companions.
  * Preserved warnings for unused concrete error types in open throws contracts.
  * Improved handling of empty build metadata.

* **Tests**
  * Added coverage for precise `throws unknown` behavior and updated iterator exception scenarios.


baml_language/crates/baml_compiler2_hir_ty/src/diagnostics.rs
```diff
@@ -1402,6 +1404,21 @@ impl fmt::Display for TirTypeError {
...
+                    f,
+                    "`throws unknown` is only allowed when this function can actually throw an `unknown` value, but this function "
+                )?;
+                if inferred_types.is_empty() {
+                    write!(f, "does not throw")
+                } else {
+                    let inferred = inferred_types.join(" | ");
+                    write!(
+                        f,
+                        "only throws `{inferred}`; use `throws {inferred}` instead"
```
hellovai: something about: baml does throws inference, so you can leave this empty unless you want to explicitly bound what this function throws.

PR #4632, opened 2026-08-30
fix(installer): bootstrap with compatible Linux wrappers

### LINKED ISSUE #4624: Official installer wrapper requires GLIBC 2.39 and cannot bootstrap on Debian bookworm
## Summary

The official Linux installer currently installs a `baml` wrapper that requires GLIBC 2.38 and
2.39. It therefore cannot bootstrap on digest-pinned `debian:bookworm-slim` (GLIBC 2.36), even
though BAML publishes musl toolchain artifacts for Linux.

This blocks using `baml pack` in a standard Debian bookworm builder stage. The installer exits 7
before a toolchain can be selected.

## Minimal reproduction

```Dockerfile
FROM debian:bookworm-slim
RUN apt-get update \
    && apt-get install -y --no-install-recommends ca-certificates curl \
    && curl -fsSL https://pkg.boundaryml.com/install.sh | sh
```

```console
$ docker build --no-cache .
/root/.baml/bin/baml: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /root/.baml/bin/baml)
/root/.baml/bin/baml: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /root/.baml/bin/baml)
error: wrapper installed, but toolchain bootstrap failed
retry: /root/.baml/bin/baml toolchain use canary
```

Observed 2026-08-28 on Linux x86_64. The base image digest was
`sha256:abd67ffcfa541b485a3dff59865ab629aa048a6c613e639d36e7456b0b229241`.

## Expected

The official installer should work on a currently supported stable Debian base. On Linux it could
install a musl wrapper, or otherwise choose an artifact whose GLIBC floor is compatible with the
host. At minimum, the installer should diagnose the host incompatibility before installing a
wrapper it cannot execute.

## Confirmed workaround

Download the exact `x86_64-unknown-linux-musl` toolchain from the public version manifest, verify
its published SHA-256, and invoke `bin/baml-cli` directly. For 0.17.0 the manifest currently gives:

- URL: `https://github.com/BoundaryML/baml/releases/download/baml-language-0.17.0/baml-language-0.17.0-x86_64-unknown-linux-musl.tar.gz`
- SHA-256: `5841cc92dea01c9c9ec5fa9bbaad082a76d81d8780bfc3630568c89147e67755`

That archive contains both `bin/baml-cli` and `bin/baml-pack-host`, so `baml pack` can run without
the incompatible wrapper.

sxlijin:
Fixes #4624

We have two structural issues with install.sh/wrapper:

- There is no version of the wrapper that works on Alpine-based images (we don't ship a musl build).
- The wrapper only works on _really_ new Linux images, specifically it requires a newer libc than the current Debian LTS (bookworm) ships with.

This PR implements three fixes:

- install.sh now can choose glibc or musl wrapper (it prefers glibc if available, but falls back to musl if it can't find it)
- we now ship both glibc and musl builds for the wrapper
  - glibc has support for a pluggable module system to modify network path at runtime (NSS), so although we _could_ just only ship musl I opted to ship glibc+musl. (other issues around malloc performance are not relevant for the wrapper).
  - note that reqwest itself respects things like `HTTPS_PROXY` so maybe musl only would be OK, if it ends up being a significant maintenance burden
- we've lowered our glibc compatibility floor: x64 now works with glibc 2.17 and arm64 works with glibc 2.28, per the findings we made when we were choosing our backwards compatibility floors for engine

Ad-hoc testing with Debian Bookworm and Alpine on native x86_64 and aarch64 GitHub runners confirms this fix works:

- **Without this change:** [published canary install.sh and wrapper](https://github.com/BoundaryML/baml/actions/runs/33456247106) — all four platform jobs fail during toolchain bootstrap.
- **With this change:** [this PR's install.sh and workflow-built wrappers](https://github.com/BoundaryML/baml/actions/runs/33454305837) — all four platform jobs pass through installer selection, wrapper execution, toolchain download, and real CLI init/check commands.


sxlijin:
@coderabbitai review

PR #4652, opened 2026-09-01
test(installer): add wrapper bootstrap e2e workflow

sxlijin:
Stacked on #4632.

Adds an on-push workflow for exercising the wrapper bootstrap path without Fly.io:

- invokes the reusable wrapper build and limits it to the four Linux targets under test
- runs Bookworm and Alpine on native x86_64 and aarch64 GitHub runners
- serves the checked-out install.sh and workflow-built wrapper artifact from a local manifest endpoint
- verifies installer selection, wrapper execution, toolchain download, and real CLI init/check commands
- includes a local runner that can consume downloaded artifacts or a GitHub Actions run ID

The reusable workflow target filter is optional; release callers still build the complete platform matrix by default.

## Before/after evidence

Both runs cover Bookworm and Alpine on native x86_64 and aarch64 GitHub runners.

- **Before:** [published canary baseline](https://github.com/BoundaryML/baml/actions/runs/33456247106) — all four jobs fail with installer exit code 7. The canary installer selects the GNU wrapper on every Linux system; the canary wrapper requires GLIBC 2.38/2.39 on Bookworm and cannot find the glibc loader on Alpine.
- **After:** [PR install.sh and workflow-built wrappers](https://github.com/BoundaryML/baml/actions/runs/33454305837) — all four jobs pass through installer selection, wrapper execution, toolchain download, and real CLI init/check commands.

For GNU wrapper cases, the E2E fixture maps the toolchain artifact to the corresponding released musl CLI so the test isolates the wrapper ABI; the currently released GNU CLI also requires newer glibc and would otherwise confound the wrapper result.

Additional validation: actionlint, shell syntax checks, the existing Python test suite, and git diff checks.


sxlijin:
Closing the stacked E2E preparation PR as requested.

PR #4660, opened 2026-09-01
Publish versioned docs metadata at release time

hellovai:
## Outcome

Nightly and canary releases now publish generated documentation data as immutable release artifacts instead of checking derived pages into Git.

Each exact toolchain asks its own stamped CLI for:

- the complete `baml describe --packages` package list;
- one structured export for every discovered standard-library package;
- the recursive CLI command tree.

The release writes that integrity-bound envelope to:

```text
pkg.boundaryml.com/manifest/v1/docs/v<version>/stdlib.json
```

The docs build is a TypeScript-only consumer of that JSON. The detailed help snapshot also pins `describe --packages` at the layer where the option is introduced.

## End-user impact

Generated language and CLI references update with the toolchain release, without generated-file churn or a docs-maintained package list. Fully qualified symbols such as `baml.http.Request` come from the exact selected version.


hellovai:

PR #4664, opened 2026-09-01
Render multiple toolchain documentation versions

hellovai:
## Outcome

The docs build can now consume several immutable BAML toolchain artifacts at once instead of selecting exactly one version.

For readers, generated references have exact versioned routes such as:

- `/baml/language/reference/v0.18.0/baml/classes/http/Request`
- `/cli/commands/v0.18.0/check`

Existing unversioned routes remain default-version aliases, and generated pages expose a version selector that preserves the current symbol or command when it exists in the selected toolchain.

For releases, channel promotion updates a bounded discovery index at `pkg.boundaryml.com/manifest/v1/docs/versions.json`. The index points at immutable `docs/v<version>/stdlib.json` artifacts, integrity-binds their payload checksums, keeps five recent versions plus live channel aliases, and triggers the docs rebuild only after release verification succeeds. Vercel remains a TypeScript-only consumer.

## Inputs

- Default: every version in the curated discovery index
- Explicit versions: `BAML_DOCS_VERSIONS`
- Explicit URLs: `BAML_DOCS_METADATA_URLS`
- Local/CI artifacts: `BAML_DOCS_METADATA_FILES`
- Default alias: `BAML_DOCS_DEFAULT_VERSION`
- Singular environment variables remain backward compatible

## Review map

- Multi-version orchestration: `docs/scripts/prepare-generated-docs.mjs`
- Pure combined renderer: `docs/scripts/versioned-reference.mjs`
- Discovery contract: `docs/scripts/docs-metadata-source.mjs`
- Release index publication: `.github/workflows/release-baml-language.yml`
- Reader selector: `docs/components/version-switcher.tsx`

## Validation

- 30 docs tests pass, including two simultaneous toolchains and discovery-index rejection cases
- TypeScript and actionlint pass
- Production build passes with 2,974 generated routes
- Exact-version route verified in the browser

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a></sub>

PR #4665, opened 2026-09-01
Scale versioned docs builds

hellovai:
## Why

The multi-version layer makes every retained toolchain available at a stable URL, but eagerly pre-rendering every versioned symbol and command makes deployment work grow linearly with the catalog. With one toolchain selected, the portal emitted 2,972 HTML pages because the default alias and `v0.18.0` tree were both rendered in full.

## Reader-visible behavior

All routes remain available. For example:

```text
/baml/language/reference/v0.18.0/baml/classes/http/Request
/cli/commands/v0.18.0/auth/login
```

Authored pages, default-version aliases, and each version landing page are still generated during deployment. Versioned symbol and command pages are rendered on their first request and then retained in the shared cache. The version selector and fully qualified symbol names are unchanged.

## Build impact

For the same one-version source tree in CI:

- Before: 2,974 generated routes; static generation took 100 seconds.
- After: 1,499 generated routes; static generation took 59 seconds.
- Complete source-build/metadata/runtime/docs job: 7m02s before, 6m14s after.
- Local production build: 45.57 seconds.
- A non-prerendered `baml.http.Request` page returned `200`, then reported `x-nextjs-cache: HIT` and `Cache-Control: s-maxage=31536000`.

The pre-render classifier uses versions from the validated release catalog, so an ordinary package path such as `vercel/classes/...` is never mistaken for a toolchain version.

## Verification

- `pnpm --filter @baml/developer-docs test` — 33 tests passed.
- Fumadocs generation, Next route type generation, and `tsc --noEmit` passed.
- `pnpm --filter @baml/developer-docs exec next build` passed.
- Production-server requests passed for non-prerendered BAML and CLI version routes.

This PR is stacked on #4664.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,11 @@ in the set. Each release JSON contains all toolchain-discovered packages—not a
 docs-maintained list—and generated symbols use fully qualified names such as
 `baml.http.Request`.
 
+The build eagerly renders authored pages, the default-version aliases, and each
+version landing page. Versioned symbol and command pages remain normal public
+routes, but render on their first request instead of multiplying release build
+work by the number of retained toolchains.
+
 To exercise the complete producer/consumer path locally:
 
 ```sh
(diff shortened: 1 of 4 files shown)


hellovai:

PR #4668, opened 2026-09-01
Show published versions in docs previews

hellovai:
## Outcome

Exact-metadata pull request previews now show the source-built BAML toolchain alongside every curated published toolchain, and the shadcn-style header renders BAML's live GitHub star count on the server.

Before this change, production knew how to consume the multi-version release index, but CI and the authoritative Vercel preview passed only one source-built JSON file. The version selector therefore had one choice in previews. The star count also depended on a second client-side request to an internal API route, leaving the loading skeleton visible when that request never completed.

After this change:

- CI captures `pkg.boundaryml.com/manifest/v1/docs/versions.json` when it is available;
- the exact source-built PR metadata is the default and replaces the same published version instead of conflicting with it;
- every other curated immutable artifact is checksum-validated and rendered in the same preview;
- the captured index travels with the verified metadata artifact, so verification and deployment use the same version set;
- GitHub stars follow shadcn's actual server-rendered `GitHubLink` pattern, including the same `k` formatting and one-day revalidation.

The public version index currently returns 404. That is expected until the earlier release-publisher PRs in this stack land and the first canary/nightly release promotes it; previews continue safely with the exact PR toolchain until then.

## End-user example

On `baml.http.Request`, the selector can move between:

```text
/baml/language/reference/v0.18.0/baml/classes/http/Request
/baml/language/reference/v0.17.0/baml/classes/http/Request
```

If a symbol does not exist in the target version, the existing selector behavior falls back to that version's reference root.

## Verification

- Production build with a two-version fixture (`0.18.0` default + `0.17.0` nightly)
- Browser-selected `0.17.0` and verified the versioned `baml.http.Request` URL
- Browser-verified the live `9k` GitHub count in the header
- 35 docs tests
- Fumadocs generation, Next route generation, and TypeScript checking
- derived-content and runner-artifact policy checks

## Source fidelity

The shadcn reference is the actual `shadcn-ui/ui` repository at commit `63c1308`, not a visual approximation. The Fumadocs context bridge is the only structural adaptation needed to place shadcn's server-rendered GitHub link inside the client-side docs header slot.


hellovai:

PR #4672, opened 2026-09-01
Publish versioned docs runtimes

hellovai:
## Outcome

Every nightly, canary, and stable BAML release now produces an immutable browser runtime beside its generated standard-library metadata.

The exact stamped release:

- builds `bridge_wasm` once in the Web SDK build job;
- packages `v<version>/runtime.json` plus content-addressed JavaScript and WASM;
- integrity-binds version, source revision, runtime commit, toolchain, sizes, and hashes;
- runs every registered example through both the exact native CLI and exact packaged WASM;
- uploads immutable files before promoting `docs/versions.json`;
- records the runtime path and payload checksum next to `stdlib.json` in the version index;
- produces the same runtime tree in release dry runs.

Historical index entries without a runtime remain readable for migration, but every newly promoted entry is statically required by the release contract to include one.

## End-user impact

This closes the producer half of version-correct runnable docs. A follow-up stack layer will make the TypeScript-only Fumadocs build consume the selected runtime and remove the transitional checked-in snapshot.

## Verification

- 40 docs tests pass.
- All 14 release workflow graph tests pass.
- Workflow YAML validation passes.
- Docs typecheck passes against two immutable toolchain versions.
- The real 18 MB `0.18.0` runtime packaged successfully and matched the native CLI output for the registered runnable example.

### DIFF
--- a/.github/workflows/build2-web-sdk.reusable.yaml
+++ b/.github/workflows/build2-web-sdk.reusable.yaml
@@ -99,6 +99,25 @@ jobs:
           test "$(jq -r '.[0].name' package-dry-run.json)" = "@boundaryml/baml-bridge-web"
           test "$(jq -r '.[0].version' package-dry-run.json)" = "$plan_version"
 
+      - name: Build frozen developer-docs runtime
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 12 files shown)


hellovai:

PR #4673, opened 2026-09-01
Consume exact docs runtime artifacts

hellovai:
## Outcome

Fumadocs now consumes the exact browser runtime associated with its selected default toolchain, while continuing to render several immutable language and CLI reference versions together.

The build:

- resolves `runtime.json` from the curated version index or explicit file/URL inputs;
- validates payload, version, source revision, JS/WASM hashes, and delivery budgets;
- downloads only the default version's runtime, not every retained toolchain;
- materializes same-origin content-addressed assets for the worker;
- hard-fails provenance mismatches instead of pairing the wrong compiler and reference;
- removes the 17 MB checked-in WASM snapshot, generated JS, and serving manifest from Git.

Pull-request CI builds metadata and WASM from the same exact source SHA, verifies native/WASM output parity, uploads both as one preview input, and gives Vercel a TypeScript-only prebuilt artifact flow.

## End-user impact

When the docs selector/default points at a released toolchain, runnable examples now execute the matching runtime rather than a repository snapshot from an unrelated version. Historical reference versions remain visible; old entries created before runtime publishing are supported during migration.

## Verification

- 42 docs tests pass.
- Derived-content policy confirms no reference, book, or runtime artifact is tracked.
- Fresh `bridge_wasm` source build packaged as BAML `0.18.0` from commit `e4381e9…`.
- Metadata generation found 13 packages, 1,433 language items, and 29 CLI commands.
- Native CLI/WASM parity returned `"hello from BAML"`.
- Typecheck and production build pass.
- Chromium runner smoke passes: cold p95 4.48 s, warm p95 2 ms.


hellovai:

PR #4674, opened 2026-09-01
fix(typescript): escape reserved words in generated declaration names

addiplus:
## Problem

`sdkgen_typescript_shared` writes class, enum and type-alias names into the generated
TypeScript raw. A BAML `enum import` emits

```ts
export enum import {
  A = "A",
}
```

which is `TS1359: Identifier expected. 'import' is a reserved word that cannot be used here.`
The whole generated file fails to parse, so nothing in the SDK is usable.

`is_js_reserved` and `JS_RESERVED` already exist in `leaf.rs`, but they are applied only to
parameters, free functions and child-namespace re-exports. Nothing escapes a declaration name.

### Repro

```bash
baml new repro && cd repro
printf 'enum import { A B }\n' > baml_src/main.baml
baml generate add typescript/node && baml generate
grep -n 'export enum' baml_sdk/index.ts
```

Reproduced on `0.17.0` and on `0.18.1-nightly.20260828.a`.

## Fix

Add `safe_decl_name` beside the existing `safe_param_name` and apply it at the two places that
name a binding:

- `emit::build_emitted`, where an IR name becomes an emitted symbol (the class, enum and
  type-alias arms).
- `translate_ty::render_name_ref`, which re-derives every cross-reference from the IR rather
  than reading the declaration back.

Escaping one side alone would turn a parse error into a dangling reference, so both move
together. Type parameters take the same escape in `generic_decl`.

Escaping at `build_emitted` also covers `_typemap.ts`, which indexes the module namespace
through a `Record<string, unknown>` cast and so would have resolved to `undefined` at runtime
with no compile-time signal.

## Wire identity is unchanged

Only the TypeScript identifier moves. Dispatch still uses `baml_fqn`, the type map is still
keyed on the raw BAML FQN, and enum member values, marshalling parameter names and the
`$generic` / `typeParams` arrays all keep their source spelling.

## Testing

`reserved_declaration_name_is_escaped_and_wire_identity_is_preserved` in
`sdkgen_typescript_shared/src/lib.rs` covers an `enum import` through the full emit path and
asserts both halves: the declaration is escaped, and every wire-facing field keeps the raw
name.

## Relationship to my earlier PRs

This is the TypeScript half of #4070, split out so it can be reviewed on its own. The Python
and C++ halves of that PR look superseded to me: `sdkgen_python_pydantic2/src/names.rs` (#4623)
now does the Python identifier projection, and `CPP_KEYWORDS` (#4430) covers the C++ alternative
operator tokens. The TypeScript declaration names are the part neither of those reached, so it
seemed more useful to you as a small separate PR than buried in a large stale one.

It is also the change #4445 was carrying. That one was closed as a duplicate of #4070, which was
correct in scope, and this is the piece that is still outstanding.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed generated TypeScript SDKs failing when BAML declarations or generic parameters use reserved TypeScript words.
  - Ensured escaped names remain consistent across declarations and type references.
  - Preserved original names for wire-level fields, enum values, and function metadata.
  - Added coverage for reserved words, legal identifiers, cross-module references, aliases, and type parameters.

### DIFF
--- a/baml_language/sdks/typescript/sdkgen_typescript_shared/src/emit/mod.rs
+++ b/baml_language/sdks/typescript/sdkgen_typescript_shared/src/emit/mod.rs
@@ -22,6 +22,7 @@ use crate::{
         method::{MethodKind, OptionalArg, RequiredArg, TypeScriptMethodBinding},
         type_alias::TypeScriptTypeAlias,
     },
+    leaf::safe_decl_name,
     routing::{LeafPath, route},
 };
 
@@ -55,6 +56,12 @@ pub(crate) fn build_emitted(pool: &SymbolPool) -> Vec<(LeafPath, EmittedSymbol, 
         // identifier char, so a `$stream` companion class is emitted as
         // e.g. `Resume$stream` (not stripped to `Resume`). Non-stream
         // symbols are unaffected (their name carries no `$stream`).
+        //
+        // `bare` stays raw. The class / enum / type-alias arms below wrap it in
+        // `safe_decl_name` because those three names bind a module-scope
+        // identifier, where a reserved word is a parse error. Every wire-facing
+        // field on the emitted symbol (`source`, `baml_fqn`, enum member
+        // `value`) keeps the raw spelling.
         let bare = key.name().as_str().to_string();
 
         match symbol {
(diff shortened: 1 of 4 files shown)


sxlijin:
Thanks! Will look at this soon!

PR #4728, opened 2026-09-03
feat(atb2): wire the pipeline up: Supabase store, Slack, PostHog intake, run_pipeline, a Fly runner deployed from CI

ATX24:
Part 5a: the pipeline wired to real infrastructure. Stacked on #4715 (merge_issue).

## What is in here

**The store.** One table per model the pipeline passes around, in project `igraichzcidsylvzkjlc`: `feedback`, `issues`, `runs`, `merge_rounds`, `events`, `cursors`, plus the views `issues_with_outcome` and `feedback_public`. Every row carries `dataset = live | eval`; `run_tests.sh` exports `ATB2_DATASET=eval` so nothing an eval writes can pass for a user's report. The anon role reads issues, runs, events and the views; `feedback` and `cursors` are service role only, so reporter identities never leave it. The DDL was applied in the dashboard and is not in the repo. `store.baml` is the PostgREST layer: upserts, reads, row serializers, all token-free tested; every save is a no-op without `FEEDBACK_SUPABASE_*`.

**Slack out.** One thread per issue in the fix channel: created, PR opened, needs a human, merged. Reads the existing bammy bot's `ATB_SLACK_BOT_TOKEN` and `ATB_SLACK_FIX_CHANNEL` from Infisical (boundary-tools); `ATB2_` names override.

**Intake.** `baml feedback` posts a PostHog event; `intake.baml` turns those into `feedback` rows (files decoded, report id as the key), resuming from a cursor. `slack.baml` can do the same for an intake channel once the bot has `channels:history`.

**The driver.** `run_pipeline(mode, stages)` runs ingest, triage, handle and merge against the store: each `handle_issue` run becomes a `runs` row and a thread reply, each `merge_issue` round a `merge_rounds` row. `pipeline_loop` is what the runner runs.

**Deploy.** `tools/atb2/deploy` is a Fly app, `atb2-runner`: cargo, gh, node and the Claude Code CLI in the image; the cached canary clone, cargo target and run dirs on a volume; canary's `baml-cli` built there at first boot. `.github/workflows/atb2-deploy.yml` redeploys it on every push to `canary` touching `tools/atb2`. It holds one secret, `FLY_API_TOKEN`, and skips itself until that exists; the runner's own secrets are set on Fly by hand and never pass through CI. The site deploys through the Vercel GitHub app once its project is linked.

## Verified

- `baml-cli check`, `fmt`, `test` on the canary 0.18.0 compiler: 80 token-free tests pass
- read-only against the real project under the prod secrets: the store layer reads the tables, the Slack token authenticates as bammy, PostHog answers
- secret scan over the commit: clean; no install scripts piped to a shell in the image or the workflow

## Not verified

- no live `run_pipeline` round has been run; the first one writes rows and posts to Slack
- the Fly app is not created or deployed: no app is visible from this login. The runner needs `GH_TOKEN` and `CLAUDE_CODE_OAUTH_TOKEN`, which are not in Infisical yet


ATX24:
Folded into #4729, which carries this branch's commits plus babysit as one part-5 PR.

PR #4752, opened 2026-09-04
docs(setup): correct README-DEV.md's Rust toolchain instructions

ritunjaym:
`README-DEV.md` describes Rust as a mise-managed tool pinned to 1.88.0.
Neither is true: `mise.toml` declares no `rust` entry, and each workspace pins
its own toolchain in `rust-toolchain.toml` for rustup to pick up.

Four stale references, all one-liners:

- "It replaces the need for nvm, rbenv, pyenv, **rustup**, …" — rustup is
  still required.
- The `mise.toml` excerpt listed `rust = "1.88.0"`; there is no such entry.
- Manual Setup ran `rustup install 1.88.0` / `rustup default 1.88.0`, which
  `rust-toolchain.toml` overrides anyway.
- The IDE section pointed at `~/.local/share/mise/installs/rust/1.88.0/`, a
  path that never exists.

Adds a short **Toolchain** section naming the three tools that actually manage
the dev environment — mise, direnv and rustup — placed before the mise
deep-dive so the reader gets the map first. direnv was not mentioned anywhere,
despite `.envrc` being what activates mise and sets the build environment.

Two changes target the maintenance burden rather than the current text. No
version number is hardcoded in anything this PR writes, so it cannot go stale
the way `1.88.0` did: the IDE line now says Rust is rustup-managed instead of
naming a versioned path, and the `[tools]` excerpt under Configuration is
replaced by a link to `mise.toml`, since the hand-synced copy had already
drifted for `go`, `python` and `node`.

The new section defers to "Rust workspace toolchains" (#4703) rather than
restating the per-workspace pins.

Net +1 line. Documentation only; no code, build or CI changes.

Investigated with AI-assisted tooling, reviewed and verified by me before
opening.

https://claude.ai/code/session_01TmU8doXq5VDwN9WrVrDjGv

## Summary by CodeRabbit

- **Documentation**
  - Updated development setup guidance to use rustup for Rust toolchain management.
  - Documented workspace-specific toolchain configuration.
  - Revised manual installation instructions and IDE configuration guidance.
  - Removed outdated mise-based Rust setup instructions.


sxlijin:
Thanks for the PR!

Two notes:

1. I'd much rather focus on _reducing_ redundant/useless documentation, so that we have less maintenance burden. So some of this is stuff I'd rather delete. README-DEV predates some of the existing infra work, so we should basically just update it to say "we use mise, direnv, and rustup"
2. We do care about addressing coderabbit feedback here; we usually just set our agents in a loop with it.

PR #4804, opened 2026-09-09
Multi-root workspace for IDE + compiler

2kai2kai2:
The identity of each package is now a unique `SourceRoot` provided to the compiler by the IDE/cli/etc, instead of using a potentially non-unique string (name)-based identity. As a result, package names are now an aspect of the *edges* of the dependency graph, instead of the nodes, enabling different packages to refer to the same package identity by different names. While packages may still declare a name for themselves (`package.name` in `baml.toml`), this is only used as the default edge name and for user display purposes, and not as the identity. This means multiple packages can use the same self-name in the same project without causing issues, as long as other packages' edges give them distinct names.

This identity enables the IDE and other tools to utilize a multi-root workspace where many independent packages have their own viewpoints but share the same compiler database/LSP session.

This also establishes the foundations for static dependencies outside the standard library, **however** pending broader design decisions this is still currently disabled for user code.

Followup work will need to make the runtime also stop treating package name-as-identity to make this fully work for conflicting names within the same world

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple independent workspace roots and dependency-aware projects.
  * Added project names and clearer labels in Playground and VS Code project lists.
  * Added per-project working-directory handling for filesystem and process operations.
  * Expanded provider/model shorthand resolution with clearer invalid-prefix guidance.

* **Bug Fixes**
  * Improved isolation of diagnostics, symbols, and generated output between projects.
  * Fixed Playground panels and sessions when switching between project ports.
  * Improved type and diagnostic rendering across packages.


baml_language/crates/baml_builtins2/baml_std/reflect/baml.toml
```diff
@@ -1,2 +1,7 @@
 [package]
 name = "reflect"
+# Implicitly available to every package (no `[dependencies]` entry needed).
+prelude = true
```
hellovai, line 4: rename:
prelude -> implicit_compiler_dependency = true

hellovai (approved):
pending 1 suggestion

baml_language/crates/baml_builtins2/baml_std/reflect/baml.toml
hellovai: maybe "automatic_global_dependency = true" or something else idk there is probably a more verbose way of putting this so that we don't need a comment describing it.

PR #4809, opened 2026-09-10
Nine standalone fixes: build script, handle table, Python bridge, runtime-mount stubs, callable throws

aaronvg:
## Summary

Nine small, independent fixes, one commit each so any can be dropped. None depends on the interface feature in #4797.

## Bugs fixed

- **`baml_artifact/build.rs` rebuilt the compiler on every cargo command** on a packed-refs checkout (it watched a loose ref file that did not exist). It now watches the nearest existing ancestor plus `packed-refs`.
- **Handle table ran the last destructor under its write lock**, so a RustData/ADT destructor that re-enters the table deadlocked; `baml_handle_release` could also unwind across the C ABI. The row is dropped after unlocking and the release is `catch_unwind`ed to `InternalError`.
- **Python builtin failures were decoded through the application typemap.** With no generated SDK loaded, an `SdkPanic` became "Unknown class FQN", masking the real diagnostic; same-host exception identity was lost; `FunctionNotFound` was classified as a panic instead of `InvalidArgument`. `BamlFailureValue` carries builtin payloads independently of the typemap; a generated model still wins when present.
- **Python callbacks ran on a Tokio async worker**, so a callback re-entering BAML synchronously panicked and an async one could deadlock; non-coroutine awaitables were returned unawaited. Dispatch moves to `spawn_blocking` and uses `inspect.isawaitable`.
- **Tests only:** host-value reclamation after an unhandled spawn error is pinned (the previously `xfail` Python test is made deterministic via `baml.sys.collect_garbage`).
- **Runtime-mount interface stubs dropped bounds, `requires`, and associated-type bounds/defaults**, so a runtime-compiled implementor relying on a default was rejected. The stub now spells the full contract.
- **`callable_throws` lowered `throws Self.Error` / `T.Error` to `<error>`** (no owner/bound scope), giving wrong effective error types for methods like `Iterator.collect`. It now builds the same context as the signature road.
- **Python codegen rendered `never` as `None`.** It is `typing.NoReturn`.
- **Rust sdk_tests harness spawned a nested `cargo test`** from inside the fixture directory. It now runs `cargo nextest` from the workspace root with an explicit `--manifest-path` (the upward-discovery guard is kept).

## Tests

Python behaviour is tested through the generated SDK in `sdk_tests/crates/python_pydantic2` (runs in CI as `sdk_test_python_pydantic2`); only the empty-typemap cases stay in `sdks/python/tests`. Rust tests for the handle table, the artifact build script, runtime-mount stubs (unit plus a `baml_tests` e2e), `callable_throws` (three TIR cases), the codegen case matrix, and the harness runner.

## Stack

Part of the split of #4797; stacked on #4808.

## Summary by CodeRabbit

* **Bug Fixes**
  * Undefined functions now return structured `InvalidArgument` errors instead of generic SDK failures.
  * Python SDKs preserve and decode built-in runtime errors and panics when generated models are unavailable.
  * Host callbacks support custom awaitables, reentrant calls, and safer resource release.
  * Runtime-mounted interfaces preserve generic bounds, requirements, associated types, and defaults.
  * Functions that never return are typed as `typing.NoReturn` in generated Python code.
  * Compiler error types resolve correctly across implementations and bounded generics.
  * C API resource-release failures are reported safely without unwinding into callers.

* **Chores**
  * Rust SDK test tooling now uses `cargo nextest` for fixture setup and execution.

---

> [!NOTE]
> **Medium Risk**
> Touches FFI/C ABI, handle-table locking, error classification at the bridge boundary, and compiler/runtime metadata for throws and mounted interfaces—wide surface but each change is narrowly scoped with targeted tests.
> 
> **Overview**
> This PR bundles nine independent fixes across the compiler, runtime mount, CFFI handle table, Python bridge, and SDK test harness.
> 
> **Compiler & runtime compile:** `callable_throws` now lowers written `throws` clauses with the same owner/bound scope as the signature (`Self.Error`, `T.Error` on bounded generics), fixing wrong effective error types for impl methods. Runtime-mounted **interface stubs** spell generic bounds, `requires`, and associated-type bounds/defaults so cross-mount implementors can rely on defaults and inherited contracts.
> 
> **Build & FFI:** `baml_artifact`’s build script stops watching a missing loose ref on packed-refs checkouts (which forced rebuilds every cargo invocation). The CFFI handle table drops the last row **after** releasing the write lock so destructors can re-enter safely; `baml_handle_release` catches panics and returns `InternalError`. Engine `FunctionNotFound` is routed as **`baml.errors.InvalidArgument`** (Rust/Python), not `SdkPanic`.
> 
> **Python SDK:** Builtin `baml.errors.*` / `baml.panics.*` decode via **`BamlFailureValue`** when no generated model exists; pass-back encoding preserves class identity. Host callback dispatch moves to **`spawn_blocking`** with **`inspect.isawaitable`** so reentrant BAML calls don’t deadlock and non-coroutine awaitables run. Codegen maps `never` to **`typing.NoReturn`**. SDK tests use **`cargo nextest`** for Rust fixtures with an explicit workspace `--manifest-path` guard.
> 
> Tests cover callable throws scope, mount interface contracts, spawn-error host-value GC, handle release ABI, and parity fixtures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bfcbb6a1b1b710adf8db857006f02d141045528. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/baml_language/crates/baml_artifact/build.rs
+++ b/baml_language/crates/baml_artifact/build.rs
@@ -17,7 +17,24 @@ fn track_git_head() {
     if let Some(reference) = git_output(&["symbolic-ref", "-q", "HEAD"])
         && let Some(path) = git_output(&["rev-parse", "--git-path", &reference])
     {
-        println!("cargo:rerun-if-changed={path}");
+        // A packed ref has no loose file. Watching a nonexistent path makes
+        // Cargo rerun this script (and rebuild the compiler) on every command.
+        // Watch an existing ancestor to notice a future loose ref, plus the
+        // packed-ref file for updates to the current packed value.
+        let mut watched = PathBuf::from(path);
+        while !watched.exists() {
+            if !watched.pop() {
+                break;
+            }
+        }
+        if watched.exists() {
+            println!("cargo:rerun-if-changed={}", watched.display());
+        }
+        if let Some(packed) = git_output(&["rev-parse", "--git-path", "packed-refs"])
+            && PathBuf::from(&packed).exists()
+        {
+            println!("cargo:rerun-if-changed={packed}");
+        }
     }
 }
 
--- a/baml_language/crates/baml_compiler2_hir_ty/src/callable.rs
+++ b/baml_language/crates/baml_compiler2_hir_ty/src/callable.rs
@@ -184,7 +184,17 @@ pub fn callable_throws<'db>(
     let data = baml_compiler2_ppir::item_data::elaborated_function_data(db, function);
     if let Some(throws_ref) = data.throws {
         let frame = crate::lower::function_generic_frame(db, function);
-        let ctx = crate::lower::lower_ctx_for_file(db, function.file(db)).with_frame(frame);
+        // A throwing effect has the same owner/bound scope as the rest of
+        // the signature. A frame alone cannot resolve `Self.Error` in an
+        // implements-block or free-impl method (it projects through the
+        // impl target's binding) or `T.Error` on a bounded parameter (it
+        // projects through `T`'s bound).
+        let impl_target = crate::lower::owner_impl_target(db, function, &frame);
+        let ctx = crate::lower::lower_ctx_for_file(db, function.file(db))
+            .with_frame(frame)
+            .with_bounds(crate::lower::function_generic_bounds(db, function))
+            .with_self_ty(crate::lower::owner_self_ty(db, function))
+            .with_impl_target(impl_target);
         let lowered = ctx.lower_type_ref(&data.type_refs, throws_ref);
         // A PARTIAL clause (`throws T | _`, spec Functions rule 3) keeps
         // inferring: fall through to the body run, whose finalize unions
--- a/baml_language/crates/baml_tests/src/compiler2_tir/callable_throws_scope.rs
+++ b/baml_language/crates/baml_tests/src/compiler2_tir/callable_throws_scope.rs
@@ -1,0 +1,108 @@
+//! `callable_throws` lowers a written `throws` clause in the same scope as
+//! the rest of the signature. `Self.Error` in an implements-block or free-impl
+//! method resolves through the impl target's binding, and `T.Error` on a
+//! bounded parameter resolves through `T`'s bound - a bare generic frame
+//! cannot lower either, which used to leave the runtime throws metadata
+//! wrong for exactly the methods that declare their error via an
+//! associated type (e.g. `baml.iter.Iterator.collect`).
+
+use baml_compiler2_ppir::item_data::MethodOwner;
+use baml_db::ProjectDatabase;
+
+use super::support::make_db;
+use crate::engine::TestDbExt;
+
+const SOURCE: &str = r#"
+interface Fallible {
+    type Error
+    function run(self) -> int throws Self.Error
+}
+
+class Worker {
+    implements Fallible {
+        type Error = string
+        function run(self) -> int throws Self.Error { throw "boom" }
+    }
+}
+
+class Other {}
+
+implement Fallible for Other {
+    type Error = int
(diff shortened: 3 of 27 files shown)


sxlijin:
Tried to reproduce all nine fixes on canary through the *generated* Python SDK (the sdk_tests `function_calls` fixture as produced by `sdkgen_python_pydantic2`, plus `baml-cli init` / `generate add python/pydantic2` / `generate` for the codegen checks), each scenario in a fresh interpreter with a timeout. Four are real user-reachable bugs through the SDK, two are real tooling bugs, one is tests-only by its own description, two are not reachable from Python.

| # | Claim | Verdict | Evidence |
|---|-------|---------|----------|
| 1 | `build.rs` rebuilds on every cargo command with packed refs | Reproduced (tooling) | Worktree with packed refs: `cargo check -p baml_artifact` prints `Compiling baml_artifact` on the 2nd and 3rd consecutive run. |
| 2 | Last handle destructor under the table write lock deadlocks; panic across the C ABI | Unreachable from Python | No row type Python can hold has a destructor that touches the table (`HandleInner::drop` releases a heap slot, `ResourceHandleInner::drop` hits the resource registry, the only user-visible `RustData` is the `CompileArtifact` mutex, `HostValueArc` is not a row). Releasing 400 closure/media handles from two threads: no deadlock. Hardening; "deadlocked" should read "would deadlock". |
| 3 | Builtin failures decoded through the application typemap | Split | (a) `FunctionNotFound` → `SdkPanic`: **reproduced** with the generated SDK loaded, via `call_function_sync(get_runtime(), "user.nope", {})` (generated bindings can't name a missing function). (b) "Unknown class FQN" masking and lost `raised is caught` identity: **only without a generated SDK**; with `baml_sdk` imported the typemap ships the `baml.errors.*` / `baml.panics.*` models and a callback's `MyErr` comes back as the same object. |
| 4 | Python callbacks ran on a Tokio async worker | **Reproduced, ordinary usage** | Sync callback calling a generated sync function → `SdkPanic: Cannot start a runtime from within a runtime`; async callback awaiting `make_adder_async` → hangs, killed at 20 s; callback returning a custom `__await__` object or an `asyncio.Future` → `TypeError: Cannot encode ...`. |
| 5 | Tests only | Confirmed | Callback not collectible after one call, collectible after `baml.sys.collect_garbage()` on canary. |
| 6 | Runtime-mount stubs drop bounds / `requires` / associated-type defaults | **Reproduced** | The PR's program run through the Python bridge: `CompilationError: missing associated type binding `Value` for interface `app.Counter``. |
| 7 | `throws Self.Error` / `T.Error` lowered to `<error>` | Not reproducible from Python | `run_worker` / `run_other` / `drive_worker` / `drive_other` all deliver the right thrown value; `Iterator.collect` works; the Python generator emits no throws types and no interface methods, so the wrong `callable_throws` never surfaces. Compiler-internal consistency. |
| 8 | `never` rendered as `None` | **Reproduced** | Generated `__init__.pyi`: `def fail() -> None: ...` for `function fail() -> never throws string`. Type hint only. |
| 9 | Rust harness nested `cargo test` | n/a | Tooling, not evaluated. |

On cost: #4 is the substantive fix, keep. #3 is ~300 lines (`BamlFailureValue`, `_decode_failure_value`, encoder pass-back, `has_class`); through the generated SDK the only reachable defect is the one-arm `FunctionNotFound` → `InvalidArgument` change in `result_to_outbound`, the rest serves bridge-without-generated-SDK usage, so it stands or falls on whether that mode is supported. #2 is small hardening for nothing reachable. #6 and #8 are correctness fixes with reproductions. #7 is harmless but "wrong effective error types for methods like `Iterator.collect`" was not observable.

PR #4810, opened 2026-09-10
bridge_ctypes/bridge_cffi: leak-free encode/decode ledgers and pinned call targets

aaronvg:
## Summary

Four ownership fixes at the C-FFI boundary, kept to two small ledgers: an outbound insert log that rolls back on a failed encode, and an inbound capture batch that takes every handle before decoding. No sessions, receipts, or new wire tags. `prepare_call` becomes the single definition of argument decoding and target pinning, replacing per-adapter copies in the C ABI, Wasm, PyO3, JNI and napi bridges.

## Bugs fixed

- **Outbound encode leaked earlier children when a later child failed.** Rows inserted for opaque children before the failing one never reached the host and were never released. `OutboundOwnership` releases them on drop and commits only when the whole aggregate encoded.
- **Inbound decode leaked keys after a malformed sibling**, and a host key with a forged kind hit a `debug_assert` at the boundary. `InboundTransfer::capture` drains every engine key and interns every host key first; undecoded leftovers are released when the batch drops; a forged kind is `CtypesError::InvalidHostValueKind`.
- **`complete_host_call` with an invalid `is_error` flag returned before interning the payload's host keys**, leaking the bridge's registry entries. References are captured before the flag check.
- **A callable's handle-table key was resolved inside the spawned task**, racing an SDK that released it right after issuing the call ("callable handle is no longer live"). `prepare_call` pins the target synchronously; `PreparedCall` holds it until `invoke_prepared` ran.

## Tests

Rust, in-crate (these are wire and ownership assertions only Rust can express, run under nextest's process-per-test): `failed_output_aggregate_rolls_back_earlier_owned_handles`, `failed_input_aggregate_releases_visited_and_unvisited_transfers`, `forged_host_kind_is_rejected_without_replacing_the_live_registration`, `invalid_completion_flag_releases_the_parsed_payload`, `prepared_handle_call_survives_release_of_its_key`.

## Stack

Part of the split of #4797; stacked on #4809.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when invoking functions and callable values across supported language bridges.
  - Ensured temporary handles and host values are released correctly when decoding or encoding fails.
  - Added validation for mismatched host-value types with clearer errors.
  - Preserved callable targets throughout asynchronous execution, even if their original handle is released.
  - Empty successful results are now handled consistently, while invalid completion payloads report clearer failures.

- **Refactor**
  - Standardized call preparation and execution across native, WebAssembly, Java, Python, and TypeScript integrations.

---

> [!NOTE]
> **Medium Risk**
> Changes core FFI ownership, host-value identity, and async call pinning across all language bridges; bugs could leak handles or break callable invocation, but behavior is covered by targeted ownership and pinning tests.
> 
> **Overview**
> This PR hardens **handle and host-value ownership** at the C-FFI boundary and **unifies how every adapter runs BAML calls**.
> 
> **Inbound:** `InboundTransfer` walks a parsed `InboundValue` (or kwargs) and **captures** every engine handle lease and host key **before** validation/decode. Failed decodes still release what was taken; forged host-value kinds use `HostValueArc::try_intern` and surface as `CtypesError::InvalidHostValueKind` instead of asserting. `complete_host_call` now captures the payload the same way **before** checking `is_error`, so rejected completions do not leak host keys.
> 
> **Outbound:** `OutboundOwnership` tracks handle-table rows inserted during one encode; **failed partial encodes roll back** inserts (including `build_to_host_call`).
> 
> **Calls:** `call_and_encode` / `call_handle_and_encode` and per-SDK decode paths are replaced by **`prepare_call`** (decode + `InboundTransfer` + **synchronous pin** of callable heap handles) and **`invoke_prepared`**. C ABI, Wasm, PyO3, JNI, and napi all use this pair, fixing the race where the SDK released a callable key between issuing a call and the async task resolving it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b6854739befc183f83aa4dfb09177c6dd334094. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/baml_language/crates/bex_resource_types/src/host_value.rs
+++ b/baml_language/crates/bex_resource_types/src/host_value.rs
@@ -81,8 +81,7 @@ impl HostValueArc {
     /// when the last clone (across the entire process) drops — even when BAML
     /// returns a host callable to the host and it is later passed back in.
     pub fn intern(key: u64, kind: HostValueKind) -> Arc<Self> {
-        let mut map = INTERNER.lock().unwrap_or_else(PoisonError::into_inner);
-        if let Some(existing) = map.get(&key).and_then(Weak::upgrade) {
+        Self::try_intern(key, kind).unwrap_or_else(|existing| {
             debug_assert_eq!(
                 existing.kind, kind,
                 "host-value key {key} re-interned with a different kind \
@@ -93,13 +92,28 @@ impl HostValueArc {
             // Prefer the existing live identity. The same key always carries
             // the same kind in practice; if a buggy caller disagrees we keep
             // the established one rather than mint a conflicting second Arc.
-            return existing;
+            existing
+        })
+    }
+
+    /// [`intern`](Self::intern) for untrusted input: a `kind` that disagrees
+    /// with the live registration for `key` is reported (as that registration)
+    /// instead of asserted, so a forged wire discriminator can neither panic
+    /// the boundary nor alias the established identity.
+    pub fn try_intern(key: u64, kind: HostValueKind) -> Result<Arc<Self>, Arc<Self>> {
+        let mut map = INTERNER.lock().unwrap_or_else(PoisonError::into_inner);
+        if let Some(existing) = map.get(&key).and_then(Weak::upgrade) {
+            return if existing.kind == kind {
+                Ok(existing)
+            } else {
+                Err(existing)
+            };
         }
         // No live entry (absent, or a dangling `Weak` whose `Arc` already
         // dropped). Create a fresh identity and record a `Weak` to it.
         let arc = Arc::new(Self { key, kind });
         map.insert(key, Arc::downgrade(&arc));
-        arc
+        Ok(arc)
     }
 }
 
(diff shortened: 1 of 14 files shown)


sxlijin:
Tried to reproduce all four "bugs fixed" on canary from the public Python SDK (plus code inspection of the Node/wasm/Go/Rust/Java bridges): none is reachable from any SDK's public surface today. #4 (callable key resolved inside the spawned task) is a real bridge defect, reproduced 500/500 by driving the async `BamlRuntime.call_function` with a `FunctionHandle` target and releasing the closure right after issuing, but every SDK's closure wrapper is sync or keeps the owner alive across the await; #1 only fires on an engine-internal invariant violation (`UnionSelectedTypeNotMember`); #2 and #3 have no producer in any bridge (host kinds only ride fresh registry keys and every `is_error` is a literal or bool).

Hardening fixes alone are not worth it: they introduce substantial complexity without adding concrete value and should generally not be done. The `prepare_call` / `invoke_prepared` consolidation is worth keeping to standardize the call pattern across adapters (and it closes #4 as a side effect), but the `OutboundOwnership` / `InboundTransfer` ledgers and `try_intern` are new abstractions for inputs nothing produces and should be dropped.

sxlijin:
Superseded by #4874 (the prepare_call / invoke_prepared consolidation on canary, without the ownership ledgers); see the findings thread above.

PR #4811, opened 2026-09-10
bridge_typescript: Node exit, coalesced host-value releases, and decode-error propagation

aaronvg:
## Summary

Five Node bridge lifetime fixes. The host-value release channel is rebuilt as a coalescing batch on napi-rs's own `ThreadsafeFunction` (no raw `napi_sys`), the dispatch channel no longer pins the event loop, and the SDK's exit hook waits for the engine to go idle via a new `Bex::wait_until_idle`.

## Bugs fixed

- **A Node process never exited while a host callback was registered.** The dispatch tsfn was strong, pinning libuv until BAML's GC dropped the closure. It is now weak; `beforeExit` awaits `_waitForRuntimeIdle()` so real pending work (including a background `spawn` re-entering BAML) still holds the loop.
- **Release notifications dropped on bursts over 4096 leaked JS registry entries forever.** Keys are parked in a pending set with at most one wakeup; JS receives batches of 1024 and a batch is only cleared after JS returns normally.
- **Thrown-payload decode failures were swallowed** (`catch { value = undefined }`), hiding the real error when a mapped class constructor failed. The error now propagates.
- **The unhandled-spawn tsfn used `Blocking` mode with a 1024 queue**, so an engine thread could block forever while the JS thread sat in `callFunctionSync`. Non-blocking, unbounded.
- **A poisoned registry mutex leaked the entry.** The poison is recovered through one accessor and the entry dropped outside the lock.

A `_hostValueCount()` diagnostic is exported, opt-in via `BAML_BRIDGE_DIAGNOSTICS=1` (always off on the Web bridge, present there only for package-root parity).

## Tests

All new behaviour tests run in CI under `sdk_test_typescript` in `sdk_tests/crates/typescript/function_calls/customizable/`: `callback_lifetime.test.ts` (three child-`node` scenarios asserting natural exit 0), `host_value_release.test.ts` (5000 thrown values drain back to baseline), `thrown_value_decode.test.ts`. Engine: three `wait_until_idle` tests in `bex_engine`. `dist/` regenerated.

## Stack

Part of the split of #4797; stacked on #4809 (rebased off #4810, which is closed; its consolidation half is #4874 on canary). Top of the stack.

## Summary by CodeRabbit

- **New Features**
  - Added runtime idle-wait support for TypeScript applications, allowing in-flight calls and background tasks to finish before shutdown.
  - Host-value release handling now supports batched cleanup for improved stability during high-volume activity.
  - Added opt-in diagnostics for monitoring retained host values.

- **Bug Fixes**
  - Improved process-exit behavior when callbacks or background tasks are still settling.
  - Errors thrown while decoding custom values now propagate correctly instead of appearing as undefined values.

---

> [!NOTE]
> **Medium Risk**
> Changes process exit timing, host-callable lifetime, and GC-driven release batching on a critical Node↔engine path; behaviour is heavily tested but regressions could affect shutdown or memory under burst host throws.
> 
> **Overview**
> Fixes **Node process lifetime** and **host-value cleanup** around registered JS callables, plus clearer error surfacing on the TypeScript bridge.
> 
> **Engine & SDK shutdown:** Adds **`BexEngine::wait_until_idle`** (and **`Bex::wait_until_idle`**) to wait for active calls and pending spawns **without** closing admission—unlike **`shutdown`**. The Node **`beforeExit`** hook now **`await`s `_waitForRuntimeIdle()`** before **`shutdownRuntime`**, so idle processes can exit while in-flight BAML work (including background **`spawn`** callbacks that re-enter BAML) still finishes.
> 
> **Event-loop pinning:** Host-callable dispatch **`ThreadsafeFunction`s are weak** so registration alone no longer pins libuv; the idle-wait promise holds the loop only while real work runs.
> 
> **Host-value release:** Replaces per-key notifications on a **4096-entry queue** (drops on **`QueueFull`**) with a **coalescing release channel** that batches up to **1024 keys** per JS wakeup and retries batches if the callback throws. TS registry uses **`_releaseHostValues(keys)`**; Wasm bridge adapts single-key releases to batches for API parity.
> 
> **Other bridge fixes:** **`decodeThrown`** no longer swallows decode/constructor failures into **`BamlError` with `value: undefined`**. Unhandled-spawn delivery is **non-blocking** with an **unbounded** queue (avoids deadlock with **`callFunctionSync`**). Registry mutex poison is recovered via a shared **`dispatch_table()`** accessor.
> 
> **Tests & diagnostics:** Engine **`wait_until_idle`** tests; SDK child-process exit tests, **5000**-throw release burst test, thrown-decode propagation test; BAML fixtures for callback lifetime. Opt-in **`_hostValueCount()`** via **`BAML_BRIDGE_DIAGNOSTICS=1`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa7a17d56e55631b03a3192ca092f01680278950. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sxlijin:
Tried to reproduce all five "bugs fixed" on canary through the public Node SDK only (`initializeRuntime` / `defineFunction` / `setTypeMap`, bridge built with `pnpm build:debug`, each scenario in a child `node` with a timeout so a hang is observable). Two are real user-reachable bugs, one needs a contrived sequence, two have no trigger from the SDK.

| # | Claim | Verdict | Evidence |
|---|-------|---------|----------|
| A | Node never exits while a callback is registered | **Reproduced, ordinary usage** | Plain async call exits 0 in 1.7 s; the same program with one completed callback hangs until killed at 15 s (also for a pending call and a background spawn). Calling `native.shutdownRuntime()` afterwards lets it exit, so the pin is the registry's strong tsfn and the `beforeExit` hook can never fire. |
| B | >4096 release burst leaks registry entries | Reproduced, contrived trigger | Only when the burst lands while the JS thread is blocked: retire the runtime with `initializeRuntime`, then immediately make a sync call; exactly 4095 of 5000 thrown errors are ever released (15905 of 20000 leak). With the JS thread free all 5000 drain. A natural trigger (GC inside a later sync call) released nothing at all in 15 s. |
| C | Thrown-payload decode failure swallowed | **Reproduced via `setTypeMap`** | A mapped class whose constructor throws yields `BamlError("baml error: user.MyError")` with `value: undefined` and the constructor's message gone. Generated classes are `constructor(init) { Object.assign(this, init) }`, so this needs a user-supplied typemap. |
| D | Blocking unhandled-spawn tsfn deadlocks a sync call | Not reproducible | Unhandled spawn errors are only discovered by a GC that finds the dead unobserved `Future`. Across sync/async, 100/2000 spawns, heap churn in the same call and in a later call, zero were delivered in-call. The only delivery path I could trigger is the exit hook's `shutdownRuntime`, where JS is awaiting a promise and free to drain the queue. |
| E | Poisoned registry mutex leaks the entry | Unreachable | No user code runs under that lock; only a bridge panic can poison it. |

On the fixes: A's actual fix is the one-line weak dispatch tsfn. `Bex::wait_until_idle` plus the exit-hook wait exist to keep fire-and-forget spawns alive past `main` (today they survive only by accident of the strong tsfn); that is a behaviour decision worth its own PR, not part of a "Node never exits" fix. B's ~120-line coalescing `ReleaseChannel` can be the same one-line change this PR already makes for the unhandled-spawn tsfn (unbounded, weak), which removes `QueueFull` with no new abstraction; `_hostValueCount` and its Web stub exist only to test B and can go with it. C's fix removes code, keep it. D and E are hardening for inputs the SDK cannot produce; dropping the `max_queue_size` is a harmless simplification, the `NonBlocking` switch and poison-recovery accessor fix nothing reachable.

Suggested shape: weak dispatch tsfn, unbounded weak release + unhandled-spawn tsfns, delete the `try/catch` in `decodeThrown`, keep `callback_lifetime.test.ts` and `thrown_value_decode.test.ts`.

PR #4844, opened 2026-09-11
Defer completion GC and reclaim idle native heaps

hellovai:
Builds on #4840. A completed call now returns without collecting. Allocation pressure is paid before the next root call imports arguments, or at the existing safepoints inside a running VM. Native engines also clean up after 100 ms of quiet; WASM services overdue cleanup at the next call.

### Before → after

| Scenario | #4840 | This PR |
| --- | --- | --- |
| A fast bridge call exceeds the budget | Full GC before returning the result | Return immediately; next call collects before allocating |
| Calls run continuously | Completion checks + VM safepoints | Entry checks + the same VM safepoints |
| A burst ends below the budget | Garbage waits for later pressure | Native worker attempts full GC after 100 ms quiet |
| Last bridge handle is dropped after that GC | Wait for later engine activity | Native worker wakes and arms another quiet collection |
| Detached child is still running or queued | No idle policy | It prevents idle cleanup until its producer exits and its registry work settles or is removed |
| A settled internal error remains in the registry | Error retained for later awaiters | Error stays rooted and readable, but no longer counts as active work |
| WASM stops receiving calls | No background cleanup | Still no background task or timer; cleanup waits for the next call |

```text
call A uses up budget:
  before:  A runs → full GC → return A
  after:   A runs → return A
                     ├─ B arrives: full GC → import B arguments → run B
                     └─ native stays quiet 100 ms: attempt idle GC
```

### Lifecycle

```text
on root-call entry:
    RootCallWork owns a new BexWorkGuard; cancel the idle deadline
    if allocation budget is spent, or WASM observed overdue cleanup:
        full GC before importing arguments
    run the call (existing allocation safepoints still apply)

on BexWorkGuard drop:
    mark cleanup pending
    if no work guards remain: deadline = now + 100 ms
    return

native worker, at deadline:
    if work resumed or shutdown started: stop this attempt
    acquire exclusive heap access, waiting at most 5 ms
    if acquisition times out: retain pending cleanup; retry after 100 ms
    if state changed while waiting: release access and stop this attempt
    otherwise: full GC; clear only the pending version it observed
```

GC bookkeeping is mandatory when constructing a futures registry, and every new entry requires a work guard. A retained settled entry takes and drops that guard exactly once; its lifecycle `Option` does not make tracking optional at construction. The worker, startup, and join methods exist only under native target configuration. WASM retains the shared bookkeeping for cooperative entry cleanup. Scheduling is an enum: `Disarmed`, `Waiting(deadline)`, `Suspended`, or `Closed`. Only `Waiting` can carry a deadline, and resuming a closed coordinator cannot rearm it.

There is one lazily started Tokio task per engine, with at most one timer. It holds a weak engine reference while waiting. A clean or busy engine has no deadline; the worker waits for notification. If its Tokio runtime was destroyed, a later call can restart the finished worker on the current runtime. Native idle cleanup requires that runtime to keep being polled.

`BexWork` in `bex_work.rs` owns work tracking, handle-release observation, and idle scheduling. Each work lifetime has a named owner with a mandatory, non-cloneable `BexWorkGuard`: `RootCallWork` for root-call registration, `FutureWork` for registry work until settlement or removal, and `SpawnedWork` for an entire producer task. Registering work with `BexWork::register_work()` disarms the idle deadline; dropping the final guard arms it. Producer scheduling goes through `SpawnedWork::new(work_guard, task).spawn()` on both native and WASM. The owner keeps its guard through queuing, execution, and cancellation cleanup, including when dropped before its first poll; the task future is destroyed before the guard is released. Registry settlement can precede a cancelled producer's exit, so both are counted. Queued task-group children count too. Streams/outstanding work therefore cannot be mistaken for idle. A retained internal-error entry releases its guard on settlement, including when cancellation won the terminal-state race. It remains a GC root so later awaiters can read the original error. Removing that retained entry later increments the existing root-release epoch and wakes cleanup, even if no other work completes. The producer keeps its independent guard through its remaining cleanup.

Bridge handles remain ordinary GC roots. Removing the last reference to a heap handle increments a release epoch and wakes the worker, including after the last call. Releases and work completions newer than the collection's snapshot remain pending. On WASM, an expired completion deadline is remembered at entry; if no deadline was armed, a newly observed handle release is eligible immediately because there was no idle observer.

Idle acquisition does **not** set the VM park-request flag. New work cancels a waiting attempt. Once collection has begun it runs to completion, and an arriving caller may wait for that full pause: **5 ms bounds only the acquisition wait, not GC or finalizers**. Shutdown suspends the worker; cancellation of shutdown resumes it; completed shutdown closes and joins it.

### Scope and validation

The retained-error lifecycle fix passed **226 unit tests** (126 engine, 100 heap) with `heap_debug,gc_profiling` in `fasttest`, plus full native and WASM Clippy. New regressions verify that retained errors survive moving GC and remain readable, a live producer still prevents idle collection, cancellation winning settlement releases registry work exactly once, and later entry removal re-arms cleanup across cancelled shutdown. Registry-count documentation includes retained engine-error entries; the count is not a pending-only metric.

After the work ownership refactor: **187 tests passed** with `heap_debug,gc_profiling` in `fasttest`: 124 engine unit tests, 8 allocation-budget tests, and 55 cancellation, fire-and-forget, future-cleanup, and task-group integration tests. New drop-order tests verify that a producer retains its guard through cancellation before or after its first poll, and through success or error cleanup. Full native workspace Clippy, the full WASM lint task, and repository commit hooks passed. The bridge regression now exhausts the slot budget with short calls, verifies that completion leaves the debt unpaid, and checks that the next entry collects while preserving the returned handle. Large-payload and old-array-growth tests verify that backing bytes alone do not trigger pressure GC.

The full collector and allocation-budget sizing remain those from #4840: charge reserved object slots only, and set headroom after full GC to `max(32 MiB, 4 × live slot bytes)`. String/image backing and container growth do not spend that budget. Quiet cleanup still runs below budget, so it can reclaim dead large payloads after a burst; continuous payload-heavy traffic can retain them until slot pressure or explicit GC. The 100 ms idle delay and 5 ms acquisition wait are internal starting constants. This adds no concurrent collection, allocator trimming, or public policy knobs. Reclaiming BAML objects does not guarantee the allocator returns their pages to the OS.

- On the initial implementation (`4a7ca4366`), 214 tests passed, one existing compiler-gap test ignored: engine unit tests plus allocation-budget, cancellation, cleanup-finalizer, fire-and-forget, host-callable, and task-group suites. Ran with `heap_debug,gc_profiling` in the optimized `fasttest` profile.
- After the state/platform refactor: 122 engine unit tests and 8 allocation-budget tests passed with `heap_debug,gc_profiling`; native all-feature Clippy and the full `mise run clippy-wasm` task passed. The added regression checks that work completion and GC completion cannot rearm suspended or closed cleanup.
- Paused-clock tests cover one worker per engine, coalescing, no repeated clean timer, release-after-last-call, acquisition timeout/retry, arriving-work cancellation, registry/producer overlap, detached completion, cleanup version races, cancelled shutdown, and worker lifetime.
- `cargo check -p bex_engine --target wasm32-unknown-unknown --features gc_profiling` passed. WASM was compile-checked; the cooperative state transition is unit-tested on native, not run in a browser.
- Targeted all-feature Clippy passed. All repository hooks passed, including full-workspace, all-target, all-feature Clippy.

Historical benchmark check (commit `4a7ca4366`, before the state/platform refactor **and before #4840 switched to slot-only accounting**; these figures do not measure the updated policy): three randomized paired runs per workload on this macOS host, same inherited harness, `GC_POLICY=current`, optimized binaries with `gc_profiling`. Percentages below are the median of paired changes relative to #4840. CPU covers the entire test process (including compilation of the BAML source, warmup, validation, and cleanup); call-loop time excludes those phases and the idle wait.

| Workload | Process CPU | Call-loop time | Peak sampled RSS |
| --- | ---: | ---: | ---: |
| 100k tiny calls | +1.4% | +0.3% | +3.4% |
| Allocation churn | +1.4% | -0.5% | -0.2% |
| 128 retained batches | +0.5% | -2.1% | -0.5% |
| 262k-object cache | +0.2% | -1.4% | -2.7% |
| Burst, then 500 ms idle | -0.5% | -1.6% | -3.1% |

In all three burst/idle runs, the parent retained **599,511 heap slots** after handles were released; this PR reached **zero runtime slots** during the idle interval, without another engine call. RSS stayed approximately flat across that collection. This proves the idle reclamation path, not OS page return. The tiny-call RSS increase is reported above rather than treated as a memory improvement.

These short, single-thread-runtime measurements are a regression check, not tuning sign-off. Busy multithreaded hosts, calls arriving during an idle collection, repeated quiet gaps with large retained caches, and target Linux allocators still need dedicated measurements before choosing final timing constants.

## Summary by CodeRabbit

- **New Features**
  - Added automatic idle-time garbage collection to reclaim unused memory between engine operations.
  - Garbage collection now runs opportunistically in the background when the engine is idle.
  - Idle cleanup responds to released resources and resumes appropriately during engine lifecycle changes.

- **Bug Fixes**
  - Improved memory cleanup timing after completed calls and released handles.
  - Improved cleanup coordination during shutdown and cancelled operations.
  - Preserved active work and resource handles while cleanup runs.
  - Improved collection scheduling when memory usage carries over between engine entries.

After the #4831 profiler refactor, explicit and idle collection pass the same compile-time real/no-op cycle recorder through permit acquisition and the parked collector. The engine lifecycle and collector contain no `gc_profiling` conditionals or feature-dependent parameters. After that profiler refactor, **230 heap/engine tests passed in each configuration**, with profiling enabled and disabled (`heap_debug` in both, optimized `fasttest`). This includes 100 heap tests, 122 engine unit tests, and 8 allocation-budget integration tests. Workspace native Clippy, full WASM Clippy with all features, and heap/engine WASM Clippy without profiling passed.

### DIFF
--- a/baml_language/crates/bex_engine/src/future.rs
+++ b/baml_language/crates/bex_engine/src/future.rs
@@ -186,18 +186,14 @@ impl FutureManagerGuard<'_> {
             .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         let id = FutureId::from_usize(id);
 
+        let work_permit = inner.idle_gc.acquire_work_permit();
         let ptr = inner
             .tlab
             .alloc_future(::bex_vm_types::Future::pending(id, returns, throws, cancel));
 
-        inner.active_futures.insert(
-            id,
-            FutureState {
-                _idle_work: inner.idle_gc.start_work(),
-                future: ptr,
-                origin,
-            },
-        );
+        inner
+            .active_futures
+            .insert(id, FutureWork::new(work_permit, ptr, origin));
         (id, ptr)
     }
 
@@ -563,7 +559,7 @@ impl HeapPermit<FutureManagerInner> for FutureManagerGuard<'_> {
 pub struct FutureManagerInner {
     tlab: Tlab,
     next_future_id: AtomicUsize,
-    active_futures: HashMap<FutureId, FutureState>,
+    active_futures: HashMap<FutureId, FutureWork>,
     // Mandatory bookkeeping on both targets. Only the native build contains
     // the background worker; WASM consumes the state at call entry.
     idle_gc: Arc<crate::idle_gc::IdleGc>,
@@ -612,9 +608,10 @@ impl TlabHolder for FutureManagerInner {
     }
 }
 
-struct FutureState {
+/// Owns a future's registry entry and its work permit until removal.
+struct FutureWork {
     // Mirrors registry ownership without making the timer inspect heap objects.
-    _idle_work: crate::idle_gc::WorkGuard,
+    _work_permit: crate::idle_gc::WorkPermit,
     /// Heap pointer to the `Object::Future`. Rooted via `RootHaver` so
     /// the heap object survives even when no awaiter / producer stack
     /// holds it directly (fire-and-forget spawn before the producer task
@@ -627,7 +624,19 @@ struct FutureState {
     /// hot-loop budget.
     origin: std::sync::Arc<str>,
 }
-impl FutureState {
+impl FutureWork {
+    fn new(
+        work_permit: crate::idle_gc::WorkPermit,
+        future: HeapPtr,
+        origin: std::sync::Arc<str>,
+    ) -> Self {
+        Self {
+            _work_permit: work_permit,
+            future,
+            origin,
+        }
+    }
+
     /// Returns an immutable reference to the heap-allocated `Future`.
     ///
     /// The [`bex_vm_types::Future`] uses interior mutability (`AtomicU8` +
@@ -650,7 +659,7 @@ impl FutureState {
         }
     }
 }
-impl RootHaver for FutureState {
+impl RootHaver for FutureWork {
     fn collect_roots(&self, roots: &mut Vec<HeapPtr>) {
         roots.push(self.future);
     }
--- a/baml_language/crates/bex_engine/src/future.rs
+++ b/baml_language/crates/bex_engine/src/future.rs
@@ -193,6 +193,10 @@ impl FutureManagerGuard<'_> {
         inner.active_futures.insert(
             id,
             FutureState {
+                _idle_work: inner
+                    .idle_gc
+                    .as_ref()
+                    .map(crate::idle_gc::IdleGc::start_work),
                 future: ptr,
                 origin,
             },
@@ -563,6 +567,7 @@ pub struct FutureManagerInner {
     tlab: Tlab,
     next_future_id: AtomicUsize,
     active_futures: HashMap<FutureId, FutureState>,
+    idle_gc: Option<Arc<crate::idle_gc::IdleGc>>,
 }
 impl FutureManagerInner {
     pub fn new(tlab: Tlab) -> Self {
@@ -570,6 +575,14 @@ impl FutureManagerInner {
             tlab,
             next_future_id: AtomicUsize::new(0),
             active_futures: HashMap::new(),
+            idle_gc: None,
+        }
+    }
+
+    pub(crate) fn with_idle_gc(tlab: Tlab, idle: Arc<crate::idle_gc::IdleGc>) -> Self {
+        Self {
+            idle_gc: Some(idle),
+            ..Self::new(tlab)
         }
     }
 
@@ -608,6 +621,8 @@ impl TlabHolder for FutureManagerInner {
 }
 
 struct FutureState {
+    // Mirrors registry ownership without making the timer inspect heap objects.
+    _idle_work: Option<crate::idle_gc::WorkGuard>,
     /// Heap pointer to the `Object::Future`. Rooted via `RootHaver` so
     /// the heap object survives even when no awaiter / producer stack
     /// holds it directly (fire-and-forget spawn before the producer task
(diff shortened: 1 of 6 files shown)


baml_language/crates/bex_engine/src/future.rs
```diff
@@ -563,13 +567,22 @@ pub struct FutureManagerInner {
     tlab: Tlab,
     next_future_id: AtomicUsize,
     active_futures: HashMap<FutureId, FutureState>,
+    idle_gc: Option<Arc<crate::idle_gc::IdleGc>>,
```
2kai2kai2, line 570: Is it possible to make this a feature flag instead of `Option` so native always has an idle gc handler and wasm never does and we can't accidentally violate that

baml_language/crates/bex_engine/src/idle_gc.rs
```diff
@@ -0,0 +1,633 @@
...
+    releases: usize,
+}
+
+struct State {
+    work: usize,
+    revision: u64,
+    collected: CleanupVersion,
+    observed_releases: usize,
+    deadline: Option<Instant>,
+    suspended: bool,
```
2kai2kai2, line 30: Can probably use an enum adt for the state stuff

PR #4895, opened 2026-09-16
bridge_typescript: Node exit, coalesced host-value releases, and decode-error propagation

sxlijin:
## Summary

Five Node bridge lifetime fixes. The host-value release channel is rebuilt as a coalescing batch on napi-rs's own `ThreadsafeFunction` (no raw `napi_sys`), the dispatch channel no longer pins the event loop, and the SDK's exit hook waits for the engine to go idle via a new `Bex::wait_until_idle`.

## Bugs fixed

- **A Node process never exited while a host callback was registered.** The dispatch tsfn was strong, pinning libuv until BAML's GC dropped the closure. It is now weak; `beforeExit` awaits `_waitForRuntimeIdle()` so real pending work (including a background `spawn` re-entering BAML) still holds the loop.
- **Release notifications dropped on bursts over 4096 leaked JS registry entries forever.** Keys are parked in a pending set with at most one wakeup; JS receives batches of 1024 and a batch is only cleared after JS returns normally.
- **Thrown-payload decode failures were swallowed** (`catch { value = undefined }`), hiding the real error when a mapped class constructor failed. The error now propagates.
- **The unhandled-spawn tsfn used `Blocking` mode with a 1024 queue**, so an engine thread could block forever while the JS thread sat in `callFunctionSync`. Non-blocking, unbounded.
- **A poisoned registry mutex leaked the entry.** The poison is recovered through one accessor and the entry dropped outside the lock.

A `_hostValueCount()` diagnostic is exported, opt-in via `BAML_BRIDGE_DIAGNOSTICS=1` (always off on the Web bridge, present there only for package-root parity).

## Tests

All new behaviour tests run in CI under `sdk_test_typescript` in `sdk_tests/crates/typescript/function_calls/customizable/`: `callback_lifetime.test.ts` (three child-`node` scenarios asserting natural exit 0), `host_value_release.test.ts` (5000 thrown values drain back to baseline), `thrown_value_decode.test.ts`. Engine: three `wait_until_idle` tests in `bex_engine`. `dist/` regenerated.

## Stack

Part of the split of #4797; stacked on #4809 (rebased off the now-closed #4810). Replaces #4811, which GitHub auto-closed when the stale stack branch was removed and whose base its stack feature would not let us change.


sxlijin:
Folded back into #4811: the stack was dissolved with `gh stack unstack`, #4811 reopened and retargeted onto #4809, so this duplicate is no longer needed.

PR #4905, opened 2026-09-17
Expose bigint literal values through reflection and JSON

hellovai:
## Literal reflection: supported values

`reflect.literal.new` accepts exactly `string | int | bigint | bool`. The returned literal view's `value()` exposes the same value, including for a literal type obtained through `reflect.Type`:

```baml
test "literal reflection" {
    assert.equal(reflect.literal.new("fixed").value(), "fixed");
    assert.equal(reflect.literal.new(-42).value(), -42);
    assert.equal(reflect.literal.new(123456789012345678901234567890n).value(), 123456789012345678901234567890n);
    assert.equal(reflect.literal.new(true).value(), true);
    assert.equal(
        (reflect.Type.of<"fixed">().as_literal() ?? throw "expected literal").value(),
        "fixed",
    )
}
```

Other runtime values, such as classes, arrays, and floats, are outside `reflect.literal.new`'s input type. `value()` returns `string | int | bigint | bool`; an unsupported internal literal kind raises `baml.errors.InvalidArgument`.

## Typed bigint JSON: supported and rejected input

BAML encodes `bigint` as a **numeric JSON string** containing an optional sign and integer digits. Typed decoding now accepts that form for `bigint`, an exact bigint literal type, and bigint fields nested in classes:

```baml
class BigintRecord { id bigint  exact 42n }

test "bigint JSON" {
    let large = 123456789012345678901234567890n;
    let encoded = baml.json.to_string(large);
    assert.equal(encoded, "\"123456789012345678901234567890\"");
    assert.equal(baml.json.from_string<bigint>(encoded), large);
    assert.equal(baml.json.from_string<42n>("\"42\""), 42n);
    let record = baml.json.from_string<BigintRecord>(
        "{\"id\":\"123456789012345678901234567890\",\"exact\":\"42\"}",
    );
    assert.equal(record.id, large);
    assert.equal(record.exact, 42n)
}
```

| Input to typed decoder | Result |
| --- | --- |
| `"42"` as `bigint` or `42n` | Accepted. |
| `42` as `bigint` | `baml.json.DecodeError`: bigint is encoded as a string. |
| `"43"` as `42n` | `baml.json.DecodeError`: literal value differs. |
| `"12.5"`, malformed numeric text, or an oversized bigint string | `baml.json.DecodeError`. |

Bigint tokens in expressions and type positions also round-trip through the formatter. Literals above the VM's bigint bit limit now receive the compiler's literal-size diagnostic before entering the constant pool.

## Verification and review path

- Both BAML examples in this description were run and passed. Focused BAML reflection and JSON tests: 4 passed. Formatter unit suite: 155 passed. Compiler AST unit suite: 81 passed. Corpus formatter, pre-commit checks, and the current CI matrix are green.
- `crates/baml_builtins2/baml_std/reflect/ns_literal/literal.baml` and `crates/bex_vm/src/package_reflect/type_kinds.rs`: reflection contract and implementation.
- `crates/bex_vm/src/package_baml/json.rs`: typed bigint decoder.
- `crates/baml_compiler2_ast/src/lib.rs`: compiler size check for bigint literals.
- `crates/baml_fmt/src/ast/`: bigint formatter support.
- `crates/baml_tests/baml_src/ns_type_kinds/type_kinds.baml` and `crates/baml_tests/baml_src/ns_json_to_from_string/json_to_from_string.baml`: examples and rejection tests.

This is the base PR for the TypeSafe AI provider stack; #4906 targets this branch.

## Summary by CodeRabbit

* **New Features**
  * Added support for serializing and deserializing `bigint` values from numeric JSON strings.
  * Added support for creating and inspecting bigint literal types through reflection, including retrieving their values.
  * Added support for formatting and parsing bigint literals.
  * Added validation for invalid, mismatched, or oversized bigint values during JSON decoding.
  * Added diagnostics for bigint literals that exceed supported size limits.

* **Documentation**
  * Clarified bigint JSON encoding and decoding behavior.


aaronvg (approved):

sxlijin:
After looking more closely at prior art, I think the quoted-decimal-string wire format is the wrong direction and we should not land the JSON part of this PR as-is.

The common arbitrary-precision JSON implementations I checked use **numeric JSON literals**, not quoted strings:

- Python's standard `json` maps JSON integers to arbitrary-precision `int` and emits them as numbers: https://docs.python.org/3/library/json.html#json-to-py-table
- Go's `math/big.Int` `MarshalJSON` emits a number and `UnmarshalJSON` accepts a numeric token: https://pkg.go.dev/math/big#Int.MarshalJSON and https://pkg.go.dev/math/big#Int.UnmarshalJSON
- `json-bigint` parses and stringifies big integers as JSON numbers: https://www.npmjs.com/package/json-bigint
- `serde_json`'s `arbitrary_precision` feature preserves arbitrary-precision JSON numbers as `Number`: https://docs.rs/serde_json/latest/serde_json/struct.Number.html

With this PR's string representation, we are wire-incompatible with those serializers in both directions: their output such as `{"data":123123123123123123123123}` is rejected by BAML's typed bigint decoder, while BAML's `{"data":"123123123123123123123123"}` is read as a string (or rejected for a typed bigint field) by those implementations. Using strings also makes a generic `json` roundtrip lose bigint identity.

I think the right contract is:

- serialize `bigint` as a bare integral JSON number;
- deserialize `bigint` only from integral JSON number tokens, preserving the token exactly via `serde_json`'s already-enabled `arbitrary_precision` support;
- add `bigint` to `baml.json.json`, so something like `{"foo":123123123123123123123123}` parses and serializes without precision loss.

Native JavaScript's `JSON.parse`/`JSON.stringify` remains a caveat, but using quoted digits here would optimize for native JS at the cost of incompatibility with the established lossless/arbitrary-precision serializers above. I'll put up a replacement PR with the numeric-literal approach.

sxlijin:
Superseded by #4934

PR #4915, opened 2026-09-17
Fix silent SAP array element discard

### LINKED ISSUE #4590: SAP silently discards array elements it cannot coerce — `baml run` returns exit 0 with fewer rows and NO signal on any channel (stdout, stderr, `--output-format debug`, `--log-file`, `BAML_LOG`)
## Summary

When Schema-Aligned Parsing cannot coerce ONE element of a list, the element is dropped and the call
SUCCEEDS with a shorter list. Nothing in the result, the exit code, stderr, the debug rendering or
the (never-written) run log records that an element was discarded, and the raw model text that would
let a caller notice is not exposed anywhere on the `baml run` surface.

For a caller that treats absence as meaning "the model found nothing there", a dropped row is
indistinguishable from a correct empty answer. That asymmetry is the problem: a bare-null REQUIRED
field at the TOP level fails the whole call loudly and recoverably, while the same defect one level
down, inside a list element, returns success with data missing.

## Why this matters to a real consumer

We run a German insolvency-administration platform (Insolvenzverwaltung). BAML functions read
statutory documents — bank statements, creditor filings, asset schedules — and every row becomes a
legally significant fact. A `Kontoauszug` page yields a `buchungen: Buchung[]`; each `Buchung` is a
payment.

A silently dropped `Buchung` is a payment the system never saw. It cannot appear in a
payment-avoidance check (§§ 129 ff. InsO), a balance reconciliation, or a distribution to creditors,
and it leaves no counter, no park, and no "needs human" marker anywhere — because from the caller's
side nothing happened. A loud failure would be strictly better: we retry, re-read, or hand the page
to a human. Success-with-less-data is the one outcome we cannot defend.

Our own architectural rule is that every automatic error must land on the visible side: a gap we can
count, never an absence we cannot see. Today SAP's list behaviour puts it on the invisible side and
we have no way to move it.

## Measurement (BAML 0.17.0, `baml wrapper 0.2.0`)

An HTTP sink stands in for the model and answers with a fixed completion. The BAML function is

```baml
class Buchung {
    datum: GroundedDate,      // required
    betrag: GroundedAmount,   // required
    gegenpartei: GroundedText?,
    verwendungszweck: GroundedText?,
    umsatzart: GroundedText?,
}
class KontoauszugFelder {
    kontostand: GroundedAmount,
    iban: GroundedText,
    girokonto_kreditinstitut: GroundedText?,
    kontostand_datum: GroundedDate?,
    buchungen: Buchung[],
}
function Extract_Kontoauszug(page_image: string) -> KontoauszugFelder { … }
```

The sink returns TWO `buchungen` rows. Row 1 carries `"datum": null` (a required field, bare null —
the shape a model produces when it read the row but could not read the date). Row 2 is well formed.

```
$ echo '{"page_image":"Kontoauszug Text"}' \
  | baml run --from baml_src --output-format json \
             -f Extract_Kontoauszug -- Extract_Kontoauszug --json-args -
```

| channel | result |
|---|---|
| exit code | **0** |
| stdout (`--output-format json`) | `{"kontostand":…,"buchungen":[ <row 2 only> ]}` — **one** row |
| stdout (`--output-format debug`) | same single `user.Buchung { … }` |
| stderr | `warning: code is unformatted; run \`baml fmt\`` — **nothing else** |
| `--log-file <path>` | **no file is written at all**, with or without `-vvv` |
| `BAML_LOG=debug` / `BAML_LOG=trace` / `RUST_LOG=debug` | stderr unchanged (45 bytes) |
| `-vvv` stderr | `Loading project…/Found 47 .baml file(s)/Bytecode cache hit/warning: …/Calling user.Extract_Kontoauszug/Completed in 22.15ms` — no mention of the row |

Control, same command, same sink, both rows well formed: stdout carries **two** rows. So the
difference is exactly the discarded element, and nothing reports it.

The same drop is measured on the native path (`baml_bridge`, generated `sdkgen_rust` crate): `Ok`,
row dropped, no signal. So this is engine behaviour, not a CLI-only gap.

## Why the SDK default makes it urgent for us

We have just made the in-process native SDK our default seam (it is ~11 ms of overhead against
1.5–5 s of CLI process spawn). On the CLI seam we at least held the process's raw stdout in our
hands and could have compared something; the measurement above shows even that was never true —
stdout is already post-coercion. On the SDK seam `baml_bridge` hands back an already-decoded value
and exposes no raw response either. So the flip does not make the defect worse, but it does remove
the last place a consumer could have hoped to instrument it, and it means BOTH of our seams are
blind to a class of data loss that is silent by construction.

## What we are asking for

Any ONE of these unblocks us; they are listed cheapest-first from our side.

1. **Report the discard.** Make a coerced value carry the elements SAP dropped, or at minimum a
   count and a path, on both the CLI result envelope and the bridge/SDK result. A number is enough:
   we can turn "3 elements discarded under `buchungen`" into a counted, human-visible gap.
2. **A strictness knob for list elements.** Opt in, per type or per project, to "an element that
   fails to coerce fails the call" — the same posture a top-level required field already has. We
   would take the loud failure for these functions without hesitation.
3. **Expose the raw model text alongside the parsed value** on the `baml run` surface and on the
   bridge/SDK result (a `--raw-output <path>` flag, or a field on the returned envelope). Then the
   comparison is ours to make. Note `--log-file` looks like it is meant to do something in this
   space but wrote no file in any invocation we tried, which may be a separate bug.

We are aware that BAML 0.17 already exposes the primitives — `ai.Client::invoke` returns an
`ai.ModelTurn` whose `terminal_text()` is the raw string, and `baml.sap.parse<T>` is an ordinary
stdlib function — so a function written imperatively can hold both. That is not a usable answer for
us at this size: it would mean rewriting 33 declarative `prompt:` functions imperatively and
widening every one of their return types to carry the raw text, on prompts whose rendering we have
just verified byte-identical through a migration. What we need is for the DECLARATIVE form to be
able to tell us what it threw away.

## Repro

Minimal repro (a class with one required field, a list of it, a client pointed at any HTTP sink that
returns a two-element list whose first element has that field as `null`) reproduces on a 16-line
project; we can attach one if useful.

---

### Notes for the orchestrator (not part of the issue body)

- Everything above is MEASURED in worktree `b3-412` on 2026-08-25 with the box toolchain
  (`baml wrapper 0.2.0`, `baml toolchain 0.17.0`); commands and raw outputs are in
  `<scratchpad>/b3-412-cli-stdout.txt` / `b3-412-cli-stderr.txt` and reproduced by the committed
  probe `baml_sdk_probe::array_row_drop_is_invisible_at_the_cli_seam`.
- The SDK-seam sentence ("Ok, row dropped, no signal") is quoted from gh #412's own posture matrix
  (the #175 spike), NOT re-measured by this slice — say so if upstream asks.
- The `--log-file` observation is arguably its own filing; folded in here as a note rather than
  split, since it is the flag a maintainer would otherwise point us at.

sxlijin:
## Issue Reference

- [x] Fixes #4590
- Linear: BAMLGH-64

## Changes

Completed SAP parses now fail through the existing parse-error channel when coercing a real array records one or more `ArrayItemParseError` flags. The error includes the number of rejected elements and preserves the original nested parsing paths, so CLI invocations exit non-zero and SDK/bridge callers receive an error instead of a shortened successful value.

Candidate selection and partial streaming remain tolerant. The completed-parse check runs immediately before deserializer metadata would otherwise be erased during VM value conversion, and it excludes the established non-array-to-singleton fallback so a text-only response can still represent an absent `image[]`.

The repository size baselines were refreshed with the exact CI measurements because current canary artifact growth had moved WASM 0.1 percentage point beyond the existing 3% delta gate; the percentage policy is unchanged.

## Testing

- [x] Added an end-to-end engine regression matching the reported two-row `buchungen` case
- [x] `cargo test -p baml_tests completed_sap_parse_rejects_discarded_array_items --lib`
- [x] `cargo test -p baml_cli --test baml_corpus baml_test -- --nocapture` (4,505 passed, 2 pre-existing tolerated failures)
- [x] `cargo test -p bex_sap` (595 passed)
- [x] `cargo test -p sys_ops` (102 passed)
- [x] `cargo clippy -p bex_sap -p sys_ops -p baml_tests --tests -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`
- [x] Full GitHub CI (44 passed, 25 intentionally skipped, 0 failed)

## Screenshots

Not applicable; no VS Code extension or playground UI files changed.

## PR Checklist

- [x] Code follows the project style
- [x] Self-review completed
- [x] Focused comments document the metadata boundary
- [x] No new warnings

## Additional Notes

BAMLGH-64 remains open until this PR is merged.


sxlijin:
Closing as unprioritized; this is not the right implementation approach for the issue.

PR #4957, opened 2026-09-21
ci: run the slowest Linux legs a second time on Trunk Kiln

eltonkl:
## Issue Reference

- [ ] This PR fixes/closes #[issue number]

No issue — this follows the ongoing conversation with the BAML team in our shared Slack channel. Happy to open an issue instead if you'd prefer the discussion there.

## Changes

Adds a shadow CI workflow that runs six of this repository's slowest Linux legs a second time, on Trunk's Kiln runners, so the two can be compared on the same commits. It duplicates those legs only; the rest of CI is untouched.

**It is not a migration, and it does not affect merging.** The workflow is not a required check and every job sets `continue-on-error`, so a failure reports neutral rather than red. Your existing CI keeps running exactly as before. To stop it at any point, delete `.github/workflows/kiln-shadow.yml`.

**It costs this repository nothing to run.** The runners are self-hosted, so they use no GitHub Actions minutes and no capacity on the runners already in use. No secrets or credentials are involved — the diff contains no `secrets.` reference.

One file, `.github/workflows/kiln-shadow.yml`: `cargo test` on linux-gnu and musl, snapshot tests, two SDK legs (TypeScript and C#), and the SDK parity lint.

sccache comes from your own direnv setup and `baml-sccache` wrapper — the jobs run `direnv allow` and `direnv export gha` like the rest of your CI, so sccache uses the same backend it always does. This PR adds no sccache configuration of its own.

## Testing

- [x] Manual testing performed — these legs have run end to end on a private mirror of this repository: six legs green, reproduced across three runs.
- [ ] Unit tests added/updated — n/a, CI configuration only.

Not yet run against this repository, which is what this PR is for.

## Screenshots

n/a.

## PR Checklist

- [ ] I have made corresponding changes to the documentation — n/a

## Additional Notes

**sccache has no shared backend while this sits as a fork PR.** GitHub does not pass secrets to a `pull_request` run from a fork, so `BAML_SCCACHE_R2_*` are absent here, and `.envrc` falls back to a host-local disk cache. The legs that lean on sccache therefore compile cold, and will look slower than your CI until this merges — after which `canary` runs hold the secrets and use your bucket exactly as the rest of CI does. Note that once merged, these duplicate legs both read and write that bucket; sccache is content-addressed, so the writes are idempotent, but it is your storage.

**These jobs emit telemetry as your CI does.** They run `baml-cli` the same way, so they produce the same events — and because they duplicate existing legs, some events will be counted twice. Setting `DO_NOT_TRACK: "1"` at the workflow level suppresses that, at the cost of two `baml_cli` tests that assume telemetry is on.

**The first runs will be slow, for two reasons that both resolve on merge.** sccache has no backend here, per above; and Trunk's `actions/cache` lane is populated by runs on the default branch, which this repository has none of yet. The first run against `canary` fixes both. The one leg that does not depend on either — `cargo test` on linux-gnu, carried by a prebuilt `target/` — already runs faster here than its counterpart in your CI.

A maintainer may need to approve workflow runs, since this comes from a fork.

Commits carry a `Co-Authored-By` trailer for the AI assistance used in preparing them.

## Summary by CodeRabbit

- **Chores**
  - Added a non-blocking validation workflow for pull requests, pushes to the canary branch, and manual runs.
  - Expanded automated coverage across Linux, GNU, musl, SDK, snapshot, and parity checks.
  - Added C# package/version validation, build caching, diagnostic uploads, and retained parity reports.
  - Snapshot differences are reported as failures, while validation jobs do not block development workflows.


.github/actions/enable-sccache/action.yml
```diff
@@ -0,0 +1,125 @@
+name: Enable sccache
```
sxlijin: delete this file: we use a direnv setup plus a custom wrapper script to launch sccache

sxlijin (approved):

